PR #3, opened 2021-03-11
Add initial pathway reconstruction algorithm classes

agitter:
I'm opening this pull request so we can use the review functionality to discuss which files to clean up or delete before merging this exploratory branch.  This branch is the former `adam-dev` branch and includes the revisions from the former `prepare-input` branch.

These files look like candidates to delete before merging:
- Config-Files/config_test.yaml
- Config-Files/prancer_config.yaml
- PRRun/log.txt
- PRRun/pathlinkerRunner.py
- docker/pathlinker/*
- test.yaml

Do we need to keep any of those?  Are there other files we can delete?


annaritz:
I think the yaml files you list can be deleted, along with `PRRun/log.txt` and `PRRun/pathlinkerRunner.py`, I believe (those were from the original BEELINE-like implementation). It would be good to get confirmation from @csmagnano about the yaml files and @ajshedivy about the docker/ directory.

agitter:
I deleted the files I listed above in 3b07513, but there are still more we should be able to prune.  @ajshedivy can you please review which files were experimental and can be deleted?  Can we remove the entire PRRun directory?  You can delete files directly and push the changes to this branch.

Because we no longer plan to instantiate runner objects, I reverted that part of the Snakefile in db0c2ce and 5ef6360.

I updated run_static for PathLinker to use the input arguments.  There is still a lot of leftover code from the previous class-based approach that was going to instantiate objects.  Once some of that has been deleted, we may be ready to merge.

agitter:
@ajshedivy and I worked on this, and he deleted the remaining files that were early prototypes and no longer needed.

I'll leave this open for a few more days in case anyone else wants to comment.  If there are no objections, I'll merge it.

PR #12, opened 2021-04-26
Update snakefile and data loader to support preparing inputs

agitter:
These initial commits sketch to how unite data loading from the config file, the Snakemake workflow, and the algorithm-specific prepare inputs.  I'd like to get some feedback before moving forward in this direction.

The major changes are:
- The user-provided inputs in the config file are renamed and reorganized to be node-, edge-, and other-centric
- Multiple input datasets can be listed in the config file
- Optional dataset labels are used to help Snakemake manage which datasets have been loaded and processed, default labels can be used if they are not provided
- The workflow now has a `merge_input` rule that runs once per dataset before the `prepare_input` rule, which runs once per dataset-algorithm pair
- The DataLoader is restructured to match the config file changes (but requires additional changes)

I propose that `DataLoader.py` be changed to `Dataset.py`.  It will not take the entire config file as input but rather the dict from the config file that corresponds to a specific dataset.  It will have methods to export and import from a pickle file to make it Snakemake-friendly.


annaritz:
Did a high-level read through, and this looks great. The only thing I had a hard time understanding was the `data0` label and then the other (default?) dataset in the config file.  Would the `node_files` and `edge_files` be overwritten for the `data0` label? What if there were multiple tags without different data labels? etc. We can talk tomorrow.

agitter:
I expect that a user would either provide data labels for all of their datasets (recommended) or none.  The current config file with one labeled and one unlabeled dataset is a little confusing and is intended to illustrate the behavior of the parser.

The code doesn't currently check that the data label is unique, but it should.   Or it should add a suffix to make it unique and log a warning.

The node and edge files won't be overwritten.  Every rule in the Snakefile will add a new layer of intermediate outputs.  The progression would be something like:
1) Original inputs with label `data0`: `sources.txt`, `targets.txt`, `node-prizes.txt`, `network.txt`
2) Inputs merged into a dataset object: `data0-merged.pickle`
3) Inputs prepared for an algorithm: `data0-pathlinker-nodetypes.txt`, `data0-pathlinker-network.txt`
4) Raw reconstructed pathway from an algorithm with the 0th parameter combination: `raw-pathway-data0-pathlinker-params0.txt`
5) Reconstructed pathway in the universal format: `pathway-data0-pathlinker-params0.txt`

Filename patterns are easy to change, and we'll likely use subdirectories to keep these files organized.

csmagnano:
I like this restructuring. It makes sense to move dataset handling to the snakemake level, and thus have an object for each dataset. 

The only problem I see is that, if certain files are shared between datasets, we would be making copies of that file. But I don't think that there are other parts of the workflow which are going to use a lot more space, so it shouldn't be an issue.

agitter:
@annaritz suggested we add a logfile that lists the label and properties of each dataset.  This will be helpful if we add default labels to some datasets.

csmagnano:
@agitter I've refactored the name and changed the code such that I think the Dataset class now works as we were planning. I also added a small merge_input function to PRRunner, though this might change as we finalize data indexing in Snakemake

agitter:
With these latest commits, the core of the workflow is functional.  It can read datasets and pathway reconstruction algorithm parameters from the config file, prepare the datasets, run the algorithms on the datasets for all parameters.  It calls the parse output rule as well, but we haven't implemented that for any algorithm yet.

@ajshedivy please see the modifications I had to make to PathLinker's `run` function.  These are the same ones we talked about during our last meeting, but I needed to add them on this branch in order to test the Snakefile.  The `run` function itself is still slightly broken and can be fixed on the `pathlinker` branch.

I'm focusing on the core features first for our initial demonstration and have ignored:
- config file validation (assumes expected fields exist)
- supporting unlabeled datasets and creating default labels
- dataset logging

agitter:
I noticed that the prepare_input rule may actually be running once per output instead of once per input.  The reason why is complicated and will require some discussion.  I suggest we tolerate that for now and fix it in a separate pull request.

annaritz:
This looks great, works as expected for me.

agitter:
I'm merging this as is so that we can sync these changes with the other active branches.  I'll open new issues to catalog the remaining known issues.  @csmagnano will review it in more detail post-merge.

PR #24, opened 2021-05-10
Omics integrator

csmagnano:
Omics integrator now appears to work in prepare_input and reconstruct. The docker containers work, though there are some longer-standing issues present. 

The pytest module is having some operating system trouble, but for the purpose of having a functional demo this is ready to merge into master from what I can tell.


csmagnano:
The checks are currently failing but it appears to be some OS issues and the fact that we don't have parse_outputs implemented on  this branch.

Dataset.py
```diff
@@ -101,6 +101,21 @@ def request_node_columns(self, col_names):
             warnings.warn("Only %0.2f of data had one or more of the following columns filled:"%(percent_hit) + str(col_names))
         return filtered_table
 
+    def contains_node_columns(self, col_names):
+        '''
+        col_names: A list-like object of column names to check or a string of a single column name to check.
+        returns: Whether or not all columns in col_names exist in the dataset.
+        '''
+        if isinstance(col_names, str):
+            if not col_names in self.node_table.columns:
```
agitter, line 110: We can simplify this logic

agitter (commented):
These changes look good to me.  I have a few minor comments noted below.  I'll push a small update and then merge.

We can double-check the tests all pass after we sync this with the output-analysis branch.  It's a good sign that the pytest tests pass and the Snakemake workflow makes it to the parse output rule for Omics Integrator.

PR #25, opened 2021-05-11
Pathlinker

ajshedivy:
pathlinker run and docker container updates

the only major difference between pathlinker and oi1 and oi2 is how the code is executed in the container. A call to pathlinker:
`python run.py network nodetypes -k 100` will generate `outk-100-ranked-edges.txt` in the current working directory. In order to redirect the output of pathlinker (from within the container) to the specified output location, I first copy the input files to a tmp directory. I then mount this tmp directory along with the output location to parallel locations in the container `/home/files` and `/home/out` where the pathlinker code is in the `home` directory. I set the working_dir to be /home/out in the container. That way pathlinker runs from the `/home/out` dir and the output file is automatically linked back to the specified location. the tmp dir is then discarded after execution. 

We may need to make changes in order to plug this into snakemake. For the future we should talk about our general approach to handling outputs, working directories and mounts to the containers.


agitter (commented):
Thanks for the PathLinker updates.  I left some initial comments and am testing locally.

I will make some edits to the run function so that it is compatible with the Snakemake reconstruct rule.  The rule defines an output filename, so the run function has to be able to generate a file with that name.  I can tweak what you have to do that.

agitter:
I fixed a couple bugs revealed by the GitHub Actions tests.  The Snakemake workflow now makes it to the parse_output rule for Omics Integrator 2.  I'll merge.

Issue #16, opened 2021-05-06
Better support for dataset labels in the config file


agitter:
There is discussion about dataset labels in #12 that was not implemented.  This includes:
- make the dataset label field optional in the config file
- checking for duplicate dataset labels and automatically making the labels unique (with a warning)
- automatically assigning dataset labels for unlabeled datasets and warning the user
- logging dataset properties
- making the config file parsing robust to missing required fields

annaritz:
There seems to be required filenames for dataset filenames. For example, there needs to be a "sources" and "targets" header. In `Datasets.py`, it appears that header names are stripped from the filename prefix in `load_files_from_dict()`.  We might want to standardize filenames too.

agitter:
For the node properties, the primary way to specify attributes is to include a header row.  The column header becomes the attribute name.  This does rely on knowing which attributes each algorithm expects, which is not the most reliable approach.

A secondary way to specify attributes is a shorthand for Boolean attributes.  There is no header row, only a list of node ids that should all have the attribute set to True.  Because there is no header, the filename (without the extension suffix) is used as the attribute name.

We though the secondary format might be convenient for listing sources, targets, and similar node attributes.  Is it too confusing to have both formats?

This does need to be documented.  Should it go in the main readme for now?

tristan-f-r:
I disagree with allowing optional dataset labels / making dataset labels unique. This requires that someone is actively monitoring the snakemake log for these warnings, which isn't something that one usually expects out of their users (especially with the massive amount of logs we produce).

PR #36, opened 2021-05-24
Precision-Recall Code

annaritz:
Opening this draft pull request to track commits and decisions as we work on the `precision-recall` branch.  Some considerations for this part of the code:
- [x] Each dataset can have a `ground_truth_files` label. These can be edge or node files. (Question: what happens if there are multiple network files for a dataset? It is currently written as a list.).
- [ ] For each dataset, all algorithms where `include: true` will be analyzed for precision-recall. One PR figure will be written for each dataset, with each algorithm is a line or point on the figure.
- [ ] I will write log files that track which algorithms are drawn for which dataset, similar to params and datasets.

Prec-rec options are: 
- [ ] `item` to be assessed: list containing one or more of `nodes`, `dir-edges`, or `undir-edges`.
- [x] `subsample` negatives: sample `<int>` times the number of positives, or specify `all` for all negatives.
- [ ] Future option: `rank` (sort by increasing order) vs. `score` (rank by decreasing order)
- [ ] Future options: customize figure, make separate panels, etc.

PRAUG also has the ability to subsample negatives multiple times, and find the one that produces the median performance for all evaluated algorithms. This is future work as well.


annaritz:
I noticed that input files need to be named in specific ways. Is this noted anywhere? Also mentioned in #16.

annaritz:
`Dataset.py` now builds an `edge_table` Pandas dataframe. This dataframe is an **undirected** accounting of the edges. It is initialized with an undirected version of the interactome, and new columns denote ground truth edge files.  The general process to build the `edge_table` dataframe is:
1. Take the "Interactor1" and "Interactor2" columns from the `interactome` DataFrame.
2. Sort each row lexicographically. That is, the row `['P15498' 'P52735']` becomes `['P52735' 'P15498']`.
3. Drop duplicate rows.

When merging a ground truth file, we do the same sorting & dropping duplicates before merging.

We should discuss the fact that this table is undirected. It will be very handy for precision-recall (both in subsampling negatives and checking for matches) but it might not be necessary to build for every pipeline.

agitter:
The file `input/netpath-nets/bcr-edges.txt` contains duplicate edges.  That is causing the test failure shown in https://github.com/Reed-CompBio/spras/issues/47#issuecomment-880972963

I do not think that we currently need to support multigraphs.  I'm not aware of any pathway reconstruction algorithms that make use of the multiple edges.  Can our network loading code ensure we have at most one edge between a pair of nodes, or does that need to happen in prepare_inputs for algorithms like MEO that require single edges?

annaritz:
@agitter I agree that we shouldn't support multigraphs. I'll fix that (this was a test file & not generated from any of the spras code). 

I think the network loading code can handle checking for duplicate edges, though there still might be an issue with bidirected edges (u,v) and (v,u) being interpreted as two undirected edges for some algs. Perhaps for pathway reconstruction algorithms on undirected graphs, we recommend that they do an additional check in `prepare_inputs` to handle this case.

annaritz:
It turns out that the background interactome (`np-union.txt`) also contained duplicate edges. Commit 4038223 removes duplicated edges from the interactome.

annaritz:
I _think_ the failed checks are because of bidirected edges for `meo`.

Screen Shot 2021-07-29 at 5 27 26 PM

This edge appears as (u,v) and (v,u) in the `prepared` edge file:

```
$ grep O00459 output/prepared-bcr-meo-edges.txt | grep O00329
O00329	(pp)	O00459	1
O00459	(pp)	O00329	1
```

src/meo.py
```diff
@@ -73,7 +74,8 @@ def generate_inputs(data, filename_map):
 
         # TODO need to support partially directed graphs
         # Expected columns are Node1 EdgeType Node2 Weight
-        edges = data.get_interactome()
+        edges = data.get_undirected_interactome()
+
```
annaritz, line 78: `meo` still had trouble with the input files (especially the interactome) because bidirected edges were represented as `(u,v)` and `(v,u)`.  I have made a `get_undirected_interactome()` function in `Dataset.py` that removes these `(u,v)` and `(v,u)` cases.

agitter: I also agree with your suggestion above.  If MEO doesn't support bidirected edges represented as `(u,v)` and `(v,u)` (and other algorithms eventually do), its prepare inputs should be responsible for converting that edge pair into an undirected edge.

annaritz:
This error seems to be in building the docker image for OmicsIntegrator 1.  Will wait for tomorrow to discuss.

```
E: Repository 'http://security.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
```

The code is now in place for writing precision-recall values and generating precision-recall curves (this is in `precrec.py`).  Matplotlib is now a requirement (added to `requirements.yaml`).  Next up is to write snakemake rules to call these functions - there are already pytest tests.

agitter:
> This error seems to be in building the docker image for OmicsIntegrator 1

I tested the fix from https://superuser.com/a/1457007 locally and it solved the error.  I pushed a commit to see if it builds in GitHub Actions successfully.

agitter:
@annaritz #170 implements our new approach to precision-recall calculations. Can we close this old draft pull request once that is merged?

agitter:
Replaced by the merged #170

PR #38, opened 2021-05-27
Add GraphSpace dependencies to conda environment

### LINKED ISSUE #30: Expand conda environment for GraphSpace dependencies
This will update the environment file per https://github.com/Reed-CompBio/spras/pull/27#discussion_r631938938

agitter:
Closes #30

I listed the new packages separately so that we can remove them later if we decide to Dockerize GraphSpace.

@annaritz do you know whether GraphSpace needs version 2.10.0 of the requests package specifically?  Only that version is allowed in https://github.com/adbharadwaj/graphspace-python/blob/42283556eb3a52212cfa9f75dfc694c70069add7/requirements.txt#L3.  That prevents us from installing it through conda.  It is so old it conflicts with other packages in our environment.  It may also be throwing warnings in our test code
```
test/OmicsIntegrator1/test_oi1.py::TestOmicsIntegrator1::test_oi1_required
  /usr/share/miniconda3/envs/spras/lib/python3.8/site-packages/requests/models.py:170: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    if isinstance(hook, collections.Callable):
```

I also split the monolithic test workflow into three jobs that can run in parallel.  Building the Docker images is tested independently from testing the conda environment and running the test code.  The conda environment job also runs in Linux, macOS, and Windows so we can confirm SPRAS can be installed on any platform.  The tests only run in Linux.

### DIFF
--- a/.github/workflows/test-spras.yml
+++ b/.github/workflows/test-spras.yml
@@ -23,7 +23,7 @@ jobs:
 
   # Builds the Docker images
   docker:
-    name: Test Docker wrappers
+    name: Test Docker images
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -69,7 +69,7 @@ jobs:
 
   # Runs the test code and Snakemake workflow
   test:
-    name: Test Docker wrappers
+    name: Run test cases and workflow
     # This job depends on the conda job but not the docker job
     # Therefore the Docker images will be pulled in both the docker job and the test job
     needs: conda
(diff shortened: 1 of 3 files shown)


agitter:
I forked my own version of GraphSpace that specifies a minimum version of requests instead of a specific version (https://github.com/adbharadwaj/graphspace-python/pull/37).  Now requests can be installed through conda and not uninstalled/reinstalled by pip.

PR #41, opened 2021-07-02
Add checkpoints to Snakemake workflow

### LINKED ISSUE #15: Snakemake rule prepare_input runs too many times
This issue was introduced in #12.  The prepare_inputs rule should run once per dataset-algorithm pair and generate multiple output files.  The number of output files depends on the algorithm but is generally >= 2.

Currently, prepare_inputs runs once per output file (dataset-algorithm-output tuple).  The workflow produces the correct files, but the redundant runs are computationally wasteful.

The challenge is that Snakemake cannot statically determine what output files will be produced by this rule.  It is okay to use a function to dynamically determine the input files but not the outputs.  Currently the experimental support for [dynamic files](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#dynamic-files) is being used for this rule.  We can consider alternatives such as [automatically writing](https://stackoverflow.com/questions/48993241/varying-known-number-of-outputs-in-snakemake) one prepare_inputs rule for each algorithm or using [directories as outputs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#directories-as-outputs).

In intermediate version of the code on #12 this wasn't a problem because I was also using the dynamic function to request the output files from the rule.  Maybe adding dynamic on the inputs of the reconstruct rule is another option to explore.

### LINKED ISSUE #17: Rerunning rules only when relevant parts of the config file change
Snakemake reruns the workflow when the config file or the input data files change.  However, ideally the scope would be more narrow.  We should only need to rerun the workflow for a particular dataset if the part of the config file describing that dataset changes.

Allowing unlabeled datasets that are assigned default labels will complicate this.  We may want a labeling scheme that is independent from the other datasets in the config file.

A similar scope can exist for algorithms.  If the config file changes parameters for algorithm A and the datasets all stay the same, we should not need to rerun algorithm B.

agitter:
Closes #15
Closes #17

This pull request moves away from the deprecated `dynamic` feature of Snakemake and uses checkpoints to prepare multiple input files with a single rule.  The rule now produces a directory as output so it only runs once time instead of one time for each input file (e.g. node file, network file).

It also adds checkpoints to have rules depend on only parts of the config file.  Previously, updating any part of the config file would indicate to downstream rules that an upstream file changed, so all rules would be rerun.  Now, the dataset and parameters logfiles are used to cache the last dataset contents and parameter combinations.  Snakemake only needs to rerun an algorithm on a dataset if the algorithm parameters or the dataset contents changed.  Otherwise, the results of those runs can stay the same even if other parts of the config file were changed.  This could help if a user performs a large parameter sweep and then wants to modify the post-processing steps that are performed without rerunning all of the pathway reconstruction steps.

To support file caching, the parameter and dataset logfiles are now written as yaml files.  That makes it easy to read them and compare the cached settings to the current settings in the config file.

Checkpoints seemed to have bugs in the version of Snakemake we were using, so this updates to the latest version 6.5.1.  6.5.0 changed Windows file path behavior, so I now use `SEP` as a hard-coded file separator instead of `os.path.join` in some places.  Snakemake can no longer match file paths correctly on Windows with `os.path.join` or the `os.sep` character.

There are still some minor bugs with the checkpoints.  It is possible to get the output files in a state where Snakemake reruns the `prepare_input` rule every time it is executed even though no input files or config file sections change.  I'll post more instructions for testing once that is resolved.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -138,7 +138,6 @@ checkpoint check_cached_parameter_log:
             print(f'Reusing cached parameters in {logfile}')
 
 # Write the mapping from parameter indices to parameter dictionaries if the parameters changed
-# TODO update reconstruct rule to depend on the parameter log instead of the entire config file
 rule log_parameters:
     # Mark the flag as ancient so that its timestamp is always considered to be older than the output file
     # Therefore, this rule is triggered when the output file is missing but not when the input flag has been updated,
@@ -218,9 +217,15 @@ def collect_prepared_input(wildcards):
     # Check whether prepare_input has been run for these wildcards (dataset-algorithm pair) and run if needed
     # The check is executed by checking whether the prepare_input output exists, which is a directory
     checkpoints.prepare_input.get(**wildcards)
+
+    # The reconstruct rule also depends on the parameters
+    # Add the parameter logfile to the list of inputs so that the reconstruct rule is executed if the parameters change
+    prepared_inputs.append(os.path.join(out_dir, f'parameters-{wildcards.algorithm}.txt'))
     return prepared_inputs
 
 # Run the pathway reconstruction algorithm
+# TODO test that the reconstruct rule is not rerun if the config file changes but the parameters for this algorithm
+# do not (requires setting up dataset configuration caching first)
 rule reconstruct:
     input: collect_prepared_input
     output: pathway_file = os.path.join(out_dir, 'raw-pathway-{dataset}-{algorithm}-{params}.txt')
--- a/Snakefile
+++ b/Snakefile
@@ -7,6 +7,8 @@ from src.analysis.summary import summary
 from src.analysis.viz import graphspace
 
 # TODO decide whether to use os.sep os.altsep or a fixed character for file paths
+# Snakemake updated the behavior in the 6.5.0 release https://github.com/snakemake/snakemake/pull/1037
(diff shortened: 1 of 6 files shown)


agitter:
My changes today remove the parameter and dataset caching.  We discussed how this was either buggy or too difficult to implement reliably with multiple checkpoints.  Now, we rely on more traditional Snakemake file generation to determine whether parts of the workflow need to be rerun.  A key assumption is that if a dataset label does not change, the lists of input files associated with that dataset do not change.  It is okay if the file contents change.

I also switched from numeric parameter indices to shortened parameter hashes (see #43).  I implemented three alternative versions that we can discuss.  The current one is slightly git-like.  It uses sha1 hashes but then base32 encodes them before truncating them to the first 7 positions.  We now get output files like:
- parameters-pathlinker-params-6SWY7JS.yaml
- parameters-pathlinker-params-VQL7BDZ.yaml
- pathway-data0-pathlinker-params-6SWY7JS.txt
- pathway-data0-pathlinker-params-VQL7BDZ.txt

This needs some additional testing and a little cleanup, but I'm much happier with this new simpler direction.

agitter:
This is ready for review.  I left in the redundant multiple version of the hash generation in case we want to discuss the alternatives during review.  Once we pick one, I'll delete the others.

PR #44, opened 2021-07-08
Changes to set config file from command line

melliott432:
Changes were made to Snakefile and util.py so that the config file can be set from the command line using --configfile.

In util.py, parse_config was changed to process_config and no longer takes a file name argument. 

Snakefile now calls process_config and does not use config_file.


Snakefile
```diff
@@ -27,7 +26,8 @@ def get_dataset_dependencies(datasets, label):
     all_files = dataset["node_files"] + dataset["edge_files"] + dataset["other_files"]
     # Add the relative file path and config file
     all_files = [os.path.join(dataset["data_dir"], data_file) for data_file in all_files]
-    return all_files + [config_file]
+    # TODO Need to restore dependency on the config file
```
agitter, line 29: This is complex and related to the changes in #41.  We will work on it in a separate pull request or #41.

src/util.py
```diff
@@ -28,14 +28,12 @@ def prepare_path_docker(orig_path: PurePath) -> str:
     return prepared_path
 
 
-def parse_config(config_file):
+def process_config(config):
     """
-    Parse the config file and return the full yaml structure as well as processed portions
+    Process the dictionary config and return the full yaml structure as well as processed portions
     @param config_file: the path to the config file
```
agitter, line 34: This argument description can be updated.  Now the parameter `config` is the configuration loaded by Snakemake from the config file and any command line arguments.

agitter (changes requested):
The changes so far look good.  There are a few other related edits to the documentation and readme.

- Update the example command in the readme `snakemake --cores 1` to include the config file argument and filename
- Update the test code in `.github/workflows/test-docker-wrappers.yml` so that the last line `snakemake --cores 1` also includes the config file argument and filename

Snakefile
melliott432: Ok, I've added these edits.

agitter (approved):
This looks good to me, and the tests pass.  We can merge after our meeting today if there are no other comments.

Issue #45, opened 2021-07-08
Support using directories that are not subdirectories of the SPRAS directory


melliott432:
The input, output, and config files all have to be in the subdirectories of the SPRAS directory. Any input files have to be copied into the spras input subdirectory. Would it be possible to set a different working directory or use directories that are not SPRAS subdirectories for input, output, and config?

agitter:
After the updates in #49, the container run interfaces now support input, output, and config files in arbitrary locations on the file system.  However, we still need to update and test our dataset loading code if we want to use alternative working directories or absolute paths.   My initial tests with absolute paths on Windows failed.  It may be related to how Snakemake expects path components to be joined in Windows that differs from `os.join`.

tristan-f-r:
While we still need to write tests, #49 did seem to make this work: [the spras-benchmarking CI runs well despite using reconstruction settings in another directory.](https://github.com/Reed-CompBio/spras-benchmarking/pull/22)

Issue #19, opened 2021-05-08
Decide whether to remove secondary pathway reconstruction output files


agitter:
Some pathway reconstruction algorithms output multiple files.  We initially will only track and process a single output.  We need to decide whether to generalize this to support multiple types of output (e.g., optimal and augmented forests from Omics Integrator 1) and what to do with logs and other output files.

agitter:
After #49, all pathway reconstruction output files are written in an output subdirectory that is unique for each dataset-algorithm-parameter combination. We could easily leave any intermediate files in that subdirectory.

tristan-f-r:
Anna, Aden, and I just talked about this issue briefly this morning. Since the intermediary files are in an output subdirectory, and they are useful for anyone looking to do additional processing specific to an algorithm, it seemed like a good idea not to remove these intermediary files.

agitter:
I also favor keeping them. Transparency and ease of debugging have become more important than storage efficiency now that we've used SPRAS more.

PR #49, opened 2021-07-27
Singularity support and Docker improvements

### LINKED ISSUE #11: Redundant docker-py code
Our run functions and Docker wrapper test code have repeated code around the `client.containers.run` call.  Once we better understand these patterns, we can likely move the shared code to a utilities file.

### LINKED ISSUE #22: Docker makes output files as root on linux
This is a known issue which happens with other docker users (there is a good discussion on the issue [here](https://github.com/moby/moby/issues/3206))

When docker runs commands, it runs them as root. Which means that all output files are owned by root. This causes permission errors when Snakemake then tries to touch the files after the reconstruct rule.

Trying to run the docker command as a different user gives permission errors for accessing the omics integrator scripts. 

What I've done for now is added a second docker run command which changes ownership of all output files to the current user, gotten with `os.getuid()` in 5cc19153a36afb12c63ea4942a326ce1d3b96c11. However, this might not work well on other operating systems and requires every new method to use a similar approach. 

There are some more complex solutions which involve creating a new user within the docker container [here](https://vsupalov.com/docker-shared-permissions/) and [here](https://denibertovic.com/posts/handling-permissions-with-docker-volumes/) we could consider.

### LINKED ISSUE #31: multithreading doesn't work
Looks to be an issue of output files being expected when they haven't been generated yet.

### LINKED ISSUE #37: Support Singularity in addition to Docker
We initially selected Docker over Singularity because Docker is easier to install on a local system and has a convenient Python interface.  However, many computing clusters offer Singularity but not Docker.

I'd like to explore whether it is feasible to support both Singularity and Docker and make that configurable via the config file.  The first step would be reading about [Singularity Python](https://singularityhub.github.io/singularity-cli/).

agitter:
This pull request adds the proof of concept support for Singularity as described in https://github.com/Reed-CompBio/spras/issues/37.  It only works with Omics Integrator 1 right now.

At our next meeting, we can discuss how to prioritize Singularity development and if we want to merge this partial support or wait for more mature support.

The tests are failing because the GitHub Actions Ubuntu runners don't have Singularity [installed](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md).

Closes #11, closes #22, closes #31, closes #37

Makes it easier to keep intermediate output files as in #19.  Adds initial support for #45 on the container side.


agitter:
Singularity has been forked, which adds some confusion to what we want to support.  Based on the [announcement](https://apptainer.org/news/community-announcement-20211130/), it sounds like the name "Singularity" is now associated with a commercial version from Sylabs. The open source version has been renamed Apptainer.  The Apptainer [repo](https://github.com/apptainer/apptainer) is still working on their first major release, so I'm going to continue targeting the old version of Singularity in our GitHub Actions workflow for now.

agitter:
I'm expanding the scope of this pull request further so that the abstract container run function can automatically clean up the file owner for files created with Docker.  This requires more granular mapping of volumes.  Originally, the entire `spras` directory was mapped as a volume.  That meant all input and output files had to be within this repository's file tree.  However, when using such a coarse mapping, we can't easily walk the entire directory recursively to check for files that have been added by the Docker run call.

Now, I create a new volume for each input and output file.  This complicates setting up the run call slightly but isolates everything.  SPRAS can check which new files were written in each of those volumes by Docker run.  Hopefully it will then be able to change the owner of those files, but that part isn't implemented.

A side effect is that this should fix #45 as well.  SPRAS should support any path relative to the working directory or any absolute path for input and output files.

This requires more formal testing.  So far I confirmed the new Docker calls work for Omics Integrator 1 locally.

agitter:
I added support for MEO to run in Singularity but the test case failed
```
2022-03-10T22:00:38.7459230Z test/MEO/test_meo.py::TestMaximumEdgeOrientation::test_meo_singularity Running Maximum Edge Orientation with arguments: /spras/QBHSOKD/meo-properties.txt
2022-03-10T22:01:27.0812010Z {'message': 'INFO:    Converting OCI blobs to SIF format\nINFO:    Starting build...\nGetting image source signatures\nCopying blob sha256:d960726af2bec62a87ceb07182f7b94c47be03909077e23d8226658f80b47f87\nCopying blob sha256:e8d62473a22dec9ffef056b2017968a9dc484c8f836fb6d79f46980b309e9138\nCopying blob sha256:8962bc0fad55b13afedfeb6ad5cb06fd065461cf3e1ae4e7aeae5eeb17b179df\nCopying blob sha256:65d943ee54c1fa196b54ab9a6673174c66eea04cfa1a4ac5b0328b74f066a4d9\nCopying blob sha256:da20b77f10accb6bbe77d0a222b481d7a083efa84c2cb9ee752723ad4fca2cf4\nCopying blob sha256:fb6a778e6477757ae1558536d2b6e9c2aff7db2d9dcc743e5cd7eb48000df6a3\nCopying blob sha256:ae7884f0e61bb35dd6f0115f3cf30ba601894c5ced6bd898c12a40b1efeaa807\nCopying blob sha256:872f66d1700e6a1351ac7032e8c877430ae389d77d9b2a1ed5fd2d829aa48f5c\nCopying blob sha256:b5e4c638ece2304dfb1346eeb2269a8c2301a1279b3fb519b05e7946def597b6\nCopying config sha256:71f7f7c0dde65b6eafcd5352c1aa44f9a37a216c388ce71b3e1a8499e405a525\nWriting manifest to image destination\nStoring signatures\n2022/03/10 22:00:45  info unpack layer: sha256:d960726af2bec62a87ceb07182f7b94c47be03909077e23d8226658f80b47f87\n2022/03/10 22:00:47  info unpack layer: sha256:e8d62473a22dec9ffef056b2017968a9dc484c8f836fb6d79f46980b309e9138\n2022/03/10 22:00:47  info unpack layer: sha256:8962bc0fad55b13afedfeb6ad5cb06fd065461cf3e1ae4e7aeae5eeb17b179df\n2022/03/10 22:00:47  info unpack layer: sha256:65d943ee54c1fa196b54ab9a6673174c66eea04cfa1a4ac5b0328b74f066a4d9\n2022/03/10 22:00:49  info unpack layer: sha256:da20b77f10accb6bbe77d0a222b481d7a083efa84c2cb9ee752723ad4fca2cf4\n2022/03/10 22:00:49  info unpack layer: sha256:fb6a778e6477757ae1558536d2b6e9c2aff7db2d9dcc743e5cd7eb48000df6a3\n2022/03/10 22:00:49  info unpack layer: sha256:ae7884f0e61bb35dd6f0115f3cf30ba601894c5ced6bd898c12a40b1efeaa807\n2022/03/10 22:00:51  info unpack layer: sha256:872f66d1700e6a1351ac7032e8c877430ae389d77d9b2a1ed5fd2d829aa48f5c\n2022/03/10 22:00:51  info unpack layer: sha256:b5e4c638ece2304dfb1346eeb2269a8c2301a1279b3fb519b05e7946def597b6\nINFO:    Creating SIF file...\nINFO:    Converting SIF file to temporary sandbox...\nFATAL:   permission denied\nINFO:    Cleaning up image...\n', 'return_code': 255}
2022-03-10T22:01:27.0816822Z FAILED
```

ajshedivy:
Here is the pythonic variation to the switch statement we talked about last week. We just have to rearrange the function definitions and then declare this global dictionary before `run_container`. 
`utily.py`:
```
CONTAINERS = {
    'docker': run_container_docker,
    'singularity': run_container_singularity
}

def run_container(framework: str, container: str, command: List[str], volumes: List[Tuple[PurePath, PurePath]], working_dir: str, environment: str = 'SPRAS=True'):
    """
    Runs a command in the container using Singularity or Docker
    @param framework: singularity or docker
    @param container: name of the DockerHub container without the 'docker://' prefix
    @param command: command to run in the container
    @param volumes: a list of volumes to mount where each item is a (source, destination) tuple
    @param working_dir: the working directory in the container
    @param environment: environment variables to set in the container
    @return: output from Singularity execute or Docker run
    """
    normalized_framework = framework.casefold()

    if framework not in CONTAINERS.keys():
        raise ValueError(f'{framework} is not a recognized container framework. Choose "docker" or "singularity".')
    return CONTAINERS.get(normalized_framework)(container, command, volumes, working_dir, environment)

```

agitter:
Thanks for the dictionary-based switch suggestion.  In this particular function, do you think the dictionary improves readability? I can see how it can declutter code if there are many options in the switch.  However, for this use case I wonder if declaring `CONTAINERS` elsewhere in the file and using the less direct call `CONTAINERS.get(normalized_framework)(container, ...` makes it harder to follow the execution.

Do you have any other suggestions, or should I do a final review myself and then merge?

agitter:
I restored the config file to run all four pathway reconstruction algorithms.  The workflow is failing in GitHub Actions.  It runs for me locally with 1 core but fails with 8 cores.

agitter:
The workflow was failing because of PathLinker.  I upgraded its run function to use the new generalized container run strategy introduced in this pull request.  That initially failed because PathLinker does not have an argument to set the output directory.  However, I was able to use the `--output` argument, which sets an output filename prefix, to specify the directory.

PR #55, opened 2022-04-09
Switch PathLinker image to Alpine Python

agitter:
The updated Dockerfile no longer uses conda to create the environment, which also eliminates the custom ENTRYPOINT.  I'm using the Alpine version of Python, which leads to smaller images.  Our current conda-based image is 303.53 MB.  This one is 21.61 MB.

Another change is that I'm downloading the `requirements.txt` from PathLinker's GitHub repository instead of tracking the required package versions here.  I have to remove the last line from the `requirements.txt` file because it isn't a valid Python package.

If the tests pass, I'll merge this and then sync it with #49.

### DIFF
--- a/docker-wrappers/PathLinker/Dockerfile
+++ b/docker-wrappers/PathLinker/Dockerfile
@@ -1,22 +1,21 @@
 # PathLinker wrapper
 # https://github.com/Murali-group/PathLinker
-# Activates the conda environment before running command inside container
-# Uses the strategy from https://pythonspeed.com/articles/activate-conda-dockerfile/
-# by Itamar Turner-Trauring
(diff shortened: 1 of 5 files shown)


agitter:
I believe MEO tests are failing because the changes in #49 have been pushed to DockerHub.  That's a good reminder that we'll need a better system for versioning our Docker images so that when the image behavior changes older SPRAS code continues to work.

agitter:
I commented out the post-Docker run steps for PathLinker that change files' owner.  Those are not working, but they aren't worth debugging because the implementation will change in #49.  I confirmed the new PathLinker image does have `chown` available.

This is ready to merge.

PR #57, opened 2022-05-14
Add Local Neighborhood algorithm for onboarding new contributors

agitter:
I wrote a simple pathway reconstruction algorithm called Local Neighborhood.  The algorithm takes a network and a list of nodes as input.  It outputs all edges in the network that have a node from the list as an endpoint.

New contributors can practice adding this algorithm to SPRAS.  I added a step-by-step guide to adding it.  I haven't tested the instructions myself, and I'll correct them once I do.  I'd also like to have someone else try to follow the instructions before we merge this.  We should create our implementations in our own forks of the repository so that they aren't as easily seen by new contributors.


agitter:
I followed these new instructions.  It took me less than an hour to complete all the steps, but the tests are failing so it will take longer to fix them.  Because I have full knowledge of the Local Neighborhood algorithm and how all the pieces of SPRAS work together, it would probably take a new contributor many hours to complete all the steps.

Implementing the wrapper functions is the hardest part, especially `generate_inputs`.  We may need better documentation of the `Dataset` class or more functions in that class to make it easier to work with.

Our config file parser doesn't support algorithms that don't have any parameters.  I'll have to fix that.

agitter:
I updated the config file parsing and Snakefile to support algorithms that do not have any parameters.  After that, I was able to fix all the errors in my Local Neighborhood wrapper and get it working.

This is ready for someone else to test the contributing instructions.

annaritz:
Hi @agitter - @nina-young and I seem to be stuck on Step 2.  At first, we just took the Alpine Linux Python image and didn't copy any specific files over. The image ran fine according to the test but I suspect that was because we had `local_neighborhood.py` in the directory we were running the `docker run` command from. 

We changed the Dockerfile to `wget` the `local_neighborhood.py` file from GitHub, so that's now in the Docker image. However, now we're getting an error when testing the `docker run` command with reading in the arguments (the network file doesn't exist). 

Any help is appreciated, we can also wait until our next meeting! Thanks.

agitter:
You're right that you will need to move a copy of `local_neighborhood.py` into the image.  `wget` will work, but I recommend using Docker's `COPY` command instead.  That should make it easier to always get the latest version of the `.py` source every time we build the image.

In Step 1, did you copy the example input files `ln-network.txt` and `ln-nodes.txt` from `test/LocalNeighborhood/input` or provide the relative path to those files?  I'm now realizing that Step 2 assumed you copied those files so that they are present in the directory from which you run `docker run`.  That will need to be updated in the instructions.

nina-young:
@agitter - We have the files copied over, so I don't think that's the issue--we can show you during the meeting on Thursday.

nina-young:
@agitter - We've spent some time trying to figure this bug out and we think it has something to do with the actual binding of /data to the input files (which are copied over). We still get this error: 

```
File "local_neighborhood.py", line 29
    raise OSError(f"Network file {str(network_file)} does not exist")
                                                                   ^
SyntaxError: invalid syntax
```
When we asked local_neighborhood .py to print the working directory and the files inside, it failed to do that when running with Docker (but not when we run it locally without the binding).

agitter:
Let's think about ways for me to reproduce this on my side so I can then try to fix the bug.  Have you pushed the Docker image to DockerHub?  I could pull it and see if I get the same error.  Can you also please post the full `docker run` command you're using?

>it failed to do that when running with Docker

Did you try printing `os.getcwd()` and that failed?  What was the error?

annaritz:
I figured this one out!  Had nothing to do with improperly-mounted files...it has to do with the new string formatting in Python. Note that it's an `invalid syntax` error.  This [stack overflow](https://stackoverflow.com/questions/56875131/syntaxerror-invalid-syntax-when-i-compose-docker-container-but-app-work-fine-w) had the hint in the comments.

If I reverted the string formatting to the `.format()` style, e.g., changing line 28 in `local_neighborhood.py` to

```
OSError("Network file {} does not exist".format(str(network_file)))
```

then the error moves to the next string formatting. If you change all string formatted lines to use the `.format()` then the docker image runs fine.  Both @nina-young and I had this issue, so maybe it's a Mac vs. Windows thing (though why that is I cannot figure out).  For reference, the conda environment has Python `3.8.13` and I'm using Docker version `Docker version 20.10.12, build e91ed57`.

agitter:
Nice find!  Do you have the Dockerfile you used in a branch that I could inspect?  I'm curious what version of Python is used in that Dockerfile.  I would expect that any Python >= 3.6 would support this f-string style formatting.

annaritz:
Yep - here it is: https://hub.docker.com/repository/docker/annaritz/local-neighborhood 

Aha! The Dockerfile (copied from PathLinker) has 
```
FROM python:3.5.10-alpine
```

agitter:
4a46b9b adds a note to help clarify the difference between the PathLinker example and the Local Neighborhood image.  Is the phrasing okay?

ntalluri:
When testing the docker file for step two of the contribution guide, clarifying what each part of the command does made setting up the command much easier and understandable. (An image/picture of what is happening is also very helpful.)

ntalluri:
Based on what happened to me when making my docker file, maybe mention that if a person is using or-tools, they shouldn't use alpine

agitter:
Thanks everyone who tested these instructions and gave feedback. I made edits to the contributing guide today in response to the comments above and am going to merge the current version. I'm happy to make more edits if there are more suggestions but would like a baseline version in place as new team members join to work on SPRAS this semester.

PR #61, opened 2022-06-10
Fix bug in Omics Integrator 1 TMPDIR

### LINKED ISSUE #60: Omics Integrator 1 "No space left" error is back
Running Omics Integrator 1 on Singularity with a large network I'm getting the error:

```
Traceback (most recent call last):
  File "/OmicsIntegrator/scripts/forest.py", line 1464, in <module>
    main()
  File "/OmicsIntegrator/scripts/forest.py", line 1424, in main
    (edgeList, info) = inputObj.runPCSF(options.msgpath, options.seed)
  File "/OmicsIntegrator/scripts/forest.py", line 576, in runPCSF
    1-float(self.undirEdges[edgeNode1][edgeNode2])))
IOError: [Errno 28] No space left on device
```

This is in the same environment as the last time we saw this issue https://github.com/Reed-CompBio/spras/issues/37#issuecomment-892714685 but I'm not certain if it's the exact same problem (I don't yet have a full grasp over which volume is mapped where).

PS. this error is printed out but what actually halts the pipeline is failure to rename the consequently non-existent output file, maybe the reconstruct rule should explicitly fail if OI1 has a nonzero return code?

agitter:
Closes #60

This bug was introduced in #49.  Previously volumes were mapped such that `/OmicsIntegrator1` inside the container corresponded to the root of the `spras` repository on the local filesystem, which made it a suitable temporary directory for Omics Integrator 1.  That is no longer the case.

Now the temporary directory is set to the output directory for this Omics Integrator 1 run.  I was able to reproduce #60, and this change fixed it for me.

### DIFF
--- a/src/omicsintegrator1.py
+++ b/src/omicsintegrator1.py
@@ -148,7 +148,7 @@ class OmicsIntegrator1(PRM):
                             command,
                             volumes,
                             work_dir,
-                            'TMPDIR=/OmicsIntegrator1')
+                            f'TMPDIR={mapped_out_dir}')
         print(out)
 
         conf_file_local.unlink(missing_ok=True)


sverchkov:
Confirming that this fixes https://github.com/Reed-CompBio/spras/issues/60

agitter:
Thanks for the bug report and testing @sverchkov

PR #66, opened 2022-06-27
Cytoscape

ajshedivy:
- add py4cytoscape container 
- add new cytoscape visualization rule

Let's focus on creating a cytoscape rule that can grab all of the generated pathways. After that I can continue work on creating the cytoscape session using docker. Right now, local support of cytoscape is integrated, make sure to start cytoscape, then run the workflow. 

#### Links 
- py4cytoscape examples: https://github.com/cytoscape/cytoscape-automation
- py4cytoscape docs: https://py4cytoscape.readthedocs.io/en/latest/index.html


Snakefile
```diff
@@ -219,6 +228,9 @@ rule viz_graphspace:
     run:
         graphspace.write_json(input.standardized_file,output.graph_json,output.style_json,directed=algorithm_directed[wildcards.algorithm])
 
+        # just run local cytoscape with graphspace for now
+        cytoscape.viz_cytoscape_local(input.standardized_file, directed=algorithm_directed[wildcards.algorithm])
```
ajshedivy, line 232: comment out this line to return to normal workflow behavior

agitter:
Thanks for working on this.  Cytoscape visualization will be a great addition.  I'll work on input for the Snakefile rule to gather all of the pathway files.

The design here can be streamlined compared to what we set up for TPS because Cytoscape will always run inside the container.  All we need on the SPRAS side is an interface to pass the list of files (and any annotations) to the container.  We should also try to avoid conda inside the container and install the required packages with pip.  That will give us a chance of creating a container that works with Singularity.

Several files in this pull request will be deleted later so we should squash and merge when it's time.

agitter:
We need to decide which pathways should be visualized in the Cytoscape session.  Should it be the pathways that correspond to the parameters in the current config file (current behavior).  Or all pathways discovered in the output directory, including some that may be left over from previous runs with parameters that were removed from the config file?

PRRunner.py
```diff
@@ -1,11 +1,13 @@
 import Dataset
-
+import os
```
agitter, line 2: Can the changes to this file be reverted?

config/config_cyto.yaml
```diff
@@ -0,0 +1,113 @@
+ # Global workflow control
```
agitter, line 1: Let's add the Cytoscape step to `config/config.yaml` and remove this additional config file.

config/egfr.yaml
```diff
@@ -0,0 +1,84 @@
+# The length of the hash used to identify a parameter combination
```
agitter, line 1: I'm not sure how the files from the EGFR update were added to this branch as new files.  You can delete this and the two new input files on this branch.

src/analysis/cytoscape/cytoscape.py
```diff
@@ -0,0 +1,50 @@
+from pathlib import PurePath
```
agitter, line 1: Let' s move this file up one level so the path is `src/analysis/cytoscape.py`  Otherwise it will be the only file in the `cytoscape` subdirectory.

src/analysis/cytoscape/cytoscape.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import List, Union
+from pathlib import PurePath, Path
+import os
+import docker
+
+def run_cytoscape_container(pathways: List[Union[str, PurePath]], out_dir: str) -> None:
```
agitter, line 7: @ajshedivy and I discussed that we will first get this working reliably with Docker.  Then we'll work on generalizing it to use the container running utility to see if it can run in Singularity as well.

src/analysis/cytoscape/py4cy/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:3.9.13
+RUN pip install py4cytoscape==1.3.0
+
+# PARAMETERS
```
agitter, line 4: Is some of this taken from the Cytoscape example?  Do we need to provide attribution?

src/analysis/cytoscape/py4cy/cytoscape_util.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import List, Dict, Any
+from numpy import append
+import py4cytoscape as p4c
+import time
+import sys
+import os
+from requests.exceptions import RequestException
+import subprocess
+
+SPRAS_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
```
agitter, line 13: Why do we need the SPRAS root directory if this is now running inside the container?

src/analysis/cytoscape/py4cy/cytoscape_util.py
```diff
@@ -0,0 +1,98 @@
...
+    '''
+    Initialize a parser and use it to parse the command line arguments
+    :return: parsed dictionary of command line arguments
+    '''
+    parser = get_parser()
+    opts = parser.parse_args()
+
+    return opts
+
+def get_pathway_files():
```
agitter, line 45: Is this left over from when it ran as a local demo?

src/analysis/cytoscape/py4cy/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    && mv cytoscape3.props.tmp cytoscape3.props \
+    && jar -uf property-impl-${CYTOSCAPE_VERSION}.jar cytoscape3.props \
+    && rm cytoscape3.props \
+    && cd /
+# Set JAVA_HOME From sudo update-alternatives --config java
+RUN echo 'JAVA_HOME="/usr/lib/jvm/java-11-openjdk-amd64"' >> /etc/environment
+
+COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+COPY cytoscape_util.py /spras/cytoscape_util.py
+CMD ["/usr/bin/supervisord"]
```
agitter, line 26: I'll have to test how this works in Singularity.

src/parse_config.py
```diff
@@ -0,0 +1,127 @@
+
+import base64
```
agitter, line 2: Was anything changed in this file?

src/util.py
```diff
@@ -17,7 +17,6 @@
 # The default length of the truncated hash used to identify parameter combinations
 DEFAULT_HASH_LENGTH = 7
 
-
```
agitter, line 20: Was anything changed in this file?

agitter (changes requested):
I'm able to run the workflow now locally with Docker and generate the Cytoscape session file.  It's very cool to have it working.  I did get what looks like several error messages from the cyrest client that I haven't looked at closely.

All of the Docker-related files can be moved to `docker-wrappers/Cytoscape` from `src/analysis/py4cy`.  Before we merge, please add a readme describing this image.  We'll also want to switch to the reedcompbio DockerHub organization.

Can we delete the two empty `__init__.py` files?

src/analysis/cytoscape/cytoscape.py
ajshedivy: done

config/config_cyto.yaml
ajshedivy: done

src/analysis/cytoscape/py4cy/cytoscape_util.py
ajshedivy: yep, I removed this function

src/util.py
ajshedivy: This file was somehow pulled into my branch, it can be removed

agitter:
The workflow runs to completion when the Cytoscape step is skipped, which is good because earlier it looked like it might be failing at one of the PathLinker runs.  I now suspect the issue is #22 and that the output .cys file is owned by root.  Switching to the `run_container` framework will confirm that.

agitter:
I pushed some updates to these Cytoscape files today. Some were minor code cleanup and reorganization changes.  I converted the container wrapper to use the `run_container` utility function as a step toward Singularity testing.  However, the volume mapping needed for this breaks the pathway names that are shown in the Cytoscape session.  Instead of names like `output/data0-pathlinker-params-VQL7BDZ/pathway.txt` now we have names like `/spras/NLZ6AAF/pathway.txt`.  That may mean the original filename needs to be passed to `cytoscape_util.py` as another command line argument.  Then we can call [`rename_network`](https://py4cytoscape.readthedocs.io/en/1.3.0/reference/generated/py4cytoscape.networks.rename_network.html).

I was able to simplify the Docker call somewhat, which makes me more optimistic about Singularity support.  The `CMD` isn't needed in the `Dockerfile`, nor is the port in the Docker run call.

There are a few other open TODOs:
- Add a readme with an explanation and attribution
- Test Singularity
- Fix the filename issue described above
- Move the Docker image to the reedcompbio organization

agitter:
I pushed [reedcompbio/py4cytoscape](https://hub.docker.com/repository/docker/reedcompbio/py4cytoscape) to DockerHub so we can test and maintain that version.

ajshedivy:
I was able to get the workflow to run with the new docker image and the changes to the cytoscape backend components.

agitter:
I enabled Singularity for the Cytoscape workflow step, but the initial test on a Linux server failed with this message
```
{'message': 'INFO:    Converting OCI blobs to SIF format\nINFO:    Starting build...\nGetting image source signatures\nCopying blob sha256:1671565cc8df8c365c9b661d3fbc164e73d01f1b0430c6179588428f99a9da2e\nCopying blob sha256:3e94d13e55e7a4ef17ff21376f57fb95c7e1706931f8704aa99260968d81f6e4\nCopying blob sha256:fa9c7528c685216129e8e67bf362a7702e7b1daa585ab85546a41508830657d6\nCopying blob sha256:53ad072f9cd16fc8eb93b182b20e758e11acc0ef60babef0bf1043c08de1901a\nCopying blob sha256:d6b983117533b718374f1701ef593dd2afa6613c7908c6553be8e2a150e6448a\nCopying blob sha256:d8092d56ded5476fe7c302256eb4dc6ff495ae8fb4dd28aa18dbcb7581e24a6c\nCopying blob sha256:5b016815b81d6117f6103cf8208d808225269972b47471e5ac111f22daeca9c4\nCopying blob sha256:c9fb4e3b48645861745962ed654e9f2ca9012df7b0d404851fec0af9a28322d1\nCopying blob sha256:18aa057f725479f7c28df59530194198c84516c376b5b450274839aab42b7e4c\nCopying blob sha256:6f17e57f1088f1eb513375db3e08a31bef7d4ed041dac646aa41337daed958e1\nCopying blob sha256:53ff195f8cf2fa1324d738f801527e2b11f9de96deda2c6b7285f56f65c99335\nCopying blob sha256:edeae988115b89239987e681f0e591e078080b2288949d8d42e0627dd4f5cc0f\nCopying blob sha256:db90e8058058a36d649856670947d1c804859898cbfe3d02e1786b4e007aa727\nCopying blob sha256:87723be63b0926b0479f8e10fddcf08f22a163450025d7476c6dfed108a2000c\nCopying blob sha256:bf37aa2207900f003fe8f7935d5e6eb9cd00df8cf145b28928ce4b9747f0f743\nCopying blob sha256:5d5ac16a7bbae7a8687ae675f2f3d73530a3e8c936bfafc17dd784bbb296239f\nFATAL:   Unable to handle docker://reedcompbio/py4cytoscape uri: while building SIF from layers: conveyor failed to get: initializing source oci:/ua/gitter/.singularity/cache/blob:e991b7e0217edc0f234bba2c1909204b21268e0ccc9932d0b261b038111d5732: writing blob: write /ua/gitter/.singularity/cache/blob/oci-put-blob188976184: disk quota exceeded\n', 'return_code': 255}
```

I'll need to see if "disk quota exceeded" refers to the server I'm using or the Singularity container's disk.

agitter:
The Singularity error above was due to a full disk quota on the server.  I was able to convert the Docker image to Singularity after removing files.

However, py4ctytoscape creates a default logfile and the path is [hard-coded](https://github.com/cytoscape/py4cytoscape/blob/cc3186c722f2789460b90547105d7ffa87d0460f/py4cytoscape/py4cytoscape_logger_settings.py#L34). That caused the workflow to fail with Singularity because the container cannot write to that path.  I believe I can create a workaround for that by mapping the path to the hard-coded log file to a path in the output directory in the local file system.  When I tried that, however, py4cytoscape hung for over 10 minutes trying to connect to Cytoscape.

agitter:
Another user is doing some large scale image export that may be helpful in https://github.com/cytoscape/py4cytoscape/issues/81.

I created an issue asking about the path of the log file https://github.com/cytoscape/py4cytoscape/issues/95.  One of the developers proposed making this customizable with an environment variable, which would be great.

I ran the command in an interactive Singularity session, which may have revealed why the connection fails
```
  File "/usr/lib/python3/dist-packages/supervisor/loggers.py", line 160, in __init__
    self.stream = open(filename, mode)
OSError: [Errno 30] Read-only file system: '/spras/supervisord.log'
An error has occurred while trying to run Cytoscape
Command '['/usr/bin/supervisord', '-c', '/etc/supervisor/conf.d/supervisord.conf']' returned non-zero exit status 1.
In cyrest_get: HTTPConnectionPool(host='127.0.0.1', port=1234): Max retries exceeded with url: /v1/version (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fa422209310>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

Running with `--writable-tmpfs` fixes that, but I still can't connect to Cytoscape. I'm wondering if the supervisord startup command is failing silently. The key environment variables are set inside the container:
```
env
SHELL=/bin/bash
PY4CYTOSCAPE_VERSION=1.3.0
PYTHON_VERSION=3.9.13
SINGULARITY_NAME=e991b7e0217edc0f234bba2c1909204b21268e0ccc9932d0b261b038111d5732
SINGULARITY_ENVIRONMENT=/.singularity.d/env/91-environment.sh
PWD=/spras
CYTOSCAPE_VERSION=3.9.1
PYTHON_SETUPTOOLS_VERSION=58.1.0
HOME=/ua/gitter
LANG=C.UTF-8
SINGULARITY_CONTAINER=/ua/gitter/.singularity/cache/oci-tmp/e991b7e0217edc0f234bba2c1909204b21268e0ccc9932d0b261b038111d5732
GPG_KEY=E3FF2839C048B25C084DEBE9B26995E310250568
TERM=xterm
SPRAS=True
SHLVL=1
SINGULARITY_BIND=output:/spras/logs,/ua/gitter/spras/output:/spras/SQX5E3U/output,/ua/gitter/spras/output/data0-pathlinker-params-VQL7BDZ:/spras/NLZ6AAF,/ua/gitter/spras/output/data0-pathlinker-params-6SWY7JS:/spras/B2PWF4V,/ua/gitter/spras/output/data0-omicsintegrator1-params-PU62FNV:/spras/KQDOF7B,/ua/gitter/spras/output/data0-omicsintegrator1-params-WY4V42C:/spras/CWEDO6Y,/ua/gitter/spras/output/data0-omicsintegrator1-params-JAZWLAK:/spras/Z5WAUEO,/ua/gitter/spras/output/data0-omicsintegrator1-params-RQCQ4YN:/spras/G4LXC7S,/ua/gitter/spras/output/data0-meo-params-GKEDDFZ:/spras/HPUTSGP,/ua/gitter/spras/output/data1-pathlinker-params-VQL7BDZ:/spras/HCAEDGP,/ua/gitter/spras/output/data1-pathlinker-params-6SWY7JS:/spras/SN752BH,/ua/gitter/spras/output/data1-omicsintegrator1-params-PU62FNV:/spras/FQ7BJ7O,/ua/gitter/spras/output/data1-omicsintegrator1-params-WY4V42C:/spras/TVZGPNE,/ua/gitter/spras/output/data1-omicsintegrator1-params-JAZWLAK:/spras/SKHRAC2,/ua/gitter/spras/output/data1-omicsintegrator1-params-RQCQ4YN:/spras/SBBNONI,/ua/gitter/spras/output/data1-meo-params-GKEDDFZ:/spras/7JW7OWC
PYTHON_PIP_VERSION=22.0.4
LD_LIBRARY_PATH=/.singularity.d/libs
PYTHON_GET_PIP_SHA256=5aefe6ade911d997af080b315ebcb7f882212d070465df544e1175ac2be519b4
PS1=Singularity>
PYTHON_GET_PIP_URL=https://github.com/pypa/get-pip/raw/5eaac1050023df1f5c98b173b248c260023f2278/public/get-pip.py
PATH=/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
SINGULARITY_COMMAND=shell
```

`JAVA_HOME` isn't set, which Cytoscape warns about if I try to start it directly without supervisord.  We can fix that in the Dockerfile.

agitter:
After the recent commits stripped away unnecessary parts of the supervisord configuration and Docker image, I tested again in Singularity. Cytoscape still failed interactively, but I'm getting closer to the real root case:
```
> /bin/bash -c 'xvfb-run -s "-screen 0 1920x1080x24" /py4cytoscape/cytoscape-unix-3.9.1/cytoscape.sh'
/py4cytoscape/cytoscape-unix-3.9.1/gen_vmoptions.sh: 50: cannot create /py4cytoscape/cytoscape-unix-3.9.1/Cytoscape.vmoptions: Permission denied
/py4cytoscape/cytoscape-unix-3.9.1/gen_vmoptions.sh: 51: cannot create /py4cytoscape/cytoscape-unix-3.9.1/Cytoscape.vmoptions: Permission denied
/py4cytoscape/cytoscape-unix-3.9.1/gen_vmoptions.sh: 52: cannot create /py4cytoscape/cytoscape-unix-3.9.1/Cytoscape.vmoptions: Permission denied
*** Missing Cytoscape.vmoptions, falling back to using defaults!
karaf.base: /py4cytoscape/cytoscape-unix-3.9.1/framework
JAVA_HOME: /usr/lib/jvm/java-11-openjdk-amd64
Unable to update instance pid: Unable to create directory /py4cytoscape/cytoscape-unix-3.9.1/framework/instances
org.osgi.framework.BundleException: Unable to cache bundle: mvn:org.apache.servicemix.specs/org.apache.servicemix.specs.jaxb-api-2.1/1.2.0
        at org.apache.felix.framework.Felix.installBundle(Felix.java:3013)
        at org.apache.felix.framework.BundleContextImpl.installBundle(BundleContextImpl.java:167)
        at org.eclipse.equinox.internal.region.BundleIdBasedRegion.installBundle0(BundleIdBasedRegion.java:117)
        at org.eclipse.equinox.internal.region.BundleIdBasedRegion.installBundleAtLocation(BundleIdBasedRegion.java:97)
        at org.apache.karaf.features.internal.service.BundleInstallSupportImpl.installBundle(BundleInstallSupportImpl.java:135)
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.installBundle(FeaturesServiceImpl.java:1138)
        at org.apache.karaf.features.internal.service.Deployer.deploy(Deployer.java:900)
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.doProvision(FeaturesServiceImpl.java:1062)
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.lambda$doProvisionInThread$13(FeaturesServiceImpl.java:998)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.io.IOException: No space left on device
        at java.base/java.io.FileOutputStream.writeBytes(Native Method)
        at java.base/java.io.FileOutputStream.write(FileOutputStream.java:354)
        at java.base/java.io.BufferedOutputStream.write(BufferedOutputStream.java:123)
        at org.apache.felix.framework.cache.BundleCache.copyStreamToFile(BundleCache.java:359)
        at org.apache.felix.framework.cache.JarRevision.initialize(JarRevision.java:170)
        at org.apache.felix.framework.cache.JarRevision.<init>(JarRevision.java:77)
        at org.apache.felix.framework.cache.BundleArchive.createRevisionFromLocation(BundleArchive.java:872)
        at org.apache.felix.framework.cache.BundleArchive.reviseInternal(BundleArchive.java:550)
        at org.apache.felix.framework.cache.BundleArchive.<init>(BundleArchive.java:153)
        at org.apache.felix.framework.cache.BundleCache.create(BundleCache.java:277)
        at org.apache.felix.framework.Felix.installBundle(Felix.java:3009)
        ... 12 more
```

My interpterion of these errors is that the `/py4cytoscape` directory where I installed Cytoscape is owned by root, and Cytoscape tries to write several files there during startup. The user in the linked issue above reported successfully creating a Singularity image at https://github.com/EnzymeFunctionInitiative/cytoscape-util/tree/dev-local-py4cytoscape so that may provide some inspiration. It doesn't guarantee that converting a Docker image to Singularity will work.

agitter:
I'm now manually creating the Cytoscape vmoptions file and adding it to the image. I ran the vmoptions generation script inside an interactive container session and modified the memory settings and set the home directory to point to the writeable `/spras` working directory. These Cytoscape [documents](https://manual.cytoscape.org/en/latest/Launching_Cytoscape.html#cytoscape-directories) catalog relevant installation and configuration directories that are used during startup, several of which need to be writeable. 

This [post](https://stackoverflow.com/questions/74538194/network-install-of-cytoscape) states that the `Unable to update instance pid` is not serious and can be ignored. However, I'm still getting the errors that follow, which are fatal. `No space left on device` could have something to do with how Singularity manages disk or temporary files inside the container. I'm still using Singularity's `--writable-tmpfs` argument for my interactive testing.

agitter:
I posted to the Cytoscape Google group about the Singularity problems I'm having and received a response from one of the developers: https://groups.google.com/g/cytoscape-helpdesk/c/NNC1Bj8fcPA/m/4HOlPmvtAQAJ

That gives me some hope for continuing to debug this. I'll do more interactive Singularity testing based on the feedback.

agitter:
Through interactive testing in Docker and Singularity and support from a developer through the Google group, I figured out the Singularity "No space left on device" error. I also believe I have a workaround.

The error was due to Cytoscape not using the `-Duser.home` setting in its vmoptions file for all of the configuration files it downloads. Some configuration bundles are downloaded to a configuration directory that is hard-coded as a subdirectory of `$HOME`. The containers run as root, so `$HOME=/root` which is okay in Docker and not in Singularity.

I was eventually able to get Singularity to recognize a different home directory. Now Cytoscape runs interactively with Singularity in my converted Docker image. I also fixed a Cytoscape bug, which I believe is new, described [here](https://cytoscape.org/common_issues.html#zipvalidation) using their recommended workaround.

Despite that, the workflow still hung at the Cytoscape rule in Singularity.

There are still some final changes to make after the workflow runs:
- Consider the latest py4cytoscape package that supports configurable log directories, although it may be more work than expected to use those now that we need to set the home directory with an environment variable. Our container framework only supports a single environment variable per call for now.
- Clean up output files and log files
- Write one Cytoscape session file per dataset

agitter:
This works in Singularity now! I also refactored it to produce one Cytoscape session per dataset instead of combining all pathways into a single dataset. I write the temporary Cytoscape logs to a dataset-specific output directory and then remove it after Cytoscape finishes, so there shouldn't be any problems with parallelism.

I'd like to look over all the code once more and then will ask someone else to test that it works before merging.

agitter:
I added test cases to run Cytoscape in Docker and Singularity. If those pass, I'll tag the image as `v1` and this should be ready to merge.

@ntalluri could you also please try testing the workflow on your machine in Docker to confirm you get session files and no errors before I merge?

agitter:
The `test_cytoscape_singularity` test that fails in GitHub actions passes in the Apptainer environment I have access to locally:
```
$ pytest -k test_cytoscape_singularity
====================================================================================== test session starts =======================================================================================
platform linux -- Python 3.8.16, pytest-7.1.3, pluggy-1.0.0
rootdir: /ua/gitter/spras
collected 67 items / 66 deselected / 1 selected

test/analysis/test_cytoscape.py .                                                                                                                                                          [100%]

======================================================================================== warnings summary ========================================================================================
../miniconda3/envs/spras/lib/python3.8/site-packages/future/standard_library/__init__.py:65
  /ua/gitter/miniconda3/envs/spras/lib/python3.8/site-packages/future/standard_library/__init__.py:65: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================================================== 1 passed, 66 deselected, 1 warning in 162.83s (0:02:42) =====================================================================
```

My test server has Apptainer 1.2.2 whereas GitHub Actions is using 1.1.3.

```
$ apptainer --version
apptainer version 1.2.2-1.el7
```

The [1.2.0 release](https://github.com/apptainer/apptainer/releases/tag/v1.2.0) introduced some important behavioral changes so I'll update what we run in GitHub Actions to see if that is causing the test failure.

agitter (approved):

agitter:
All tests pass (after marking the Cytoscape Singularity test xfail) and I tagged the image as v1.

@ajshedivy I'm merging in the SPRAS Cytoscape pull request!

Issue #46, opened 2021-07-08
Modifications to summary files


melliott432:
In the output, each returned network has a corresponding summary file with a table that reports degree, clustering, and centrality for each node. It may be helpful to sort these tables by degree or centrality so that nodes with high degree or high centrality can be easily identified. It also may be helpful to add a single summary file of all the networks that reports basic statistics such as number of nodes, diameter, number of edges, number of componenets, etc, so that output networks can be more readily compared.

PR #68, opened 2022-08-03
Summary

jhaddad2:
Added summarization function and test cases/files


src/analysis/summary/summary.py
```diff
@@ -1,9 +1,62 @@
...
 import os
+import pandas as pd
 #import json
 #wrapper functions for nx methods here
 
+def summarize_networks(file_paths, node_table):
+    # Initialize list to store input nodes that have property data
+    nodes_by_col = []
+    # Save new labels
+    nodes_by_col_labs = ("nodes in " + node_table.columns[1:]).tolist()
```
agitter, line 12: Let's capitalize the first letter of each column header entry ("Nodes in...")

src/analysis/summary/summary.py
```diff
@@ -1,9 +1,62 @@
...
+        nw_info.append(cur_nw_info)
+
+    # Convert the network summary data to pandas dataframe
+    nw_info = pd.DataFrame(
+        nw_info,
+        columns = [
+            "name",
+            "number of nodes",
+            "number of edges",
+            "number of connected components"
```
agitter, line 53: Capitalize these as well

test/analysis/test_summary.py
```diff
@@ -0,0 +1,31 @@
+# Run from SPRAS root with python -m test.analysis.test_summary
+
+import pandas as pd
```
agitter, line 3: I'll first run these tests locally and then help convert them to pytest test cases.

agitter (commented):
This looks great @jhaddad2. I'm leaving a few initial comments. The next major steps will be for me to run your test cases locally to confirm the behavior matches our expectations and then add a rule to the Snakefile based on the Cytoscape rule in #66 so that the summarization is run as part of the workflow.

test/analysis/test_summary.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Make output directory
+if not os.path.isdir("test/analysis/output"):
+	os.makedirs("test/analysis/output")
+
+# Toy networks
+toy_network_files = glob("test/analysis/input/toy/*")
+toy_node_table = pd.read_csv("test/analysis/input/toy_node_table.csv")
+summarize_networks(toy_network_files, toy_node_table).to_csv("test/analysis/output/toy_summary.csv", index = False)
+
```
agitter, line 21: Instead of writing the summary table to csv, here you can compare the generated table to the stored expected table that you read from disk.  You can use `assert` when comparing the two data frames.

We'll also convert from comma to tab separated.

agitter:
@annaritz we are getting close to merging this new feature that summarizes all pathways generated for a single dataset.  The example file `test/analysis/output/egfr_summary.txt` demonstrates the output.  We wanted to keep the table columns simple for now and can easily extend the network attributes later.  The workflow now generates files with names like `data0-pathway-summary.txt` if a summary is requested.

Let us know if you have any thoughts about this feature as we wrap up the final code review.

I'm also noting that this pull request will partially address #46.

test/analysis/test_summary.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+from glob import glob
+from src.analysis.summary.summary import summarize_networks
+
+# Notes:
+# - Column labels are required
+# - 'NODEID' is required as the first column label
+# - file_paths must be a list, even if a single file path is passed
+
+class TestSummary:
```
agitter, line 12: These tests pass in GitHub Actions but fail when I test them locally.  I wonder if it could be because of the file path differences across operating systems?  We should decide whether to make the comparison to the expected summary file more robust or skip this test on Windows, which we already do elsewhere in the SPRAS test suite.

agitter:
7d7e953 fixed the file path issues that caused the tests to fail locally on Windows.

I did another round of code review, and it is looking good to me.  I manually computed some of the summary statistics for the smaller toy networks to confirm the matched what the table reports.  I did notice that the toy network 1 has more edges than the table reports because it has A-B and B-A.  We work with undirected graphs, so the behavior is correct.  However, I wanted to make this more explicit, so I updated the column header in the summary table in 57ad81e.

agitter (approved):

PR #70, opened 2022-09-16
Add min cost flow pathway reconstruction algorithm

ntalluri:


config/config.yaml
```diff
@@ -28,14 +28,14 @@
  algorithms:
         - name: "pathlinker"
           params:
-                include: true
+                include: false
```
agitter, line 31: We'll change these back once we are done testing min cost flow

config/config.yaml
```diff
@@ -56,12 +56,21 @@
...
                     max_path_length: [3]
                     local_search: ["Yes"]
                     rand_restarts: [10]
+      
+        - name: "mincostflow"
+          params:
+                include: true
+                directed: true
+                run1:
+                    flow: [1]
```
agitter, line 71: Please add a comment in this line that `flow` needs to be an int

docker-wrappers/MinCostFlow/Dockerfile
```diff
@@ -0,0 +1,8 @@
+#uses ubuntu as the base
+FROM python:3.10.7
+
+WORKDIR /MinCostFlow
+
+RUN wget https://raw.githubusercontent.com/gitter-lab/min-cost-flow/ea1added536d422b5ead7ba544af4e6fe0532210/minCostFlow.py
+
+RUN pip install ortools==9.4.1874
```
agitter, line 8: Let's switch the order of the two `RUN` commands to get better caching behavior.  We are more likely to update the version of `minCostFlow.py` frequently, and currently we will need to reinstall `ortools` every time we download a new version of the `.py` file.  If we reorder them, we can keep `ortools` cached.

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
+from codeop import CommandCompiler
```
agitter, line 1: Remove this

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+            # take nodes one column data frame, call sources/ target series/ vector
+            nodes = nodes.loc[nodes[node_type]]
+            # creates with the node type without headers 
+            nodes.to_csv(filename_map[node_type], index=False, columns=['NODEID'], header=False)
+
+        # create the network of edges 
+        edges = data.get_interactome()
+        
+        # creates the edges files that contains the head and tail nodes and the weights after them
+        edges.to_csv(filename_map['edges'], sep='\t',index=False, columns = ["Interactor1","Interactor2","Weight"], header=False) #removed columns and now it works columns = [Node1, Node2, Weight]
```
agitter, line 39: I believe we can remove the comment at the end of this line

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @staticmethod
+    def run (sources = None, targets = None, edges= None, output_file = None, flow = None, capacity = None, singularity=False):
+        """
+        Run min cost flow with Docker (or singularity)
+        @param sources:  input sources (required)
+        @param targets: input targets (required)
+        @param edges:  input network file (required)
+        @param output: output file name (required)
```
agitter, line 48:
```suggestion
        @param output_file: output file name (required)
```

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+        @param flow: amount of flow going through the graph (optional)
+        @capacity: amount of capacity allowed on each edge (optional)
```
agitter, line 50:
```suggestion
        @param capacity: amount of capacity allowed on each edge (optional)
```

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+        @param singularity: if True, run using the Singularity container instead of the Docker container
+        """
+
+        # ensures that these parameters are required
+        if not sources or not targets or not edges or not output_file:
+            raise ValueError('Required PathLinker arguments are missing')
```
agitter, line 56:
```suggestion
            raise ValueError('Required MinCostFlow arguments are missing')
```

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # the data files will be mapped within this directory within the container
+        work_dir = '/mincostflow'
+
+        # the tuple is for mapping the sources, targets, edges, and output 
+        volumes = list()
+
+        # required arguments 
+        bind_path, sources_file = prepare_volume(sources, work_dir) # sources_file = workingdirectory/joiefhowihfj/sources.txt (which is sources)
```
agitter, line 65: Can remove the comment

src/mincostflow.py
```diff
@@ -0,0 +1,118 @@
...
+            command.extend (['--flow', str(flow)])
+        if capacity is not None:
+            command.extend (['--capacity', str(capacity)])
+
+        # choosing to run in docker or sigularity container
+        container_framework = 'singularity' if singularity else 'docker'
+
+        # constructs a docker run call 
+        out = run_container(container_framework,
+                            'ntalluri2/mincostflow',
```
agitter, line 99: We can update this after I push the Docker image to the reedcompbio Docker organization

test/MinCostFlow/test_mcf.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+import shutil
+from pathlib import Path
+from src.mincostflow import MinCostFlow
+from test.MEO.test_meo import OUT_FILE
```
agitter, line 5: Can we delete this line?

test/MinCostFlow/test_mcf.py
```diff
@@ -0,0 +1,62 @@
...
+
+TEST_DIR = 'test/MinCostFlow/'
+OUT_FILE = TEST_DIR+'output/mincostflow-output.txt'
+
+class TestMinCostFlow:
+    def test_mincostflow_required(self):
```
agitter, line 11: Do you have the expected output file for these test cases? We could commit that as well to `test/MinCostFlow/expected/mincostflow-output.sif` even if we aren't using it yet. We may need to encode the flow and capacity in the filename so we can match it with the respective inputs.

test/MinCostFlow/test_mcf.py
```diff
@@ -0,0 +1,62 @@
...
+        out_path = Path(OUT_FILE)
+        out_path.unlink(missing_ok=True)
+
+        MinCostFlow.run(sources=TEST_DIR + 'input/sources.txt',
+            targets=TEST_DIR + 'input/targets.txt',
+            edges=TEST_DIR + 'input/edges.txt',
+            output_file=OUT_FILE)
+        assert out_path.exists()
+
+    def test_mincostflow_missing_capacity(self):
```
agitter, line 21: We should think about whether we need additional test cases to observe what happens when there is too much flow to push through the network.

test/MinCostFlow/test_mcf.py
```diff
@@ -0,0 +1,62 @@
...
+            singularity=False)
+        assert out_path.exists()
+
+    def test_mincostflow_missing(self):
+        # Test the expected error is raised when required arguments are missing
+        with pytest.raises(ValueError):
+            MinCostFlow.run(sources=TEST_DIR + 'input/sources.txt',
+                targets=TEST_DIR + 'input/targets.txt',
+                output_file=OUT_FILE)
+
```
agitter, line 62: Because min cost flow supports Singularity, we can add another test that runs in Singularity. It can follow this pattern https://github.com/Reed-CompBio/spras/blob/6d42d2f951c3feb1662cf3ca3b72d41364e25c0f/test/PathLinker/test_pathlinker.py#L47-L60

That skips that test if Singularity is not installed locally.

agitter (changes requested):
Now that min cost flow is implemented and working correctly, I'm leaving an initial round of feedback.  This is looking great so far.

We discussed a few more changes that are in progress such as:
-  checking` all_modified_volume_contents` in `util.py` to avoid the `chown` error with empty arguments
- Adding a readme for the docker wrapper
- Assessing the behavior with too much flow

I haven't tried running the code locally yet.  I will also build the Docker image and push it to the reedcompbio Docker organization.

test/MinCostFlow/test_mcf.py
ntalluri: I believe so, I'm not sure why that is in there

test/MinCostFlow/test_mcf.py
ntalluri: Yes I do
 
I also have more test graphs, so I could add all of them with their respective expected outputs

agitter, line 11: If we want to run multiple graphs, we can use syntax like this TPS test case https://github.com/gitter-lab/tps/blob/master/workflow/tests/test_workflow.py#L16

We would add a new argument `graph` and then provide a list of the graphs that are the hard-coded values that argument can take.

src/mincostflow.py
agitter: I pushed the image to the reedcompbio Docker organization: https://hub.docker.com/repository/docker/reedcompbio/mincostflow
```
$ docker build -t reedcompbio/mincostflow -f Dockerfile .
$ docker push reedcompbio/mincostflow
```

We can now change this line to use `reedcompbio/mincostflow` and pull the container from DockerHub if the user does not have it locally.  We can also update the GitHub actions workflow to pull the container for DockerHub as a cache when it builds the container as part of the test workflow.

agitter (approved):
I reviewed all the the latest changes, and the look good.  I pushed a few updates myself updating the Docker readme, reducing which tests cases run on all graphs (to speed up testing), and applying code formatting.

The Snakemake workflow fails, and the error message suggests the problem is `https://github.com/snakemake/snakemake/issues/1899`.  I'm upgrading the version of Snakemake in our conda environment to see if that resolves the problem.  Once the tests pass, I'll merge.

PR #71, opened 2022-09-22
Notes

ntalluri:
DON'T MERGE THIS, delete oct 5th


agitter:
Discussed this with @ntalluri and we'll use https://github.com/ntalluri/spras/pull/1 instead, so we can close this.

PR #72, opened 2022-11-19
Migrate GitHub Actions to Apptainer

agitter:
The version of "Singularity" that joined the Linux Foundation is known as Apptainer. This pull request migrates our GitHub Actions workflow to install Apptainer.

My expectation is that Apptainer currently uses symlinks so that we can continue running `singularity` commands: https://apptainer.org/docs/user/main/singularity_compatibility.html#singularity-command-symlink


agitter:
The four test cases that run pathway reconstruction algorithms in Singularity containers still pass, so I'll merge this.

PR #73, opened 2022-11-28
All pairs shortest paths (AllPairs) Code

annaritz:
I took a look at @nina-young's code from the summer, added a few additional details, and I _believe_ it's ready for review. We used the Contributing Guide as a template. A few notes:

1. The original source code is in `docker-wrappers/AllPairs/all-pairs-shortest-paths.py`. This could be moved outside of spras and into its own little github repo.
2. The docker file is currently on DockerHub under `annaritz/allpairs`. It uses `python:3.9-alpine3.16` and and installs `networkx==2.6.3`.
3. The test functions need to be improved (this is probably the biggest thing).  If the test network produces shortest paths with ties, there is no guarantee that a particular networkx/python distribution will return the same subnetwork. Right now we read in the PathLInker inputs and makes sure that an output file is generated.
4. The code can be streamlined with `pandas`, but this is such a simple algorithm it might not need it.
5. Finally, we're using PathLinker input notation (a single nodes file that denotes "source" or "target"). We could (should?) switch to a `sources.txt` and `targets.txt` files.


annaritz:
Looks like `shutil` isn't defined. I think this is happening with the singularity test on ubuntu.

.github/workflows/test-spras.yml
```diff
@@ -126,3 +126,12 @@ jobs:
         tags: latest
         cache_froms: reedcompbio/mincostflow:latest
         push: false
+    - name: Build AllPairs Docker image
+      uses: docker/build-push-action@v1
+      with:
+        path: docker-wrappers/AllPairs/.
+        dockerfile: docker-wrappers/AllPairs/Dockerfile
+        repository: annaritz/allpairs
```
agitter, line 134: One of us will need to move it to the reedcompbio organization before finalizing the pull request

config/config.yaml
```diff
@@ -28,14 +28,14 @@
  algorithms:
         - name: "pathlinker"
           params:
-                include: true
+                include: false
```
agitter, line 31: Noting to flip these back to true before merging but we can leave them false to keep up testing during review

docker-wrappers/AllPairs/sample-in-net.txt
```diff
@@ -0,0 +1,21 @@
+#Node1	Node2
```
agitter, line 1: Are these the same files in the test subdirectories? Do we need the extra copies here as well or can they be deleted?

src/allpairs.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+    @staticmethod
+    def run(nodetypes=None, network=None, output_file=None, singularity=False):
+        """
+        Run AllPairs with Docker
+        @param nodetypes:  input node types with sources and targets (required)
+        @param network:  input network file (required)
+        @param output_file: path to the output pathway file (required)
+        @param singularity: currently inactive, implement later?
```
agitter, line 53: Does it not work with Singularity or has it not been tested with Singularity? I could test it with Singularity if that's the concern. Because you are running the container with `run_container`, it _should_ work with Docker and Singularity.

src/allpairs.py
```diff
@@ -0,0 +1,108 @@
...
+        volumes.append(bind_path)
+
+        # AllPairs does not provide an argument to set the output directory
+        # Use its --output argument to set the output file to specify an absolute path and filename
+        out_dir = Path(output_file).parent
+        # AllPairs requires that the output directory exist
+        out_dir.mkdir(parents=True, exist_ok=True)
+        bind_path, mapped_out_dir = prepare_volume(str(out_dir), work_dir)
+        volumes.append(bind_path)
+        mapped_out_file = mapped_out_dir + '/out.txt'  # Use posix path inside the container
```
agitter, line 76: We may be able to directly construct the desired final filename here so we don't have to rename the output filename later.

src/local_neighborhood_practice.py
```diff
@@ -0,0 +1,114 @@
+# Test wrapper function as a part of the contributor tutorial #
```
agitter, line 1: Let's delete this file from this pull request.

test/AllPairs/test_ap.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+import docker
```
agitter, line 2:
```suggestion
import shutil
```
It doesn't look like we use docker directly, and shutil will resolve the testing error.

test/AllPairs/test_ap.py
```diff
@@ -0,0 +1,46 @@
...
+        out_path.unlink(missing_ok=True)
+        # Only include required arguments
+        AllPairs.run(
+            nodetypes=TEST_DIR+'input/sample-in-nodetypes.txt',
+            network=TEST_DIR+'input/sample-in-net.txt',
+            output_file=OUT_FILE
+        )
+        assert out_path.exists()
+
+    def test_pathlinker_missing(self):
```
agitter, line 25:
```suggestion
    def test_allpairs_missing(self):
```

agitter (changes requested):
I gave an initial cursory review. The overall structure looks good. I will go through again with a line-by-line review and run the code myself before we merge.

agitter:
Merging #79 created merge conflicts here, which I cleaned up. I also switched to tab-separated outputs.

agitter:
Responding to the questions above:
1. This source code is so short I suggest we make an exception to our usual practice of putting source code in external repositories.
2. I'll push the image to DockerHub.
3. Neha and I will look at the testing. One possibility would be to have a test case for which there is no ambiguity in the shortest paths.
4. I'm fine leaving the current implementation without pandas. It is readable this way.
5. Both versions should work and have the same internal representation after they are ready by `dataset.py`.

docker-wrappers/AllPairs/sample-in-net.txt
ntalluri: I deleted them, and will add a note in the README where to find the format of the files

src/allpairs.py
ntalluri: I am unsure by this comment added in the code, I believe it works based on how it looks

test/AllPairs/test_ap.py
```diff
@@ -0,0 +1,84 @@
...
+from pathlib import Path
+
+import pytest
+
+from src.allpairs import AllPairs
+
+TEST_DIR = 'test/AllPairs/'
+OUT_FILE = TEST_DIR+'output/out.txt'
+
+EXPECTED_FILE = TEST_DIR+'/expected/out.txt' ## TODO not currently checked.
```
agitter, line 11: Is this the expected output for `sample-in-net.txt` and `sample-in-nodetypes.txt`? We should give it a `sample-in` prefix to distinguish it from the `correctness` inputs.

test/AllPairs/test_ap.py
```diff
@@ -0,0 +1,84 @@
...
+
+        with open(out_path, 'r') as f:
+            edge_pairs = f.readlines()
+        output_edges = []
+        for edge in edge_pairs:
+            node1, node2 = sorted(edge.split())
+            output_edges.append((node1, node2))
+        output_edges.sort()
+
+        expected_output = [
```
agitter, line 75: Should we store this in a file even though the edge list is small? I also am going to suggest making a larger correctness test.

test/AllPairs/input/correctness-network.txt
agitter, line 1: We should discuss a correctness test that has more complicated behavior but is still deterministic. This one only has a single linear path, so the algorithm could be broken but still pass the test.

agitter (commented):
Some quick comments on testing.

ntalluri:
the output files for all pairs are not tab separated, but I can't find where in the code the tab separation is missing. I looked at the wrapper code, and everywhere that is writing to a csv file has 'sep='\t''. I also tried adding a 'dellimeter='\t'' to line 68 of all-pairs-shortest-paths.py, and that also didn't work.

agitter:
> the output files for all pairs are not tab separated

It's because we are only calling `AllPairs.run` in the test cases but not `AllPairs.parse_output`. So we're seeing the raw pathway output format from all pairs, which comes from the `nx.write_edgelist(output, output_file, data=False)` line. That uses whitespace (space?) as the default delimiter unless we change it: https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.edgelist.write_edgelist.html

We could change the raw pathway output to also be tab-separated.

test/AllPairs/test_ap.py
```diff
@@ -0,0 +1,85 @@
...
+            nodetypes=TEST_DIR+'input/sample-in-nodetypes.txt',
+            network=TEST_DIR+'input/sample-in-net.txt',
+            output_file=OUT_DIR+'sample-out.txt',
+            singularity=True
+        )
+        assert out_path.exists()
+
+    def test_correctness(self):
+        """
+        Tests algorithm correctness of all_pairs_shortest_path.py by using AllPairs.run
```
agitter, line 54:
```suggestion
        Tests algorithm correctness of all_pairs_shortest_path.py by using AllPairs.run
        The shortest paths are:
        A-B-C
        A-E
        B-C
        B-A-E
        so the union of the unique edges in these paths will be returned as the pathway.
```

agitter:
NetworkX seems to allow a shortest path from a source to itself, so I removed that assert statement: https://networkx.org/documentation/stable/_modules/networkx/algorithms/shortest_paths/weighted.html#single_source_dijkstra_path (search `if target in sources:`). We should test this before merging to confirm.

I build and pushed the Docker image to reedcompbio with 
```
docker build -t reedcompbio/allpairs -f Dockerfile .
docker push reedcompbio/allpairs:latest
```
and then switched the pull request to use that version.

agitter:
@ntalluri I reviewed all of the files and your changes, and this looks just about ready to merge. Can you please double-check the commits I made to see if they look okay? I'll need to make sure the tests still pass.

We could possibly change the Python code to use tab delimiters as a final change.

agitter (approved):

ntalluri:
I checked over the commits and tested the code on my end, and everything looked good/ passed. I do think we need to add the tab delimiters  into the python code.

agitter:
I also updated the node files parsing and the test files. I rebuilt the Docker image and pushed it. If the tests pass, we can merge.

agitter:
It looks like the error comes from the graphspace step in the workflow
```
RuleException:
TypeError in file /home/runner/work/spras/spras/Snakefile, line 235:
Failed to convert rank data D to type <class 'float'>.
  File "/home/runner/work/spras/spras/Snakefile", line 235, in __rule_viz_graphspace
  File "/home/runner/work/spras/spras/src/analysis/graphspace.py", line 14, in write_json
  File "/home/runner/work/spras/spras/src/analysis/graphspace.py", line 41, in get_gs_graph
  File "/home/runner/work/spras/spras/src/analysis/graphspace.py", line 62, in load_graph
  File "/usr/share/miniconda3/envs/spras/lib/python3.8/site-packages/networkx/utils/decorators.py", line 845, in func
  File "<class 'networkx.utils.decorators.argmap'> compilation 5", line 5, in argmap_read_edgelist_1
  File "/usr/share/miniconda3/envs/spras/lib/python3.8/site-packages/networkx/readwrite/edgelist.py", line 376, in read_edgelist
  File "/usr/share/miniconda3/envs/spras/lib/python3.8/site-packages/networkx/readwrite/edgelist.py", line 290, in parse_edgelist
  File "/usr/share/miniconda3/envs/spras/lib/python3.8/concurrent/futures/thread.py", line 57, in run
```

ntalluri:
I saw this error on github "Failed to convert rank data D to type <class 'float'>." for the snakemake workflow, but I can't seem to recreate this error on my end.

PR #80, opened 2023-04-26
Version 2 ideas of what ml can do

ntalluri:


test/ml/test_ml.py
```diff
@@ -18,25 +18,32 @@ def setup_class(cls):
         Path(OUT_DIR).mkdir(parents=True, exist_ok=True)
 
     def test_summarize_networks(self):
-        dataframe = ml.summarize_networks([INPUT_DIR + 's1/s1.txt', INPUT_DIR + 's2/s2.txt', INPUT_DIR + 's3/s3.txt',
-                                           INPUT_DIR + 'longName/longName.txt', INPUT_DIR + 'longName2/longName2.txt',
-                                           INPUT_DIR + 'empty/empty.txt'])
+        dataframe = ml.summarize_networks([INPUT_DIR + 'test-s1/s1.txt', INPUT_DIR + 'test-s2/s2.txt', INPUT_DIR + 'test-s3/s3.txt',
+                                           INPUT_DIR + 'test-longName/longName.txt', INPUT_DIR + 'test-longName2/test-longName2.txt',
```
agitter, line 22: We will investigate why this code ran with the `test-longName2/test-longName2.txt` file that doesn't exist

agitter: We saw that this does fail with missing files here https://github.com/Reed-CompBio/spras/actions/runs/4876715224/jobs/8700546913#step:6:423

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
...
     X = dataframe.reset_index(drop=True)
-    X = X.transpose()
-    model = AgglomerativeClustering(distance_threshold=0.5, n_clusters=None)
+    column_names = X.head()
+    column_names = [element.split('-')[1] for element in column_names]
+    X = X.transpose() 
+
+    # creating the colors per algorithms
+    custom_palette = sns.color_palette("tab10", len(column_names))
+    label_color_map = {label: color for label, color in zip(column_names, custom_palette)}
```
agitter, line 204: Are we able to pass this colormap to the `sns.scatterplot` used for the PCA plot? If we can, then it would make the results consistent to define the colormap ourselves both times to make sure they are consistent. In that case, we could have a new small function in this file that takes a list of column names and returns the map in a deterministic way such as sorting the names.

Snakefile
```diff
@@ -72,9 +72,11 @@ def make_final_input(wildcards):
     if config["analysis"]["ml"]["include"]:  
         final_input.extend(expand('{out_dir}{sep}{dataset}-pca.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
         final_input.extend(expand('{out_dir}{sep}{dataset}-pca-components.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-hac.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-hac-clusters.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
+        final_input.extend(expand('{out_dir}{sep}{dataset}-hac-vertical.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
+        final_input.extend(expand('{out_dir}{sep}{dataset}-hac-clusters-vertical.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
```
agitter, line 76: We discussed how the vertical and horizontal clusters files are redundant because they contain the same clusters. However, we are leaving the behavior as it is because in the next version we might support running vertical or horizontal clustering.

config/config.yaml
```diff
@@ -31,15 +31,17 @@
                 include: true
                 directed: true
                 run1:
-                    k: range(100,201,100)
+                    k: range(100,201,10)
```
agitter, line 34: Revert these changes to the algorithm parameters before merging

config/config.yaml
```diff
@@ -117,4 +120,13 @@
           include: true
         # Machine learning analysis (e.g. clustering) of the pathway output files for each dataset
         ml:
-          include: true
+          include: true 
+          # specify how many components to calculate
```
agitter, line 124:
```suggestion
          # specify how many principal components to calculate
```

config/config.yaml
```diff
@@ -117,4 +120,13 @@
...
+          components: 4
+          # boolean to show the labels on the pca graph
+          labels: true
+          # 'ward', 'complete', 'average', 'single'
+          # if linkage: ward, MUST USE metric: euclidean 
+          linkage: 'ward'
+          # 'euclidean', 'l1', 'l2', 'manhattan', 'cosine', or 'precomputed'
```
agitter, line 131: We don't support precomputed, do we?

src/analysis/ml.py
```diff
@@ -10,14 +9,15 @@
...
 from sklearn.decomposition import PCA
 from sklearn.preprocessing import StandardScaler
-
+from adjustText import adjust_text
 from src.util import make_required_dirs
 
 plt.switch_backend('Agg')
 
+linkage_methods = ["ward", "complete", "average", "single"]
+distance_metrics = ["euclidean", "l1", "l2", "manhattan", "cosine", "precomputed"]
```
agitter, line 18: Probably need to remove `precomputed`. Making these sets is slightly more efficient if we are checking membership and not using the order.

src/analysis/ml.py
```diff
@@ -72,8 +70,7 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
 
     return concated_df
 
-
-def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord: str):
+def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord: str, components=2, labels=True):
```
agitter, line 73: Can specify types for the new arguments and add them to the docstring

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
     @param output_file: the filename to save the variance explained by each component
     @param output_coord: the filename to save the coordinates of each algorithm
     """
+
     df = dataframe.reset_index(drop=True)
+    column_names = df.head()
```
agitter, line 85: Remove this line?

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
+    column_names = [element.split('-')[1] for element in column_names]
     df = df.transpose()  # based on the algorithms rather than the edges
     X = df.values
 
+    min_shape = min(df.shape)
+    if components < 2:
+        raise ValueError(f"components={components} must be greater than or equal to 2 in the config file.")
+    elif components > min_shape:
+        print(f"n_components={components} is not valid. Setting components to {min_shape}.")
```
agitter, line 94:
```suggestion
        print(f"components={components} is not valid. Setting components to {min_shape}.")
```

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
-    variance = pca_2.explained_variance_ratio_ * 100
+    pca = PCA(n_components=components)
+    pca.fit(X_scaled)
+    X_pca = pca.transform(X_scaled)
+    variance = pca.explained_variance_ratio_ * 100
 
     # making the plot
     plt.figure(figsize=(10, 7))
-    sns.scatterplot(x=X_pca_2[:, 0], y=X_pca_2[:, 1], s=70)
+    sns.scatterplot(x=X_pca[:, 0], y=X_pca[:, 1], s=70, hue=column_names, legend=True);
```
agitter, line 111: It would be nice to use the same color map for the algorithms here so that it matches the hierarchical clustering

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
     plt.title("PCA")
     plt.xlabel(f"PC1 ({variance[0]:.1f}% variance)")
     plt.ylabel(f"PC2 ({variance[1]:.1f}% variance)")
 
-    # saving the PCA plot
-    make_required_dirs(output_png)
-    plt.savefig(output_png)
+    # saving the coordinates of each algorithm
```
agitter, line 116: To organize the code, I would move this after the graph labeling so that all of the plotting and plot saving code is together.

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
+    # labeling the graphs
+    if (labels):
+        algorithm_names = df['algorithm'].to_numpy()
+        algorithm_names = [element.split()[0].split('-')[1] for element in algorithm_names]
+        x_coord = df['x'].to_numpy()
+        y_coord = df['y'].to_numpy()
+
+        texts = []
+        for i, algorithm in enumerate(algorithm_names):
+            # TODO: can add a threshold here to allow for labeled points
```
agitter, line 138: What does this TODO refer to? Do we still need it?

src/analysis/ml.py
```diff
@@ -83,44 +80,70 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
+            texts.append(plt.text(x_coord[i], y_coord[i], algorithm, size=9))
+        
+        adjust_text(texts, force_points= [5,5], arrowprops=dict(arrowstyle='->', color='red'))
```
agitter, line 141: Can we use black lines without arrowheads? I prefer a simpler style to make it less intense and less cluttered.

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
     # Plot the corresponding dendrogram
     dendrogram(linkage_matrix, **kwargs)
 
-
-def hac(dataframe: pd.DataFrame, output_png: str, output_file: str):
+def hac_vertical(dataframe: pd.DataFrame, output_png: str, output_file: str, linkage='ward', metric='euclidean'):
     """
     Performs hierarchical agglomerative clustering on the dataframe,
-    creates a dendrogram of the resulting tree,
+    creates a dendrogram of the resulting tree USING SEABORN and SCKIT-LEARN for the cluster groups,
```
agitter, line 181:
```suggestion
    creates a dendrogram of the resulting tree using seaborn and scipy for the cluster groups,
```

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
...
+    model = AgglomerativeClustering(linkage=linkage, affinity=metric,distance_threshold=0.5, n_clusters=None)
     model = model.fit(X)
+    columns = dataframe.columns.tolist()
+    data = {'algorithm': columns, 'labels': model.labels_}
+    df = pd.DataFrame(data)
+    make_required_dirs(output_file)
+    df.to_csv(output_file, sep='\t', index=False)
+
 
+def hac_horizontal(dataframe: pd.DataFrame, output_png: str, output_file: str, linkage='ward', metric='euclidean'):
```
agitter, line 230: Can specify types for the new arguments and add them to the docstring

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
     # Plot the corresponding dendrogram
     dendrogram(linkage_matrix, **kwargs)
 
-
-def hac(dataframe: pd.DataFrame, output_png: str, output_file: str):
+def hac_vertical(dataframe: pd.DataFrame, output_png: str, output_file: str, linkage='ward', metric='euclidean'):
```
agitter, line 178: Can specify types for the new arguments and add them to the docstring

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
...
+    clustergrid.ax_col_dendrogram.set_visible(False)
+    legend_labels = [plt.Rectangle((0, 0), 0, 0, color=label_color_map[label]) for label in label_color_map]
+    plt.legend(legend_labels, label_color_map.keys(), bbox_to_anchor=(1.02, 1), loc='upper left')
+    
+    make_required_dirs(output_png)
+    plt.savefig(output_png, bbox_inches="tight")
+
+    # getting the label of which group each algorithm combination falls under
+    model = AgglomerativeClustering(linkage=linkage, affinity=metric,distance_threshold=0.5, n_clusters=None)
     model = model.fit(X)
```
agitter, line 222: We have to run clustering twice? We can't get this information from the seaborn/scipy clustering? That seems inefficient for large collections of graphs.

src/analysis/ml.py
```diff
@@ -152,27 +175,92 @@ def plot_dendrogram(model, **kwargs):
...
+    data = {'algorithm': columns, 'labels': model.labels_}
+    df = pd.DataFrame(data)
+    make_required_dirs(output_file)
+    df.to_csv(output_file, sep='\t', index=False)
+
 
+def hac_horizontal(dataframe: pd.DataFrame, output_png: str, output_file: str, linkage='ward', metric='euclidean'):
+    """
+    Performs hierarchical agglomerative clustering on the dataframe,
+    creates a dendrogram of the resulting tree USING SCKIT LEARN and makes cluster groups USING SCKIT LEARN,
```
agitter, line 233: Let's avoid all caps. Isn't the dendrogram from scipy and the clustering from scikit-learn?

src/util.py
```diff
@@ -340,7 +340,23 @@ def process_config(config):
...
+    
+    pca_params = {
+        "components": ml_params["components"] if "components" in ml_params else 2,
+        "labels": ml_params["labels"] if "labels" in ml_params else True,
+    }
+
+    hac_params = {
+        "linkage": ml_params["linkage"] if "linkage" in ml_params else 'ward',
+        "metric": ml_params["metric"] if "metric" in ml_params else 'euclidean',
+    }
```
agitter, line 355: Do we need to set default values here or do we let the default values in the functions specify the behavior? I'm unsure. We wouldn't want the hard-coded defaults to diverge though.

agitter (changes requested):

src/analysis/ml.py
agitter: We discussed getting rid of this automated threshold now that the labels can be specified in the config file.

src/analysis/ml.py
agitter: Here is some relevant discussion:
- https://stackoverflow.com/questions/56526707/extract-rows-of-clusters-in-hierarchical-clustering-using-seaborn-clustermap
- https://stackoverflow.com/questions/27924813/extracting-clusters-from-seaborn-clustermap

src/analysis/ml.py
ntalluri: specify them as a new name?

src/analysis/ml.py
ntalluri: Need that to get the column names, used as the labels in the graphs

src/analysis/ml.py
ntalluri: For clustermap, even though it generate the hierarchical clustering and displays the dendrogram, it does not directly return the cluster assignments of the data. 

Rather than using the AgglomerativeClustering to calculate the cluster groups, we can use fcluster from the scipy library. 
1. we would grab the row cluster grid object, which gives a linkage matrix from the sns.clustermap dendrogram
2. we then have to use fcluster
- but it takes in an argument t that cuts the dendrogram at a specific distance to form clusters

` # Use linkage matrix from seaborn clustergrid to generate cluster assignments
    linkage_matrix = clustergrid.dendrogram_row.linkage
    # t = the maximum inter-cluster distance allowed
    clusters = fcluster(linkage_matrix, t=0.5, criterion='distance')`

Or we can get rid of it all together

src/analysis/ml.py
agitter: If found that confusing because `head` returns the first n rows (default 5). Can we use https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.columns.html to get them directly?

src/analysis/ml.py
agitter: Do we know whether the fcluster t argument has the same role as the sklearn AgglomerativeClustering distance_threshold argument? distance_threshold is currently set to 0.5 in our code. If they have same behavior, then using fcluster and setting t=0.5 (or making it another option in the config file) might make sense.

ntalluri: Yeah, the fcluster takes an argument t which is the distance threshold, which is the same (theoretically) as the distance_threshold in AgglomerativeClustering. I made both 0.5, and the way the groupings were identical so that made sense in my mind. The only difference right now is that fcluster is starting the numbering of the groups at 1, and AgglomerativeClustering starts the numberings of groups at 0 (but I am going to see if I can make fcluster start at 0 as well)

ntalluri: the numbering differs between vertical and horizontal now, however the groupings are correct. So, using fcluster does work

config/config.yaml
```diff
@@ -118,3 +118,12 @@
...
           include: true
+          # specify how many principal components to calculate
+          components: 2
+          # boolean to show the labels on the pca graph
+          labels: true
+          # 'ward', 'complete', 'average', 'single'
+          # if linkage: ward, MUST USE metric: euclidean
+          linkage: 'ward'
+          # 'euclidean', 'l1', 'l2', 'manhattan', 'cosine'
+          metric: 'euclidean'
```
agitter, line 129: The two clustering frameworks support different linkage and metric arguments. We decided to support only the options that work with both frameworks.

src/analysis/ml.py
```diff
@@ -82,45 +91,66 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_file: str, output_coord
...
     plt.title("PCA")
     plt.xlabel(f"PC1 ({variance[0]:.1f}% variance)")
     plt.ylabel(f"PC2 ({variance[1]:.1f}% variance)")
 
-    # saving the PCA plot
-    make_required_dirs(output_png)
-    plt.savefig(output_png)
+    # saving the coordinates of each algorithm
+    make_required_dirs(output_coord)
+    data = {'algorithm': columns.tolist(), 'x': X_pca[:, 0], 'y': X_pca[:, 1]}
```
agitter, line 136: If components is > 2, should we write all of the components to this file? Otherwise we are not using them.

agitter (commented):
In my final testing, with data0 and the cosine metric I received this error:
```
RuleException:
ValueError in line 265 of C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\Snakefile:
The condensed distance matrix must contain only finite values.
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\Snakefile", line 265, in __rule_ml_analysis
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\src\analysis\ml.py", line 219, in hac_vertical
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 1258, in clustermap
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 1129, in plot
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 974, in plot_dendrograms
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 687, in dendrogram
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 495, in __init__
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 562, in calculated_linkage
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\seaborn\matrix.py", line 530, in _calculate_linkage_scipy
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\scipy\cluster\hierarchy.py", line 1065, in linkage
  File "C:\Users\agitter\.conda\envs\spras\lib\concurrent\futures\thread.py", line 57, in run
Removing output files of failed job ml_analysis since they might be corrupted:
output/data0-pca.png, output/data0-pca-components.txt, output/data0-pca-coordinates.txt
```

I believe this is caused by all the identical graphs in this dataset. We can leave the behavior for now but may want a more informative error message in the future.

I renamed the PCA output file names. `data1-pca-components.txt` describes the variance explained, not the principal components.

I tested running with multiple cores, and somehow the wrong figure is sometimes saved to `data1-pca.png`. Once it was horizontal hierarchical clustering plot. Once it was the data0 PCA plot. I'm not sure how that happens when the rules run in parallel, but we'll need to debug that. It's fine when I run with 1 core. @ntalluri can you reproduce that behavior? Here's an example where it looks like the same figure object was being manipulated in both parallel rules.
`data1-pca.png`:

src/analysis/ml.py
ntalluri: It would make sense to only give them the top 2, but there is a chance that giving the user all the components could be of value to them. It could help a user understand how much information is kept or lost based on the number of components they choose. Or it could also show them that for their use case, PCA could not be the best choice for using dimensionality reduction. 

We could either make it les confusing and show them the top two (which already happens in the graph), or we can show them the whole thing and hopefully it will aid them when using pca in spras.

agitter: I think it would make the most sense to visualize the top 2 in the plot, which we already do, and then show the requested number of components in the written file here.

ntalluri: I believe that is what it is doing

agitter: The columns of `coord_df` should correspond to the number of components. We currently only create an `algorithm` column and then an `x` and `y` column for the first two components even if `X_pca` has larger dimension. We need to rename the column header so that instead of `x` and `y` is it more general like `comp1`, `comp2`, ... (or whatever is considered standard terminology, let's check the sklearn docs).

To confirm the behavior, I ran the EGFR config file with 4 components. The variance output file has 4 values:
```
60.39857079045826
20.81242547010328
10.34166118187788
8.447342557560589
```
but this coordinates file does not have 4 numeric columns:
```
algorithm	x	y
tps-egfr-pathlinker-params-7S4SLU6	5.169316618551315	-0.46981993977169956
tps-egfr-pathlinker-params-TCEMRS7	8.623647079828245	2.8919659890867155
tps-egfr-omicsintegrator1-params-ITO5EQS	-1.3678061290895147	-4.656992756839353
tps-egfr-omicsintegrator1-params-3THRXWW	-4.330653935914073	-2.8668877196329863
tps-egfr-omicsintegrator1-params-5QH767V	-8.094503633375975	5.1017344271573215
```

agitter (approved):

agitter:
Great work with these final updates. This is ready to merge.

I noticed the algorithm name parsing didn't work in the EGFR dataset because the dataset is called `tps-egfr`. I fixed that in f4f8ad1, but we could have future problems if an algorithm name ever contains `-`. Before the correction, the EGFR plots looked like this.
tps-egfr-hac-vertical

Issue #53, opened 2022-01-24
Use pre-commit hooks to standardize formatting


agitter:
These [pre-commit hooks](https://github.com/pre-commit/pre-commit-hooks) can catch possible errors and formatting mistakes before files are committed.  They can be run locally or during GitHub Actions testing.  We could start small, perhaps by adding a YAML file linter, and then add more over time if it is useful.

agitter:
#83 added pre-commit hooks. However, as noted there I'm leaving this open to discuss what other formatting rules to enforce. For instance, [black](https://github.com/psf/black) is a popular Python code formatter and there are many more [rules](https://beta.ruff.rs/docs/rules/) that ruff supports and [hooks](https://github.com/pre-commit/pre-commit-hooks#hooks-available) available.

PR #83, opened 2023-05-21
Enable pre-commit hooks and fix associated code formatting problems

agitter:
This pull requests implements pre-commit hooks as described in #53. However, I would like to leave #53 open because there are many more formatting rules and external tools we may decide to add over time. For instance, [black](https://github.com/psf/black) is a popular Python code formatter and there are many more [rules](https://beta.ruff.rs/docs/rules/) that ruff supports and [hooks](https://github.com/pre-commit/pre-commit-hooks#hooks-available) available.

The new section of the contributing guide explains how the pre-commit checks work and how developers would use them. It take some adjustment because `git commit` will fail and the commit with _not_ be made if formatting errors are encountered in the files that are in the scope of the commit.

I'll follow up with a commit or two that introduce formatting errors so we can observe the new GitHub Actions test failing.

### DIFF
--- a/.github/workflows/test-spras.yml
+++ b/.github/workflows/test-spras.yml
@@ -126,3 +126,17 @@ jobs:
         tags: latest
         cache_froms: reedcompbio/mincostflow:latest
         push: false
+
+  # Run pre-commit checks on source files
+  pre-commit:
+    name: Run pre-commit checks
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    - name: Setup Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.8' # Match this to the version specified in environment.yml
+    - name: Run pre-commit checks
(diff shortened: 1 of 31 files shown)


agitter:
In 71e06d4 I broke the formatting of the `config.yaml` file to show how the GitHub Actions test that runs the pre-commit hooks detects the invalid yaml file. The new test case [fails](https://github.com/Reed-CompBio/spras/actions/runs/5039448785/jobs/9037597994).

agitter:
Likewise in ae7cf1d I introduced three Python errors or formatting problems in to `dataset.py`. The test case again [fails](https://github.com/Reed-CompBio/spras/actions/runs/5039472037/jobs/9037641180). One of the errors, the unused variable, would have been autocorrected if `pre-commit run --all-files` had been run locally. The docstring formatting and the undefined variable need to be corrected manually.

annaritz (approved):
This pull request looks good - it already cleans up a lot of our code. Thanks @agitter.

PR #85, opened 2023-05-25
Local neighborhood

Lyce24:
The remaining steps for me are step 3 and step 5.
In `src/local_neighborhood.py`:

- I have made modifications to the generation process of the `nodes.txt` and `network.txt` files. However, I have noticed that the inputs generated contain the first row as the title. As an example, one of the generated nodes.txt files follows this format:
NODEID
A
C
I would like to inquire if there is a way to remove the first row and match the format of the input files found in the directory `spras/test/LocalNeighborhood/input`.

- Additionally, I am facing some difficulty in understanding how to implement the `run` function from `local-neighborhood.py`. I am uncertain about the appropriate `command` that should be utilized.

I think I have to implement the `run` function before writing testing function in `test_ln.py`.


annaritz:
We got through running the `run` function in our meeting; however now there's an error in the `summary` rule. 
1. First, set the summary to `false` in the config file and confirm that the `run` function works. Then move on to figuring out why summary doesn't work.
2. Your config file currently runs a single dataset - is this potentially the problem?
3. Is there something we need to add for downstream analysis that is method-dependent? (I don't think so).

agitter:
What was the error in the `summary` rule? Neha was also having errors there recently, so I would like to confirm they are different and not something I broke in a recent pull request.

Her's was related to `ModuleNotFoundError: No module named ‘Dataset’`

Lyce24:
> What was the error in the `summary` rule? Neha was also having errors there recently, so I would like to confirm they are different and not something I broke in a recent pull request.
> 
> Her's was related to `ModuleNotFoundError: No module named ‘Dataset’`

The error occurred during the execution of the Snakemake command. No modifications were made to the code in the `Snakefile` and `src/analysis/summary.py` files. The specific error message encountered is as follows:
image

agitter:
Thanks for the error. That does look different.

What is the rest of the error that is cut off? Is it `line 57, in runner.py`? If so, this could possibly be coming from 
```
from src.local_neighborhood import LocalNeighborhood as local_neighborhood
```

Snakemake sometimes is misleading about the true location of errors. It's possible the underscore in `local_neighborhood` is causing problems, though I don't immediately know _why_ that would cause problems.

Lyce24:
> What is the rest of the error that is cut off? Is it `line 57, in runner.py`? If so, this could possibly be coming from

The screenshot is complete. It just says `File "/home/lyce/miniconda3/envs/spras/lib/python3.8/concurrent/futures/thread.py", line 57, in run`.
image

I guess it's saying the problem is in line 57 of the `run` function in `thread.py` as shown below.
image

agitter:
I think I see what's happening now. It is coming from `summary.run` at line 223 of the Snakefile. The problematic part is `algorithm_directed[wildcards.algorithm]`. The `config.yaml` file hasn't set `directed: false` for the new algorithm so the dictionary doesn't have a value for that key.

This is something we should improve in the config file parsing:
https://github.com/Reed-CompBio/spras/blob/ea5b5fd7154364e0d734caa5241c96c1f1b1a741/src/util.py#L312C1-L313

We should open a new issue to log this problem.

Lyce24:
I believe I have completed the contributing guide by finishing steps 3 and 5. The relevant codes can be found in the latest git commit (if you are unable to locate them, please inform me as I am new to using pull requests). However, while testing the code, I encountered another bug.

Upon running `pytest` for the first time, I encountered the following errors, which appear to be related to graphspace.
image
image
image
image

Upon examining the source code, I found that the issues are present in the following two functions. However, I am uncertain about the cause of these errors since I am not very familiar with graphspace usage.
image

Nevertheless, if I comment out these two functions, the test will pass as shown below.
image

Lyce24:
> I think I see what's happening now. It is coming from `summary.run` at line 223 of the Snakefile. The problematic part is `algorithm_directed[wildcards.algorithm]`. The `config.yaml` file hasn't set `directed: false` for the new algorithm so the dictionary doesn't have a value for that key.

And yes, adding `directed: false` for LocalNeighborhood in `config.yaml` fixes the problem. Thanks for the help!

Lyce24:
> Actually, the `environment.yml` file specifies `networkx=2.8`. @Lyce24 can you verify which version of `networkx` and `graphspace-python` you have? If it's version 2, then can you attach a plain text file of the entire error? That will help me sort through it. Thanks!

Initially, I installed `networkx` (which is version 3) and `pytest` using `pip` in my local development environment. However, when I executed `pytest -s`, the test utilized my local environment instead of the conda environment, resulting in an error. To resolve this, I uninstalled the local versions of `networkx` and `pytest`, and instead used the versions specified in the conda environment. As a result of this adjustment, the test successfully passed as shown below. Thanks!

image

Lyce24:
Reviewed by Anna.

PR #92, opened 2023-06-16
RWR and TieDIE Integration

Lyce24:
## Summary:
This pull request introduces Docker images for RWR (Random Walk with Restart) and TieDIE algorithms, along with the necessary code implementation and tests. Additionally, it includes a new utility function and updates the GitHub Actions workflow.

## Changes Made:
- Created Docker images for RWR and TieDIE in `docker-wrapper/RandomWalk/Dockerfile` and `docker-wrapper/TieDIE/Dockerfile` respectively.
- Implemented the algorithm execution code in `src/random-walk.py` and `src/tiedie.py`.
- Added tests for RWR and TieDIE in the `test/RandomWalk` and `test/TieDIE` directories.
- Updated the GitHub Actions workflow to include the new tests in `.github/workflows/test-spras.yml`.
- Added the `add_rank_column` function to `src/utils.py` for adding an extra column to a dataframe.

## Next Steps:
Transfer the Docker images from the personal DockerHub account to the official DockerHub account for the project once access is granted.

PR #97, opened 2023-07-04
Update contribution guide

### LINKED ISSUE #84: Contribution guide improvements
This issue is to collect challenges and confusion that new team members face as they follow the steps in the contribution guide so that we can improve the document.

One major issue is that "Step 3: Write the Local Neighborhood wrapper functions" is a huge leap from the previous steps. More scaffolding or smaller sub-steps are needed to build toward full implementations of `generate_inputs`, `run_container`, and `parse_output`. For instance, that could involve interactive inspecting the data structures available through the `Dataset` class and running pandas queries with them before implementing `generate_inputs`.

agitter:
Closes #84

This first draft addresses most of the ideas we came up with in #84 but not yet everything. @Lyce24 @ellango2612 @livj4711 you all went through the contribution guide most recently, so you can please give feedback on these updates. Are there parts that require more detail or instructions?

The new Step 0 on forking the repository feels too complicated to me. I like the idea of providing guidance here, but there are many ways to do this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,31 @@ Call the new class within `local_neighborhood.py` `LocalNeighborhood` and set `_
 Specify the list of `required_input` files to be `network` and `nodes`.
 These entries are used to tell Snakemake what input files should be present before running the Local Neighborhood algorithm.
 
-Implement the `generate_inputs` function, following the `omicsintegrator1.py` example.
+Before implementing the `generate_inputs` function, explore the structure of the `Dataset` class interactively.
+In an interactive Python session, run the following commands to load the `data0` dataset and explore the nodes and interactome.
+```python
+> from src.dataset import Dataset
+> dataset_dict = {'label': 'data0', 'node_files': ['node-prizes.txt', 'sources.txt', 'targets.txt'], 'edge_files': ['network.txt'], 'other_files': [], 'data_dir': 'input'}
+> data = Dataset(dataset_dict)
+> data.node_table.head()
+  NODEID  prize sources targets
+0      A    2.0    True     NaN
+1      B    NaN     NaN     NaN
+2      C    5.7     NaN    True
+> data.interactome.head()
+  Interactor1 Interactor2  Weight
+0           A           B    0.98
+1           B           C    0.77
+```
+Also test the functions available in the `Dataset` class.
+```python
+> data.request_node_columns(['sources'])
+  sources NODEID
+0    True      A
+```
+Note the behavior of the `request_node_columns` function when there are missing values in that column of the node table.
+
+Now implement the `generate_inputs` function, following the `omicsintegrator1.py` example.
 The nodes should be any node in the dataset that has a prize set, any node that is a source, or any node that is a target.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -73,8 +111,33 @@ Call the new class within `local_neighborhood.py` `LocalNeighborhood` and set `_
...
+0           A           B    0.98
+1           B           C    0.77
+```
+Also test the functions available in the `Dataset` class.
+```python
+> data.request_node_columns(['sources'])
+  sources NODEID
+0    True      A
+```
+Note the behavior of the `request_node_columns` function when there are missing values in that column of the node table.
```
livj4711, line 136: Also make clear that the NODEID column is always returned by this function

agitter:
The DOMINO pull request #103 adds a new node attribute "active" that will need to be considered when generating inputs for the local neighborhood algorithm.

agitter:
@zhiqianx does the change in 08e2575 clarify how the run function should prepare the output file? If not, can you please make a suggestion.

ctrlaltaf:
The new contributing guide looks better and clarifies many of the steps that was difficult.

I noticed now on step 3's parse output function it says `"See the add_rank_column function in src.util.py"` but I was unable to find this. Is this function in src/util.py or is this something that has not been merged yet?

@agitter This step is related to when someone tries to run an algorithm through spras using `snakemake --cores 1 --configfile config/config.yaml`. I noticed that suppose local neighborhood was able to produce all the inputs and outputs correctly in the newly generated `output` and `.snakemake` directory. If changes are made to the code that break the workflow, for example the `parse_output` function now does not read the csv file correctly, and the user has not deleted the previously generated `output` and `.snakemake` directory, then spras does not throw an error since I believe the correct output files are already there(?). I understand this is a niche case, but would it be wise to include something like this in step 4:

`After running the above spras command, if everything goes well, two new directories 'output' and '.snakemake' will be generated. The output directory verifies if the algorithm is working and is a good thing to investigate if you have errors. It may also be useful that once you have successfully implemented the algorithm and do not need the output files anymore, you delete using rm -rf output .snakemake `

agitter:
Thanks for the feedback @ctrlaltaf

The `add_rank_column` will be added by https://github.com/Reed-CompBio/spras/pull/92. The new instructions anticipate that other pull request being merged soon, but it is confusing in the meantime.

I agree we need to give more guidance about what Snakemake does once the outputs have been written. You shouldn't even have to remove the `.snakemake` directory though. That is used to track history of past runs and logs. Do the updates in 4de05999ff10528981ede43692badff666cd9120 help?

ctrlaltaf:
> Thanks for the feedback @ctrlaltaf
> 
> The `add_rank_column` will be added by #92. The new instructions anticipate that other pull request being merged soon, but it is confusing in the meantime.
> 
> I agree we need to give more guidance about what Snakemake does once the outputs have been written. You shouldn't even have to remove the `.snakemake` directory though. That is used to track history of past runs and logs. Do the updates in [4de0599](https://github.com/Reed-CompBio/spras/commit/4de05999ff10528981ede43692badff666cd9120) help?

Yes I think the new addition is helpful! Thank you for clarifying the .snakemake

PR #101, opened 2023-07-12
Ensemble Networks

ntalluri:


src/analysis/ml.py
```diff
@@ -289,3 +287,22 @@ def hac_horizontal(dataframe: pd.DataFrame, output_png: str, output_file: str, l
...
+    @param output_file: the file name to save the ensemble network
+    """
+    # Compute row sums, then dropped the index
+    row_sums = np.sum(dataframe, axis=1).reset_index()
+    # Rename the columns
+    row_sums.columns = ['Edges', 'Frequency']
+
+    make_required_dirs(output_file)
+    #TODO: user won't be able to open this up in excel  (as well as some of the other files in this code)
+    # should we care about that, or should I change everything to comma seperated and save as csv
```
agitter, line 307: We could try to match the pathway output file format for the individual pathway graphs. For example
```
B	A	1
B	C	1
```

src/analysis/ml.py
```diff
@@ -289,3 +287,22 @@ def hac_horizontal(dataframe: pd.DataFrame, output_png: str, output_file: str, l
...
+    """
+    this method will sum up the binary values in the provided dataframe
+    -  we're counting the number of times an edge appears in a set of pathway reconstruction networks
+    edges that appear more frequently will have a higher number, thus are more likey to be robust,
+    so this information can be used to include an edge to include in a final network
+    @param datafram: binary dataframe of edge comparison between algorithms from summarize_networks
+    @param output_file: the file name to save the ensemble network
+    """
+    # Compute row sums, then dropped the index
+    row_sums = np.sum(dataframe, axis=1).reset_index()
```
agitter, line 301: Can use mean here instead of sum to make everything a relative frequency.

src/analysis/ml.py
```diff
@@ -289,3 +287,21 @@ def hac_horizontal(dataframe: pd.DataFrame, output_png: str, output_file: str, l
...
+    this method will calculate the mean of the binary values in the provided dataframe
+    -  we're counting the number of times an edge appears in a set of pathway reconstruction networks / by the length of the set
+    edges that appear more frequently will have a higher number, thus are more likey to be robust,
+    so this information can be used to include an edge to include in a final network
+    @param datafram: binary dataframe of edge comparison between algorithms from summarize_networks
+    @param output_file: the file name to save the ensemble network
+    """
+    row_means = np.mean(dataframe, axis=1).reset_index()
+    row_means.columns = ['Edges', 'Frequency']
+    row_means[['Node1', 'Node2']] = row_means['Edges'].str.split('\|\|\|',expand=True)
```
agitter, line 302:
```suggestion
    row_means[['Node1', 'Node2']] = row_means['Edges'].str.split(NODE_SEP, expand=True)
```
Will this still work? I'd like to use the constant `NODE_SEP` in case we ever change it.

test/ml/test_ml.py
```diff
@@ -48,3 +48,14 @@ def test_hac_vertical(self):
 
         assert filecmp.cmp(OUT_DIR + 'hac-clusters-vertical.txt', EXPECT_DIR + 'expected-hac-vertical-clusters.txt')
 
+    def test_ensemble_network(self):
+        dataframe = ml.summarize_networks([INPUT_DIR + 'test-data-s1/s1.txt', INPUT_DIR + 'test-data-s2/s2.txt', INPUT_DIR + 'test-data-s3/s3.txt'])
+        ml.ensemble_network(dataframe, OUT_DIR + 'ensemble-network.csv')
```
agitter, line 53: I prefer only using the `.csv` file extension for comma-separated files with actual commas as the separators. Can we change back to `.txt` or `.tsv` for tab-separated, which is also fairly common? We'll need to change the expected file below as well.

src/analysis/ml.py
ntalluri: The problem with using '|||' is that in the str.split is that it is treating the each character in the string as a separate delimiter. Also the '|' means or, so we needed to escape it.

agitter: I tested this with `regex=False`. Initially I had an old version of pandas that didn't support that argument, but when I updated to pandas 1.4.4. (the version in the SPRAS environment) it worked
```
> df = pd.DataFrame({'Edges': ['A|||B', 'C|||D', 'E|||F']})
> df['Edges'].str.split('|||', regex=False, expand=True)
   0  1
0  A  B
1  C  D
2  E  F
```

I still didn't test it here, but it looks closer to what we want.

ntalluri: It worked once I updated my pandas as well

agitter:
I ran into Snakemake bug `https://github.com/snakemake/snakemake/issues/1882` when testing to I had to update the Snakemake version in the conda environment to get the bug fix.

I made some small changes to fix warnings in my editor about how the mean was being taken. I also changed the output filename from "network" to "pathway" to match what the individual output graphs are called.

If the tests pass, I'll merge.

agitter (approved):

PR #103, opened 2023-07-19
Add DOMINO pathway reconstruction algorithm

### LINKED ISSUE #94: Add DOMINO
DOMINO: a network-based active module identification algorithm with reduced rate of false calls
https://doi.org/10.15252/msb.20209593

Available as a Python package
https://github.com/Shamir-Lab/DOMINO

agitter:
Closes #94


docker-wrappers/DOMINO/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# DOMINO wrapper
+# https://github.com/Shamir-Lab/DOMINO
+FROM python:3.7
+
+RUN pip install domino-python
```
agitter, line 5:
```suggestion
RUN pip install domino-python==0.1.1
```
We'll pin a specific version of the package in case they release updates

src/domino.py
```diff
@@ -0,0 +1,174 @@
...
+                            'otjohnson/domino',
+                            command,
+                            volumes,
+                            work_dir)
+        print(out)
+
+
+        ########
+
+        slices_file.unlink(missing_ok=True)
```
agitter, line 129:
```suggestion
        Path(slices_file).unlink(missing_ok=True)
```
My Python editor is complaining that `slices_file` is a string, not a Path, so it doesn't have an `unlink` function. We can convert it to a Path here or earlier to use `unlink` on it.

agitter:
I build the Docker image and pushed it to [Docker Hub](https://hub.docker.com/repository/docker/reedcompbio/domino) so we can change the image name to use `reedcompbio` in the code and start testing the workflow in GitHub Actions. The commands I used
```
docker build -t reedcompbio/domino -f Dockerfile .
docker push reedcompbio/domino
```

docker-wrappers/DOMINO/README.md
```diff
@@ -0,0 +1,16 @@
+# DOMINO Docker image
+
+A Docker image for [DOMINO](https://github.com/Shamir-Lab/DOMINO) that is available on [DockerHub](https://hub.docker.com/repository/docker/otjohnson/domino).
+
```
agitter, line 4: This readme is one place where we can briefly document some implementation decisions, like merging all output modules into a single pathway.

src/domino.py
```diff
@@ -0,0 +1,208 @@
...
+
+        edges_df.to_csv(filename_map['network'],sep='\t',index=False,columns=['Interactor1','ppi','Interactor2'],header=['ID_interactor_A','ppi','ID_interactor_B'])
+
+
+
+    @staticmethod
+    def run(network=None, active_genes=None, output_file=None, use_cache=True, slices_threshold=None, module_threshold=None, singularity=False):
+        """
+        Run DOMINO with Docker
+        Let visualization be always true, parallelization be always 1 thread
```
agitter, line 60: This is another good place to document how the multiple output modules will be merged into a single pathway.

src/domino.py
```diff
@@ -0,0 +1,208 @@
...
+        @param network:  input network file (required)
+        @param active_genes:  input active genes (required)
+        @param output_file: path to the output pathway file (required)
+        @param use_cache: if True, use auto-generated cache network files (*.pkl) from previous executions with the same network (optional)
```
agitter, line 64: Do we know where these are written?

src/domino.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # domino creates a new folder in out_dir to output its modules files into /active_genes
+        out_modules_dir = Path(out_dir, 'active_genes')
+
+        # concatenate each module html file into one big file
+        with open(output_file, "w") as fo:
+            for html_file in out_modules_dir.glob('module_*.html'):
+                with open(html_file,'r') as fi:
+                    fo.write(fi.read())
+                # Path(html_file).unlink(missing_ok=True)
```
agitter, line 155: I'm not sure whether we want to delete these or leave them. They are potentially useful visualizations.

src/domino.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    return ID_PREFIX + node_id
+
+
+def post_domino_id_transform(node_id):
+    """
+    Remove prefix
+    @param node_id: the node id to transformed
+    """
+    node_id = node_id[len(ID_PREFIX):]
```
agitter, line 207: This works and will be fast. I want to think about whether we want to do a string replacement instead to protect against a node ID that doesn't have the expected prefix.

test/DOMINO/test_domino.py
```diff
@@ -0,0 +1,89 @@
...
+        # Test the expected error is raised when network argument is missing
+        with pytest.raises(ValueError):
+            # No network
+            DOMINO.run(
+                active_genes=TEST_DIR+'input/domino-active-genes.txt',
+                output_file=OUT_FILE_DEFAULT)
+
+    def test_domino_parse_output(self):
+        # Show how the complicated file formats work in terms of input and expected output
+        # Concatenated module_0.html and module_1.html file as input
```
agitter, line 60: Please add to the comment the dataset used to obtain these modules. Was it the EGFR dataset?

test/DOMINO/test_domino.py
```diff
@@ -0,0 +1,89 @@
...
+        out_path_exp = Path(OUT_FILE_PARSE_EXP)
+        DOMINO.parse_output(
+            TEST_DIR+'input/domino-concat-modules.txt',
+            OUT_FILE_PARSE)
+        # assert out_path.exists()
+        with open(out_path, 'r') as output_file:
+            generated_output = output_file.read()
+        with open(out_path_exp, 'r') as output_file:
+            expected_output = output_file.read()
+        assert generated_output == expected_output
```
agitter, line 75: Can we use the `util.compare_files` function for this?

agitter (commented):
The wrapper, Docker container, and test cases look great. I left small comments about details as we finalize the testing and documentation.

test/DOMINO/test_domino.py
livj4711: I will change/clarify this comparison to assert the comparison like the ML tests do: 
assert filecmp.cmp(OUT_FILE_PARSE, OUT_FILE_PARSE_EXP)

src/domino.py
livj4711: If our inputted nodes were already Ensembl gene ids, would doing a string replacement of 'ENG0' replace both the prefix we manually added and the beginning 'ENG0' of the gene id?

test/DOMINO/test_domino.py
livj4711: Done, the dataset used is from https://github.com/Shamir-Lab/DOMINO/tree/master/examples.

config/config.yaml
```diff
@@ -65,12 +66,20 @@
...
                 run1:
                     flow: [1] # The flow must be an int
                     capacity: [1]
 
+        - name: "domino"
+          params:
+                include: true
+                directed: false
+                run1:
+                    use_cache: ["true"]
```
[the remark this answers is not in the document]
livj4711: The auto-generated cache network files will be written to the same place the slices.txt file gets written to, right now in spras/. I'm not sure if the run function is able to access this, and am not sure of how to test that either.

src/domino.py
agitter: I decided we should leave them.

src/domino.py
agitter: I wanted to use `removeprefix` but that wasn't added until Python 3.9 https://docs.python.org/3/library/stdtypes.html#str.removeprefix

ntalluri:
before merging, can we add a test for bidirectional edges?

agitter:
@ntalluri I added a copy of the undirected edge A-B as B-A to the network used for DOMINO testing, but it doesn't tell us anything given the way the current tests cases work. DOMINO already raises an error with small files and writes an empty output file, so it will still write an empty output file due to an error even if the new B-A edges raises an error.

I created a new config file locally with the network
```
A	B	0.98
B	C	0.77
C	B	0.77
```
DOMINO ran until it raised the usual error for small inputs
```
raise ValueError(\'cannot apply union_all to an empty list\')\nValueError: cannot apply union_all to an empty list\n'
```

That suggests it was able to read the network. Is that sufficient testing, or should we think about how to design a better formal test case for these pairs of undirected edges?

PR #120, opened 2023-08-23
Adding Directionality

ntalluri:
…orithms in master

### DIFF
--- a/src/domino.py
+++ b/src/domino.py
@@ -11,7 +11,15 @@ __all__ = ['DOMINO', 'pre_domino_id_transform', 'post_domino_id_transform']
 ID_PREFIX = 'ENSG0'
 ID_PREFIX_LEN = len(ID_PREFIX)
 
+"""
+Domino will construct a fully undirected graph from the provided input file
+- in the algorithm, it uses nx.Graph()
 
+Expected raw input format: 
+Interactor1     ppi     Interactor2
+- the expected raw input file should have node pairs in the 1st and 3rd columns, with a 'ppi' in the 2nd column
+- it can include repeated and bidirectional edges
+"""
 class DOMINO(PRM):
     required_inputs = ['network', 'active_genes']
 
--- a/src/meo.py
+++ b/src/meo.py
@@ -46,6 +46,16 @@ def write_properties(filename=Path('properties.txt'), edges=None, sources=None, 
         # Do not need csp.phase, csp.gen.file, or csp.sol.file because MAXCSP is not supported
 
 
+"""
+MEO can support partially directed graphs
+
+Expected raw input format: 
+Interactor1   pp/pd   Interactor2   Weight
+- the expected raw input file should have node pairs in the 1st and 3rd columns, with a directionality in the 2nd column and the weight in the 4th column
+- it use pp for undirected edges and pd for directed edges
+- it cannot include repeated and bidirectional edges
+- there is a chance MRO assumes that it should just be an undirected edge instead
+"""
 class MEO(PRM):
     required_inputs = ['sources', 'targets', 'edges']
 
--- a/src/mincostflow.py
+++ b/src/mincostflow.py
@@ -7,7 +7,16 @@ from src.util import prepare_volume, run_container
 
 __all__ = ['MinCostFlow']
 
+"""
+MinCostFlow deals with fully directed graphs
+- OR Tools MCF is designed for directed graphs
+- when an edge (arc), it has a source and target node, so flow it only allowed to moced from source to the target
 
+Expected raw input format: 
+Interactor1  Interactor2   Weight
+- the expected raw input file should have node pairs in the 1st and 2nd columns, with the weight in the 3rd column
+- it can include repeated and bidirectional edges
+"""
 class MinCostFlow (PRM):
     required_inputs = ['sources', 'targets', 'edges']
 
--- a/src/omicsintegrator1.py
+++ b/src/omicsintegrator1.py
@@ -35,7 +35,17 @@ def write_conf(filename=Path('config.txt'), w=None, b=None, d=None, mu=None, noi
         f.write('processes = 1\n')
         f.write('threads = 1\n')
 
+"""
+Omics Integrator 1 will construct works with partially directed graphs
+- it takes in the universal input directly 
 
+Expected raw input format: 
+Interactor1    Interactor2   Weight    Direction
+- the expected raw input file should have node pairs in the 1st and 2nd columns, with a weight in the 3rd column and directionality in the 4th column
+- it can include repeated and bidirectional edges
+- it uses 'U' for undirected edges and 'D' for directed edges
+
+"""
 class OmicsIntegrator1(PRM):
     required_inputs = ['prizes', 'edges']
 
--- a/src/omicsintegrator2.py
+++ b/src/omicsintegrator2.py
@@ -9,7 +9,16 @@ from src.util import prepare_path_docker
 
 __all__ = ['OmicsIntegrator2']
 
+"""
+Omics Integrator 2 will construct a fully undirected graph from the provided input file
+- in the algorithm, it uses nx.Graph() objects, which are undirected
+- uses a pcst_fast solver which supports undirected graphs
 
+Expected raw input format: 
+Interactor1   Interactor2   Weight
+- the expected raw input file should have node pairs in the 1st and 2nd columns, with a weight in the 3rd column
+- it can include repeated and bidirectional edges
+"""
 class OmicsIntegrator2(PRM):
     required_inputs = ['prizes', 'edges']
 
--- a/src/pathlinker.py
+++ b/src/pathlinker.py
@@ -8,6 +8,15 @@ from src.util import prepare_volume, run_container
 
 __all__ = ['PathLinker']
 
+"""
+Pathlinker will construct a fully directed graph from the provided input file
+- an edge is represented with a head and tail node, which represents the direction of the interation between two nodes
+
+Expected raw input format: 
+Interactor1   Interactor2   Weight
+- the expected raw input file should have node pairs in the 1st and 2nd columns, with a weight in the 3rd column
+- it can include repeated and bidirectional edges
+"""
 class PathLinker(PRM):
     required_inputs = ['nodetypes', 'network']
 


src/util.py
```diff
@@ -313,9 +313,6 @@ def process_config(config):
             # Do not parse the rest of the parameters for this algorithm if it is not included
             continue
 
-        if "directed" in cur_params:
```
agitter, line 316: Would someone ever provide an old format config file that still had the directed keyword? Do we check for this and pop it without storing it?

src/dataset.py
```diff
@@ -131,3 +172,147 @@ def get_other_files(self):
 
     def get_interactome(self):
         return self.interactome.copy()
```
agitter, line 174: We may need this to be a deep copy now that we are manipulating the interactome differently for each algorithm.

src/util.py
ntalluri: added a message indicating the update and popping the param and ignoring it

ntalluri:
Update Summary.py For Mixed Graphs #128

agitter:
Now that we can support mixed graphs, I can update the EGFR example network to the original version that included directed and undirected edges. I'll need to adjust the edge weight that was greater than 1 again. The file came from https://raw.githubusercontent.com/koksal/tps/master/data/networks/phosphosite-irefindex13.0-uniprot.txt

agitter:
I merged the Cytoscape changes and resolved merge conflicts in 39b4748. We can extend Cytoscape to support the U and D edge types in the pull request and then update the image to v2.

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+
+Modify Parse Outputs:
+
+Purpose: Handle and verify the outputs of the new algorithm.
+
+Instructions:
+1. Save the raw-pathway output (from the run function in your wrapper) to /test/ParseOutputs/input. Name it as  {algorithm name}-raw-pathway.txt.
+2. Place the expected parsed output in the test/ParseOutputs/expected directory, following the naming pattern: {algorithm name}-pathway-expected.txt.
+3. Add the new algorithm's name to the algorithms list.
+
```
agitter, line 208: We'll also need to modify the "General steps for contributing a new pathway reconstruction algorithm" below to describe these testing updates.

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
 First pull the image `<username>/local-neighborhood` from Docker Hub.
 Then build the Docker image using the `Dockerfile` that was completed in Step 2.
 
+Modify Generate Inputs:
```
agitter, line 191: Lets use the "``" syntax for filenames in the new instructions below. Please also add the names of the text files to modify.

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+
+Purpose: Update the references and expected files for the new algorithm.
```
agitter, line 193: What is "references" here?

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+
+Instructions:
+1. Append the algorithm name to the existing algorithms list.
```
agitter, line 196: We discussed that this can be removed, we can loop over keys.

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+2. Include a key-value pair in the algo_exp_file dictionary that links the specific algorithm to its expected network file.
+3. Store the expected network file at /test/GenerateInputs/expected using the naming convention: {algorithm name}-{network file name}-expected.txt.
```
agitter, line 198: To make this as user-friendly as possible for new contributors, I recommend adding a sentence "Obtain the expected network file from..."

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+
+Modify Parse Outputs:
+
+Purpose: Handle and verify the outputs of the new algorithm.
```
agitter, line 202:
```suggestion
Purpose: Verify the outputs of the new algorithm are converted to the universal pathway format correctly.
```

CONTRIBUTING.md
```diff
@@ -188,6 +188,24 @@ Follow the example for any of the other pathway reconstruction algorithm.
...
+
+Instructions:
+1. Save the raw-pathway output (from the run function in your wrapper) to /test/ParseOutputs/input. Name it as  {algorithm name}-raw-pathway.txt.
```
agitter, line 205: May want to say this can be obtained by running the Snakemake workflow.

Snakefile
```diff
@@ -60,7 +60,8 @@ def make_final_input(wildcards):
     #TODO analysis could be parsed in the parse_config() function.
     if config["analysis"]["summary"]["include"]:
         # add summary output file for each pathway
-        final_input.extend(expand('{out_dir}{sep}{dataset}-{algorithm_params}{sep}summary.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
+        #TODO: reuse in the future once we make summary work for mixed graphs
```
agitter, line 63:
```suggestion
        #TODO: reuse in the future once we make summary work for mixed graphs. See https://github.com/Reed-CompBio/spras/issues/128
```

Snakefile
```diff
@@ -219,14 +220,15 @@ rule parse_output:
     run:
         runner.parse_output(wildcards.algorithm, input.raw_file, output.standardized_file)
 
+#TODO: reuse in the future once we make summary work for mixed graphs
```
agitter, line 223:
```suggestion
#TODO: reuse in the future once we make summary work for mixed graphs. See https://github.com/Reed-CompBio/spras/issues/128
```

src/analysis/graphspace.py
```diff
@@ -3,15 +3,18 @@
 import sys
 
 import networkx as nx
+import pandas as pd
 from graphspace_python.api.client import GraphSpace
 from graphspace_python.graphs.classes.gsgraph import GSGraph
 
 
-def write_json(graph_file,out_graph,out_style,directed=False) -> None:
+# remove all the the directed = bool in the function names
```
agitter, line 11:
```suggestion
```
I don't think we need to comment on this. Once they are gone, the comment won't have much context.

src/analysis/graphspace.py
```diff
@@ -3,15 +3,18 @@
...
+def write_json(graph_file,out_graph,out_style) -> None:
 
+	print("graph file", graph_file)
```
agitter, line 14: Was this for debugging? Can we remove it?

src/analysis/graphspace.py
```diff
@@ -55,12 +59,27 @@ def get_gs_graph(graph_file:str,graph_name:str,directed=False) -> GSGraph:
 			G.add_edge_style(u,v,width=2,color='#281D6A')
 	return G
 
-## TODO this is a duplicated function in summary.py.
-## Pull this and others into a utils.py function.
-def load_graph(path: str,directed=False) -> nx.Graph:
-	if not directed:
-		G = nx.read_edgelist(path,data=(('rank',float),))
+
+def load_graph(path: str) -> nx.Graph:
```
agitter, line 63: This is more complicated now, please add a comment or docstring describing the behavior. The return type isn't always a Graph, it can be a tuple of a Graph or a DiGraph along with a Boolean.

src/analysis/graphspace.py
```diff
@@ -55,12 +59,27 @@ def get_gs_graph(graph_file:str,graph_name:str,directed=False) -> GSGraph:
...
+		G = nx.read_edgelist(path,data=(('rank',float), ('Direction',str)))
+		directed = False
+	elif mask_d.all():
+		G = nx.read_edgelist(path,data=(('rank',float),('Direction',str)), create_using=nx.DiGraph)
+		directed = True
 	else:
-		# note - self-edges are not allowed in DiGraphs.
-		G = nx.read_edgelist(path,data=(('rank',float),),create_using=nx.DiGraph)
-	return G
+		print("graphspace does not deal with mixed direction type graphs currently")
```
agitter, line 83:
```suggestion
		print("GraphSpace does not deal with mixed direction type graphs currently")
```
Do we need to print the filename or some identifier for the graph as part of this message saying we could not not visualize graph X?

src/analysis/graphspace.py
```diff
@@ -55,12 +59,27 @@ def get_gs_graph(graph_file:str,graph_name:str,directed=False) -> GSGraph:
...
+		pathways = pd.read_csv(path, sep="\t", header=None)
+	except pd.errors.EmptyDataError:
+		print(f"The file {path} is empty.")
+		return G, directed
+	pathways.columns = ["Interactor1", "Interactor2", "Rank", "Direction"]
+	mask_u = pathways['Direction'] == 'U'
+	mask_d = pathways['Direction'] == 'D'
+
+	if mask_u.all():
+		G = nx.read_edgelist(path,data=(('rank',float), ('Direction',str)))
```
agitter, line 77: What is the `('Direction',str)` part storing in the graph? Should we drop this column from the dataframe before creating the networkx graph?

src/analysis/ml.py
```diff
@@ -45,7 +47,13 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
                 if len(parts) > 0:  # in case of empty line in file
                     node1 = parts[0]
                     node2 = parts[1]
-                    edges.append(NODE_SEP.join(sorted([node1, node2])))  # assumes edges are undirected
+                    direction = str(parts[3]).strip()
+                    if direction == "U":
+                        edges.append(UNDIR_CONST.join([node1, node2]))
```
agitter, line 52: Do we still need to sort nodes for undirected edges? Otherwise we may not match edges A-B and B-A in different graphs.

src/analysis/ml.py
```diff
@@ -300,8 +308,5 @@ def ensemble_network(dataframe: pd.DataFrame, output_file: str):
     """
     row_means = dataframe.mean(axis=1, numeric_only=True).reset_index()
     row_means.columns = ['Edges', 'Frequency']
-    row_means[['Node1', 'Node2']] = row_means['Edges'].str.split(NODE_SEP, expand=True, regex=False)
```
agitter, line 303: Did the output format of the ensemble change here? We are no longer splitting nodes.

agitter (commented):
I started a code review but only made it through `ml.py`.

src/analysis/ml.py
ntalluri: before we were splitting the nodes, but we had to factor in the directionality into it as we now differ between the two in ml. 

This is the new format: 
Edges	Frequency
A-->B	0.16666666666666666
B-->C	0.16666666666666666
A---B	0.3333333333333333
B---C	0.4166666666666667
B---A	0.08333333333333333
B-->A	0.08333333333333333
C-->B	0.08333333333333333

ntalluri: the idea was this based on the comment for the ensemble function: Counts the number of times an edge appears in a set of pathways and divides by the total number of pathways. Edges that appear more frequently across pathways are more likely to be robust,

spras/dataset.py
```diff
@@ -63,25 +63,47 @@ def load_files_from_dict(self, dataset_dict):
...
 
-        #Load everything as pandas tables
-        self.interactome = pd.read_table(os.path.join(data_loc,interactome_loc), names = ["Interactor1","Interactor2","Weight"])
+        # Load everything as pandas tables
+        self.interactome = pd.read_table(
+            os.path.join(data_loc, interactome_loc), sep="\t", header=None
+        )
+        num_cols = self.interactome.shape[1]
+        if num_cols == 3:
+
```
agitter, line 79:
```suggestion
```

spras/dataset.py
```diff
@@ -63,25 +63,47 @@ def load_files_from_dict(self, dataset_dict):
...
         data_loc = dataset_dict["data_dir"]
 
-        #Load everything as pandas tables
-        self.interactome = pd.read_table(os.path.join(data_loc,interactome_loc), names = ["Interactor1","Interactor2","Weight"])
+        # Load everything as pandas tables
+        self.interactome = pd.read_table(
+            os.path.join(data_loc, interactome_loc), sep="\t", header=None
+        )
+        num_cols = self.interactome.shape[1]
+        if num_cols == 3:
```
agitter, line 78: Do we document anywhere that if the interaction file has three columns we assume it is undirected? We need to update `input/README.md` to describe these changes to the input edge file format and show an example of four column format, keeping the three column example.

spras/dataset.py
```diff
@@ -63,25 +63,47 @@ def load_files_from_dict(self, dataset_dict):
...
+
+            self.interactome.columns = ["Interactor1", "Interactor2", "Weight"]
+            self.interactome["Direction"] = "U"
```
agitter, line 81:
```suggestion
            # When no direction is specified, default to undirected edges
            self.interactome["Direction"] = "U"
```

spras/dataset.py
```diff
@@ -63,25 +63,47 @@ def load_files_from_dict(self, dataset_dict):
...
+        elif num_cols == 4:
+            self.interactome.columns = [
+                "Interactor1",
+                "Interactor2",
+                "Weight",
+                "Direction",
+            ]
+        else:
+            raise ValueError(
+                f"edge_files must have three or four columns but found {num_cols}"
```
agitter, line 92:
```suggestion
                f"Edge file {interactome_loc} must have three or four columns but found {num_cols}"
```

spras/dataset.py
```diff
@@ -63,25 +63,47 @@ def load_files_from_dict(self, dataset_dict):
...
+            )
+
         node_set = set(self.interactome.Interactor1.unique())
         node_set = node_set.union(set(self.interactome.Interactor2.unique()))
 
-        #Load generic node tables
+        # Load generic node tables
```
agitter, line 98: Can you confirm all the node table changes are only code formatting?

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
+"""
+Author: Neha Talluri
+07/19/23
+
+Methods for converting from the universal input and recreating the universal output
```
agitter, line 5:
```suggestion
Methods for converting from the universal network input format and to the
universal network output format
```

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+
+import pandas as pd
+
+
+def convert_undirected_to_directed(df: pd.DataFrame) -> pd.DataFrame:
+    """
+    turns a graph into a fully directed graph
+    - turns every unidirected edges into a bi-directed edge
+    - with bi-directed edges, we are not loosing too much information because the relationship of the undirected egde is still preserved
```
agitter, line 15:
```suggestion
    - turns every undirected edge into a pair of directed edges
    - with the pair of directed edges edges, we are not loosing too much information because the relationship of the undirected edge is still preserved
```

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+
+   * A user must keep the Direction column when using this function
```
agitter, line 17: What does this mean?

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    df.loc[mask, 'Direction'] = 'D'
+    df = pd.concat([df, new_df], ignore_index=True)
+    return df
+
+
+def convert_directed_to_undirected(df: pd.DataFrame) -> pd.DateOffset:
+    """
+    turns a graph into a fully undirected graph
+    - turns all the directed edges directly into undirected edges
+    - we will loose any sense of directionality and the graph won't be inherently accurate, but the basic relationship between the two connected nodes will still remain intact.
```
agitter, line 36:
```suggestion
    - we will lose any sense of directionality and the graph won't be inherently accurate, but the basic relationship between the two connected nodes will still remain intact.
```

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    @param df: input network df of edges, weights, and directionality
+    @return a dataframe with no directed edges in Direction column
+    """
+
+    df["Direction"] = "U"
+
+    return df
+
+
+def add_constant(df: pd.DataFrame, new_col_name: str, const: str) -> pd.DataFrame:
```
agitter, line 49: Does the `const` have to be a string?

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    adds a constant at the end of the input dataframe that is needed inbetween columns for a specifc algorithm
```
agitter, line 51:
```suggestion
    adds a new column at the end of the input dataframe with a constant value in all rows
```

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    @param const: some type of constant needed in the df
+    @return a df with a new constant added to every row
+    """
+
+    df.insert(df.shape[1], new_col_name, const)
+
+    return df
+
+
+def add_directionality_constant(df: pd.DataFrame,  col_name: str, dir_const: str, undir_const: str) -> pd.DataFrame:
```
agitter, line 64: Would `add_directionality_sep` be a better name for this? The value isn't a constant in this case because it is different for directed and undirected edges. I would then change the arguments to `dir_sep` and `undir_sep` as well. If we change them here, we should also change them in `readd_direction_col_mixed`.

It would likely help document these functions if this function and all of the above functions include something in the docstring about being used to generate inputs whereas the following functions are used for parse output.

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    @param undir_const: the undirected edge sep
+    @return a df converted to show directionality differently
+    """
+    df.insert(df.shape[1], col_name, dir_const)
+    mask = df['Direction'] == 'U'
+    df.loc[mask, col_name] = undir_const
+
+    return df
+
+def readd_direction_col_mixed(df: pd.DataFrame, existing_direction_column: str, dir_const: str, undir_const: str) -> pd.DataFrame:
```
agitter, line 82: These next three function names are long, and I find "readd" to be similar to "read" with a typo, which confused by the first time I saw it. Is there a simpler naming scheme we can use? Maybe `add_direction_mixed`, etc.?

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    readds a 'Direction' column that puts a 'U' or 'D' at the end of provided dataframe
+    based on the dir/undir constants in the existing direction column
+
+    *user must keep the existing direction column when using the function
```
agitter, line 87: I'm still unsure what these comments are suggesting. Does this mean don't delete the Direction column after calling this function?

spras/interactome.py
```diff
@@ -0,0 +1,123 @@
...
+    @param existing_direction_column: the name of the existing directionality column
+    @param dir_const: the directed edge sep
+    @param undir_const: the undirected edge sep
+    @return a df with universal Direction column added back
+    """
+
+    df.insert(df.shape[1], "Direction", "D")
+
+    mask_undir = df[existing_direction_column] == undir_const
+    df.loc[mask_undir, "Direction"] = "U"
```
agitter, line 99: This doesn't use the `dir_const`. Do we need to check whether the edges that aren't undirected are directed and throw an error otherwise?

spras/meo.py
```diff
@@ -46,6 +47,16 @@ def write_properties(filename=Path('properties.txt'), edges=None, sources=None,
...
 
+"""
+MEO can support partially directed graphs
+
+Expected raw input format:
+Interactor1   pp/pd   Interactor2   Weight
+- the expected raw input file should have node pairs in the 1st and 3rd columns, with a directionality in the 2nd column and the weight in the 4th column
+- it use pp for undirected edges and pd for directed edges
+- it cannot include repeated and bidirectional edges
+- there is a chance MRO assumes that it should just be an undirected edge instead
```
agitter, line 58:
```suggestion
- it cannot include repeated and directed edges in opposite directions(A->B and B->A in the input)
- MEO assumes that it should be an undirected edge instead
```
Is this the case you are referring to?

spras/meo.py
```diff
@@ -167,4 +179,7 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
         # TODO what should be the edge rank?
         # Would need to load the paths output file to rank edges correctly
         df.insert(5, 'Rank', 1)  # Add a constant rank of 1
-        df.to_csv(standardized_pathway_file, columns=['Source', 'Target', 'Rank'], header=False, index=False, sep='\t')
+
+        df = readd_direction_col_mixed(df, "Type", "pd", "pp")
```
agitter, line 183: Does it output mixed graphs or only directed graphs? The algorithm is designed to assign directions to undirected edges.

spras/mincostflow.py
```diff
@@ -2,12 +2,25 @@
...
+)
 from spras.prm import PRM
 from spras.util import prepare_volume, run_container
 
 __all__ = ['MinCostFlow']
 
+"""
+MinCostFlow deals with fully directed graphs
+- OR Tools MCF is designed for directed graphs
+- when an edge (arc), it has a source and target node, so flow it only allowed to moced from source to the target
```
agitter, line 17:
```suggestion
- when an edge (arc) is added, it has a source and target node, so flow is only allowed to move from source to the target
```

spras/mincostflow.py
```diff
@@ -127,4 +143,5 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
 
         df = pd.read_csv(raw_pathway_file, sep='\t', header=None)
         df.insert(2, 'Rank', 1)  # adds in a rank column of 1s because the edges are not ranked
+        df = readd_direction_col_directed(df)
```
agitter, line 146: The original paper using min cost flow on biological networks said the direction of flow was not meaningful. We should consider what directionality is appropriate for the output graph.

spras/omicsintegrator1.py
```diff
@@ -35,7 +36,17 @@ def write_conf(filename=Path('config.txt'), w=None, b=None, d=None, mu=None, noi
         f.write('processes = 1\n')
         f.write('threads = 1\n')
 
+"""
+Omics Integrator 1 will construct works with partially directed graphs
```
agitter, line 40:
```suggestion
Omics Integrator 1 works with partially directed graphs
```

spras/meo.py
```diff
@@ -167,4 +179,7 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
...
+
+        df.to_csv(standardized_pathway_file, columns=['Source', 'Target', 'Rank', "Direction"], header=False, index=False, sep='\t')
```
agitter, line 185: Eventually we want to start writing the header row as part of the standard pathway output format. Is that part of this pull request or a separate change?

spras/omicsintegrator2.py
```diff
@@ -34,11 +47,17 @@ def generate_inputs(data: Dataset, filename_map):
         else:
             raise ValueError("Omics Integrator 2 requires node prizes or sources and targets")
 
-        # Omics Integrator already gives warnings for strange prize values, so we won't here
-        node_df.to_csv(filename_map['prizes'], sep='\t', index=False, columns=['NODEID', 'prize'], header=['name', 'prize'])
+        #Omics Integrator already gives warnings for strange prize values, so we won't here
```
agitter, line 50:
```suggestion
        # Omics Integrator already gives warnings for strange prize values, so we won't here
```

spras/omicsintegrator2.py
```diff
@@ -34,11 +47,17 @@ def generate_inputs(data: Dataset, filename_map):
...
+
+        # Create network file
         edges_df = data.get_interactome()
 
-        # We'll have to update this when we make iteractomes more proper, but for now
+        # Format network file
+        # edges_df = convert_directed_to_undirected(edges_df)
+        # - technically this can be called but since we don't use the column and based on what the function does, it is not truly needed
+
+        #We'll have to update this when we make iteractomes more proper, but for now
```
agitter, line 60:
```suggestion
        # We'll have to update this when we make iteractomes more proper, but for now
```

agitter (changes requested):
I continued reviewing and made it through all of the source files under the `spras` directory.

spras/mincostflow.py
agitter: The information is in the supplement https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2733244/bin/NIHMS109939-supplement-supp.pdf and the section is "Directionality of ResponseNet output". We could make a reference to that in the comments to explain why we treat the output as undirected. It states
> The flow algorithm we employ provides a directed network. However the directionality
of the interactions in the network is determined by the fact that we have connected all
genetic hits to the source of flow and all differentially expressed genes to the flow sink.
Therefore, except for the interactions between transcription factors and their targets, the
flow does not necessarily reflect a causal order of events

spras/meo.py
agitter: We discussed this and the meaning of this part of the MEO readme
> The edges are annotated as pp for edges that were undirected in the original network and pd for edges that were originally directed.

MEO tracks the directionality of the original edges, but all of its output edges are directed. That makes its pathways quite different from all of the other SPRAS algorithms. However, to remain accurate to MEO's design we will also treat the output graph's as directed.

test/ParseOutputs/expected/meo-pathway-expected.txt
```diff
@@ -0,0 +1,2 @@
+GENEA	GENEC	1	U
```
agitter, line 1: These should end up as directed edges

test/ParseOutputs/expected/mincostflow-pathway-expected.txt
```diff
@@ -0,0 +1,2 @@
+B	A	1	D
```
agitter, line 1: These will end up as undirected edges

test/ParseOutputs/input/omicsintegrator1-raw-pathway.txt
```diff
@@ -0,0 +1,2 @@
+A	pp	C
```
agitter, line 1: We could add a directed edge to make sure that conversion works

test/GenerateInputs/test_generate_inputs.py
```diff
@@ -0,0 +1,36 @@
...
+
+        data0_dataset = next((ds for ds in config["datasets"] if ds["label"] == "data0"), None)
+
+        for algo in algo_exp_file.keys():
+            inputs = runner.get_required_inputs(algo)
+            runner.merge_input(data0_dataset, test_file)
+            filename_map = {input_str: os.path.join("test", "GenerateInputs", "output", f"{algo}-{input_str}.txt") for input_str in inputs}
+            runner.prepare_inputs(algo, test_file, filename_map)
+            exp_file_name = algo_exp_file[algo]
+            assert filecmp.cmp(OUTDIR +f"{algo}-{exp_file_name}.txt", EXPDIR + f"{algo}-{exp_file_name}-expected.txt")
```
agitter, line 36:
```suggestion
            assert filecmp.cmp(OUTDIR +f"{algo}-{exp_file_name}.txt", EXPDIR + f"{algo}-{exp_file_name}-expected.txt", shallow=False)
```

test/ParseOutputs/test_parse_outputs.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        Path(OUTDIR).mkdir(parents=True, exist_ok=True)
+
+    def test_parse_outputs(self):
+        for algo in algorithms:
+            test_file = INDIR + f"{algo}-raw-pathway.txt"
+            out_file = OUTDIR + f"{algo}-pathway.txt"
+            print(out_file)
+            runner.parse_output(algo, test_file, out_file)
+            assert filecmp.cmp(OUTDIR +f"{algo}-pathway.txt", EXPDIR + f"{algo}-pathway-expected.txt")
```
agitter, line 34:
```suggestion
            assert filecmp.cmp(OUTDIR +f"{algo}-pathway.txt", EXPDIR + f"{algo}-pathway-expected.txt", shallow=False)
```

test/ParseOutputs/test_parse_outputs.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_domino_parse_output(self):
```
agitter, line 36: Does DOMINO still need a separate test case? It looks like it follows the same pattern as the others except it unlinks the output file before running.

If we merge it with the others, I would like to keep the comments somewhere that describe where the DOMINO input came from.

test/analysis/input/example/data0-meo-params-GKEDDFZ_pathway.txt
```diff
@@ -1,2 +1,2 @@
-A	B	1
-B	C	1
+A	B	1	U
```
agitter, line 1: Not sure whether this matters, but does this MEO example need to be directed?

test/analysis/input/example/data1-meo-params-GKEDDFZ_pathway.txt
```diff
@@ -1,4 +1,4 @@
-A	B	1
-B	C	1
-A	D	1
-C	D	1
+A	B	1	U
```
agitter, line 1: Likewise, direct the MEO edges?

agitter (changes requested):
I finished reviewing the rest of the files in the pull request.

spras/dataset.py
ntalluri: We're checking to see if there are three or four columns, but I'm not sure where we are throwing an error if a fourth column in the network file has missing data or random values.
Right now, it appears like different areas of the code or algorithm code are raising the issues, but there isn't a specific location where the error is being produced. 

Should I include anything that checks for this and throws an error, or should I leave it up to the user to ensure their file is correct once I update the documentation?

agitter: That's a good point. We could check the fourth column to ensure it does not have missing values and that it only has "U" or "D" as values.

agitter (approved):

PR #122, opened 2023-08-24
Initial test: Adding LocalNeighborhood to SPRAS

oliverfanderson:
First attempt at adding LocalNeighborhood method to SPRAS using the CONTRIBUTING.md guide. Output appears to be in the desired format. Ran into some hurdles with the summary and graphspace analyses in config.yaml: needed to set these as false to get the local neighborhood operational.


oliverfanderson:
Excellent, thank you. Altaf helped me debug the error causing summary and graphspace to not work with Local Neighborhood, so I pushed another commit.

agitter:
We're done with this contributing guide pull request

PR #123, opened 2023-08-25
Update Omics Integrator 2 container and usage

### LINKED ISSUE #54: Add Singularity support for other pathway reconstruction algorithms
#49 adds general support for Singularity and converts Omics Integrator 1 to the new abstract container interface that can run commands in Docker or Singularity.  This new interface should be used for the other pathway reconstruction algorithms.  This will require either converting their images to not use conda or resolving issues with automatically running commands in conda environments in Singularity.

agitter:
Closes #54 

Omics Integrator 2 is the last pathway reconstruction algorithm that does not support Singularity. This pull request will update it to use the new generic container running interface and update its Docker image to a format that can be converted to work with Singularity. The current image does not because it relies on an ENTRYPOINT that runs commands in a named conda environment.

### DIFF
--- a/.github/workflows/test-spras.yml
+++ b/.github/workflows/test-spras.yml
@@ -77,7 +77,7 @@ jobs:
     - name: Pull Docker images
       run: |
         docker pull reedcompbio/omics-integrator-1:latest
-        docker pull reedcompbio/omics-integrator-2:latest
+        docker pull reedcompbio/omics-integrator-2:v2
         docker pull reedcompbio/pathlinker:latest
         docker pull reedcompbio/meo:latest
         docker pull reedcompbio/mincostflow:latest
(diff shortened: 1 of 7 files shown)


agitter:
@jhiemstrawisc this pull request is going to update the Dockerfile for one of the pathway reconstruction algorithms. I'd like to manually tag the original and new version even before we figure out the new, general image versioning scheme.

What do you suggestion for the initial image tagging convention? `rev1`, `rev2`, etc.? Something else?

ntalluri:
I checked over the code, and from what I could see, everything looks good

jhiemstrawisc:
Sorry for the delayed response, this appears to have gotten buried in email for me. For now, I think keeping it simple is okay -- something like `rev1/2` like you suggest, or even just `v1/2`.

Ultimately, I think the two pieces of information we want to get from a particular container version are "what is the minimal version of SPRAS that this container is compatible with" and "what is the actual revision of this packaged container". The former tells users what they need to know for actually using the container, and the latter helps the developer keep track of any bugs that might appear between revisions. How does something like v1.2.3-4 sound, where v1.2.3 indicates the minimal spras version (according to semver) that can run this container, and -4 represents the particular version of the container? Any time the first half of the version bumps to something new, the container revision number can reset. 

I don't think the second portion of the tag needs to be semantically versioned, because in theory any time a breaking change is made to the container itself, spras will also have to be updated, resulting in a new vX.X.X number.

agitter:
I built the original version of the image, tagged it as `v1`, and pushed that to DockerHub. Then I did the same with `v2` using the updated image and changed the wrapper use to use `v2` instead of the implicit `latest`.

If the tests pass, I'll merge.

PR #126, opened 2023-09-12
Make SPRAS installable

### LINKED ISSUE #86: Create installable Python package
If we packaged SPRAS as a proper Python package it could be distributed more easily through PyPI or conda. I'm not sure how that would work with our config files or data files. An initial first step could be to distribute the source code as a package.

I came across this great [blog post](https://drivendata.co/blog/python-packaging-2023) describing modern Python packaging that is now based on `pyproject.toml` instead of `setup.py`. #83 adds an initial `pyproject.toml` to our project but does not specify all of this package metadata.

jhiemstrawisc:
This commit adds the ability to install spras as a python module using pip, which is something we'll need for running in remote environments such as on the OSPool. The installation "instructions" used by pip during building are laid out in `pyproject.toml`. While I chose to use setup-tools as a builder backend in this first pass, other builder backends with more robust build features are available and may eventually be switched to (eg flit, poetry).

Several additional files had to be modified to adhere to python packaging guidelines, the most notable of which was changing `src` to spras, and referencing the new spras directory instead of src. Several lines of omicsintegrator2.py were also modified becasue they were dealing with an absolute path, which made the workflow un-runnable outside of the repo. I verified that all tests and the entirety of the snakefile still run both in the conda environment from the repo root (which is how the tool is currently used) and in a separate directory with spras installed where the input data, snakefile, and config file have been brought along.

For testing, install spras using `pip install .` in the repo root. If the extra development packages are needed, spras can be installed with the command `pip install .[dev]`. Spras can also be installed via `pip install -e .` to make the installation editable. That is, when installed via this command, you can continue to edit spras in the repo and have those changes reflected anywhere you run spras on the machine.

Note that several package versions had to be modified (they're noted in pyproject.toml) to get past a few versioning issues between conda and pypi. To the extent of my knowledge and testing, these changes didn't affect the way spras runs. Famous last words?

If this passes tests and is merged, it closes #86.


jhiemstrawisc:
P.S. I only added Tony in the `pyproject.toml` authors because I wasn't sure who else to reference here. Feel free to adjust as necessary!

pyproject.toml
```diff
@@ -1,3 +1,56 @@
+[project]
+name = "SPRAS"
+version = "v0.0.1"
+description = "Signaling Pathway Reconstruction Analysis Streamliner (SPRAS)"
+authors = [
+  { name = "Anthony Gitter", email = "agitter@morgridge.org" },
```
agitter, line 6:
```suggestion
  { name = "Anthony Gitter", email = "gitter@biostat.wisc.edu" },
  { name = "Anna Ritz", email = "aritz@reed.edu" },
```

pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+]
+license = { file = "LICENSE" }
+readme = "README.md"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
```
agitter, line 13:
```suggestion
    "Operating System :: OS Independent",
    "Development Status :: 3 - Alpha",
    "Intended Audience :: Science/Research",
    "Topic :: Scientific/Engineering :: Bio-Informatics",
```
Pulling from https://pypi.org/classifiers/. Should we sort these?

pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+dependencies = [
+    "adjusttext==0.7.3",
+    "snakemake==7.19.1",
+    "docker==5.0.3", # Switched from docker-py to docker because docker-py is not maintained in pypi. This appears to have no effect
+    "matplotlib==3.5",
+    "networkx==2.8",
+    "pandas==1.4",
+    "pip==22.1",
+    "requests==2.28",
+    "scikit-learn==1.2", #Version bumped from 1.1-->1.2 to get past compilation errors. No other errors encountered
```
agitter, line 25: Let's update the version in `environment.yml` as well so they match

pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+    "seaborn==0.12",
+    "spython==0.2",
+    # Only required for GraphSpace
+    "commonmark==0.9",
+    "docutils", #==0.18
```
agitter, line 30: Was there a reason for not pinning this version?

pyproject.toml
```diff
@@ -1,3 +1,56 @@
...
+[project.optional-dependencies]
+dev = [
+   # Only required for development
+    "pre-commit==2.20",
+    "pytest==7.1",
+]
+
+[project.urls]
+"Homepage" = "https://github.com/Reed-CompBio/spras"
+"Bug Tracker" = "https://github.com/Reed-CompBio/spras/issues"
```
agitter, line 48:
```suggestion
"Issues" = "https://github.com/Reed-CompBio/spras/issues"
```

I prefer a more generic term because we use issues for bugs and general discussion about the project

agitter (changes requested):
I reviewed https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata and made suggestions for the metadata. We should let @annaritz comment as well.

Do we want to comment on installing with `pip` either in the main readme or in the contributing guide for developers?

Everything else looks good to me.

annaritz:
`pyproject.toml` authors and emails look good to me.

README.md
```diff
@@ -35,6 +35,13 @@ conda activate spras
 to create a conda environment with the required packages and activate that environment.
 If you have a different version of Python already, you can install the specified versions of the required packages in your preferred manner instead of using Anaconda.
 
+SRAS can also be installed with Python's package manager `pip` by running `python -m pip install .` from the repo's root directory. This makes SPRAS more portable by allowing you to run
+snakemake anywhere the Snakefile, data, and configuration file can be accessed. However, use caution when pip installing directly to your computer without using some form of virtual
+environment as this can alter your system's underlying python modules, which could lead to unexpected behavior.
+
+For developers, SPRAS can be installed via `pip` with the `-e` flag, as in `python -m pip install -e .`. This points Python back to the SPRAS repo so that any changes made to the source
+code are reflected in the installed module.
+
```
agitter, line 44: Later in this file there is a reference to the `src/` subdirectory we should update.

agitter (changes requested):
I tested the installation locally
```
$ conda create -n spras-new python=3.8
$ conda activate spras-new
$ python -m pip install -e .[dev] .
```
This failed with
```
...
Collecting semver>=2.8.1 (from spython==0.2->spras==0.0.1)
  Obtaining dependency information for semver>=2.8.1 from https://files.pythonhosted.org/packages/9a/77/0cc7a8a3bc7e53d07e8f47f147b92b0960e902b8254859f4aee5c4d7866b/semver-3.0.2-py3-none-any.whl.metadata
  Downloading semver-3.0.2-py3-none-any.whl.metadata (5.0 kB)
INFO: pip is looking at multiple versions of spras[dev] to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install spras 0.0.1 (from \spras) and spras[dev]==0.0.1 because these package versions have conflicting dependencies.
The conflict is caused by:
    The user requested spras 0.0.1 (from \spras)
    spras[dev] 0.0.1 depends on spras 0.0.1 (from \spras)
To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```
It looks like the command tries to install the dev and non-dev versions. Using
```
$ python -m pip install -e .[dev]
```
succeeded. If you agree that is the correct command, can you please update the documentation accordingly?

I checked the machine learning figures and Cytoscape session produced by the default Snakemake workflow, and everything still looks good.

agitter (approved):

PR #142, opened 2023-12-22
Adding Header Lines

ntalluri:


ntalluri:
2 things left:

1) For some reason I am not able to remove the header line of each files from the cytoscape files.
2) I had to find a way to ignore the header lines in summary.py, but now if there is an emtpy fil, I ignore, which I don't think is the best way to deal with this/

ntalluri:
I still cannot get the header line to be skipped, it might be something on my end version wise, but I'm not sure

ntalluri:
rebuild and retag the image for docker

agitter:
I resolved a merge conflict that occurred after updating the py4cytoscape image tag.

Were you able to get Cytoscape working now?

ntalluri:
Let me check after the merge, but before the merge I was able to resolve it.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

spras/analysis/ml.py
```diff
@@ -44,7 +44,7 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
             edges = []
             for line in lines:
                 parts = line.split('\t')
-                if len(parts) > 0:  # in case of empty line in file
+                if len(parts) >= 4:  # in case of empty line in file or line doesn't include all values
```
agitter, line 47: Which case were we checking for here? Blank lines at the end of the file?

If a line has 2 parts, such as nodes but no rank or direction, should we raise an error instead of continuing?

spras/analysis/ml.py
```diff
@@ -55,7 +55,10 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
                         # node order does matter for directed edges
                         edges.append(DIR_CONST.join([node1, node2]))
                     else:
-                        ValueError(f"direction is {direction}, rather than U or D")
+                        if direction == 'Direction': # if reading the header
```
agitter, line 58: This looks like it can be combined into an `elif` so we don't have the additional level of nesting.

spras/domino.py
```diff
@@ -205,8 +205,10 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
             edges_df['source'] = edges_df['source'].apply(post_domino_id_transform)
             edges_df['target'] = edges_df['target'].apply(post_domino_id_transform)
             edges_df = reinsert_direction_col_undirected(edges_df)
-
-        edges_df.to_csv(standardized_pathway_file, sep='\t', header=False, index=False)
+            edges_df.columns = ['Node1', 'Node2', 'Rank', 'Direction']
+            edges_df.to_csv(standardized_pathway_file, sep='\t', header=True, index=False)
+        else:
+            edges_df.to_csv(standardized_pathway_file, sep='\t', header=None, index=False)
```
agitter, line 211: Will this output a completely blank file? Should we output the header row now instead? We could do that without pandas if it is hard to do with pandas.

spras/allpairs.py
```diff
@@ -112,6 +112,8 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
         @param standardized_pathway_file: the same pathway written in the universal format
         """
         df = pd.read_csv(raw_pathway_file, sep='\t', header=None)
+
         df['Rank'] = 1  # add a rank column of 1s since the edges are not ranked.
         df = reinsert_direction_col_undirected(df)
-        df.to_csv(standardized_pathway_file, header=False, index=False, sep='\t')
+        df.columns = ['Node1', 'Node2', 'Rank', 'Direction']
```
agitter, line 118: I'm trying to think about where to document that this header row is now part of the canonical pathway output. We don't have that type of user documentation anywhere yet. In the meantime, we can update the parts of the contributing guide that refer to `parse_output`.

agitter (commented):
The changes look good overall. My initial review focused only on the file updates. I will still test the updated code locally before we merge.

I built the v3 py4cytoscape image and pushed it to DockerHub.

spras/analysis/ml.py
ntalluri: I added a check if a line in one of the files has > 4 or < 4 values in a line, and raise an error

spras/domino.py
ntalluri: Randomly the output file for domino (and for only domino) will add a '\n' to the file when header = True. This was our version of a quick fix to allow for that to not happen.

spras/analysis/ml.py
```diff
@@ -54,9 +54,10 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
                     elif direction == "D":
                         # node order does matter for directed edges
                         edges.append(DIR_CONST.join([node1, node2]))
-                    else:
-                        ValueError(f"direction is {direction}, rather than U or D")
-
+                    elif direction != 'Direction' and direction != 'U' and direction != 'D':
+                        raise ValueError(f"direction is {direction}, rather than U or D")
+                elif len(parts) > 4 or (len(parts) < 4 and len(parts) > 0): # empty line in file is okay
```
agitter, line 59: Does this logic simplify to `len(parts) != 0`? The length has to be non-negative, and if we are at this line we know it is not 4.

test/ml/test_ml.py
```diff
@@ -21,10 +22,26 @@ def setup_class(cls):
...
+    def test_summarize_networks_less_values(self):
+        with pytest.raises(ValueError):
+            ml.summarize_networks([INPUT_DIR + 'test-data-unexpected-amount-of-values/less.txt'])
+
+    def test_summarize_networks_more_values(self):
+        with pytest.raises(ValueError):
+            ml.summarize_networks([INPUT_DIR + 'test-data-unexpected-amount-of-values/more.txt'])
+
+    def test_summarize_networks_empty_line(self):
+        with pytest.raises(ValueError):
```
agitter, line 38: Why does this test throw the ValueError? The comment indicates that blank lines are okay and we should skip over them.

spras/analysis/ml.py
agitter: We talked about having an initial step to process the header row. Then we continue and interate, processing the rest of the edges and requiring there are always 4 parts.

spras/pathlinker.py
```diff
@@ -139,4 +139,5 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
         # What about multiple raw_pathway_files
         df = pd.read_csv(raw_pathway_file, sep='\t').take([0, 1, 2], axis=1)
         df = reinsert_direction_col_directed(df)
-        df.to_csv(standardized_pathway_file, header=False, index=False, sep='\t')
+        df.columns = ['Node1', 'Node2', 'Rank', "Direction"]
+        df.to_csv(standardized_pathway_file, header=True, index=False, sep='\t')
```
agitter, line 143: We want to better understand the behavior of `to_csv` when the data frame is empty. Our intended output is the header row instead of an empty file.

test/ml/test_ml.py
ntalluri: now empty lines are not okay

spras/pathlinker.py
ntalluri: I was testing what was going on with oi1. When we read in empty files, we catch the pd.errors.EmptyDataError error and don't write any df, which is most likely why no header is written. I checked same error on mcf which also showed that no df was being created when an empty file is read in. 

I changed the code to allow an empty df to be created and written out for oi1, but I want to double-check it before updating all of the other wrapper codes to do the same thing.

ntalluri: I also just noticed that all the headers for the output pathway files aren't the same, so I will need to fix that too

ntalluri:
I've finished the review/comments of the pull request. When you review it again, could you pay special attention to the domino section? I'm not sure that my changes are that good but I was hesitant to make extensive changes to the code.

ntalluri:
Ready for Code review

ntalluri:
Update contributing guide

agitter:
We need to revisit https://github.com/Reed-CompBio/spras/pull/143#discussion_r1480856302 before merging this because I expect to merge #143 first

agitter:
We should add documentation of the standard pathway output format now that it includes rank, direction, and universal header rows. Where would that go? In the `doc` directory?

It can copy the input file example from `input/README.md` and add the header row.

ntalluri:
the `doc` directory seems like the right place for now.

agitter:
@ntalluri I re-reviewed all of the files and pushed a few commits with code cleanup. Can you please review my changes to see if they look okay?

If the GitHub actions tests pass, then I will run the workflow locally to inspect if the output pathways with headers look good. It is hard to be fully confident all of the parse output logic will work as intended with empty and non-empty pathways for all algorithms.

ntalluri:
The changes look good. It seems like the parse output logic is working how I thought it would, but I'm sure what you mean on it not being fully confident on it working.

agitter:
I meant that even though the parse output code changes look good, I don't fully remember all of the behaviors of the various pathway reconstruction algorithms. There could be some corner cases I'm forgetting that would affect how they write headers or deal with empty pathways. It's hard to consider all the possibilities in a big change like this.

ntalluri:
One case to consider is if the raw pathway file only containing a header line and nothing else. 

It seems like the code already deals with this, but I haven't written any test cases for it.

agitter:
That's a good example of algorithm-specific behavior. Some algorithms may output an empty file. Others may output a header only. Their output parsers prior to this change should have already handled that though.

agitter (approved):
I ran the config and egfr workflows locally and reviewed the Cytoscape files, raw pathways, and standardized pathways. Everything looked okay.

PR #143, opened 2023-12-23
Debugging ML

ntalluri:


ntalluri:
I was seeing these messages pop up when I was running the ML post processing on pathlinker: 

Running Docker containers
/opt/anaconda3/envs/spras/lib/python3.8/site-packages/sklearn/decomposition/_pca.py:527: RuntimeWarning: invalid value encountered in divide
  explained_variance_ratio_ = explained_variance_ / total_var
/opt/anaconda3/envs/spras/lib/python3.8/site-packages/seaborn/matrix.py:615: UserWarning: Attempting to set identical left == right == 0 results in singular transformations; automatically expanding.
  ax.set_xlim(0, max_dependent_coord * 1.05)
/opt/anaconda3/envs/spras/lib/python3.8/site-packages/scipy/cluster/hierarchy.py:2845: UserWarning: Attempting to set identical bottom == top == 0 results in singular transformations; automatically expanding.
  ax.set_ylim([0, dvw])
  
  However these messages describe what I would expect to see in the files, which I do see. So I think the code is working

spras/analysis/ml.py
```diff
@@ -82,9 +82,11 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
     concated_df = concated_df.fillna(0)
     concated_df = concated_df.astype('int64')
 
-    # don't do ml post processing if there is an empty dataframe
+    # don't do ml post processing if there is an empty dataframe or the number of samples is <= 1
     if concated_df.empty:
-        raise ValueError("The summarize network dataFrame is empty.\nEnsure that the output files and configuration parameters are correct and non-empty to produce a non-empty dataframe for ml post processing.")
+        raise ValueError("ML post-processing cannot proceed because the summarize network dataFrame is empty.\nCheck that the output files and configuration parameters are correct and non-empty to produce a non-empty dataframe for ml post processing.")
+    if min(concated_df.shape) <= 1:
+        raise ValueError (f"ML post-processing cannot proceed as the available number of samples is insufficient. The process requires more than one sample, but currently, there are only {min(concated_df.shape)} samples.")
```
agitter, line 89: I suggest rephrasing "samples" to "pathways" because that may be more familiar to the user.

spras/analysis/ml.py
```diff
@@ -82,9 +82,11 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
     concated_df = concated_df.fillna(0)
     concated_df = concated_df.astype('int64')
 
-    # don't do ml post processing if there is an empty dataframe
+    # don't do ml post processing if there is an empty dataframe or the number of samples is <= 1
     if concated_df.empty:
-        raise ValueError("The summarize network dataFrame is empty.\nEnsure that the output files and configuration parameters are correct and non-empty to produce a non-empty dataframe for ml post processing.")
+        raise ValueError("ML post-processing cannot proceed because the summarize network dataFrame is empty.\nCheck that the output files and configuration parameters are correct and non-empty to produce a non-empty dataframe for ml post processing.")
```
agitter, line 87: Let's help the user by making the message more instructive and suggest setting `include: true` in the configuration file to avoid this error.

agitter (commented):
This general strategy for fixing the errors looks good.

test/ml/input/test-data-single/single.txt
```diff
@@ -0,0 +1 @@
+L	M	1	U
```
agitter, line 1: If #142 is merged before this, we should add the header line here

test/ml/test_ml.py
```diff
@@ -25,6 +26,14 @@ def test_summarize_networks(self):
         dataframe.to_csv(OUT_DIR + 'dataframe.csv')
         assert filecmp.cmp(OUT_DIR + 'dataframe.csv', EXPECT_DIR + 'expected-dataframe.csv', shallow=False)
 
+    def test_summarize_networks_empty(self):
+        with pytest.raises(ValueError): #raises error if empty dataframe is used for post processing
+            ml.summarize_networks([INPUT_DIR + 'test-data-empty/empty.txt'])
```
agitter, line 31: Should there be an empty file added with this name? I don't see it in the list of files to add in the pull request. Currently I expect this to return an OSError instead of the new ValueError.

agitter: I believe this test isn't running. It has the same name as the next test, so the function is redefined.

ntalluri:
I've finished the review/comments of the ml bug pull request.

ntalluri:
On VSCode, I am getting a weird highlighting/lighting issue where after the df is concatted in the ml summarize_networks function, the rest of the lines after it are shown as unreachable. However when I was testing the code, it does seem like it is reachable and the function is returning the dataframe. If you could look and see if anything weird is happening on your end when reviewing again, that would be great. 

```
concated_df = pd.concat(edge_dataframes, axis=1, join='outer') # all lines of code after this line are "unreachable"
concated_df = concated_df.fillna(0)
concated_df = concated_df.astype('int64')

    # don't do ml post processing if there is an empty dataframe or the number of samples is <= 1
    if concated_df.empty:
        raise OSError("ML post-processing cannot proceed because the summarize network dataFrame is empty.\nWe suggest setting ml's include: true in the configuration file to false to avoid this error.")
    if min(concated_df.shape) <= 1:
        raise OSError (f"ML post-processing cannot proceed because the available number of pathways is insufficient. The ml post processing requires more than one pathway, but currently, there are only {min(concated_df.shape)} pathways.")

return concated_df
```

ntalluri:
Ready for code review

agitter:
> On VSCode, I am getting a weird highlighting/lighting issue where after the df is concatted in the ml summarize_networks function, the rest of the lines after it are shown as unreachable. However when I was testing the code, it does seem like it is reachable and the function is returning the dataframe. If you could look and see if anything weird is happening on your end when reviewing again, that would be great.
> 
> ```
> concated_df = pd.concat(edge_dataframes, axis=1, join='outer') # all lines of code after this line are "unreachable"
> concated_df = concated_df.fillna(0)
> concated_df = concated_df.astype('int64')
> 
>     # don't do ml post processing if there is an empty dataframe or the number of samples is <= 1
>     if concated_df.empty:
>         raise OSError("ML post-processing cannot proceed because the summarize network dataFrame is empty.\nWe suggest setting ml's include: true in the configuration file to false to avoid this error.")
>     if min(concated_df.shape) <= 1:
>         raise OSError (f"ML post-processing cannot proceed because the available number of pathways is insufficient. The ml post processing requires more than one pathway, but currently, there are only {min(concated_df.shape)} pathways.")
> 
> return concated_df
> ```

I don't see this in PyCharm

test/ml/test_ml.py
agitter: Now I see that https://github.com/Reed-CompBio/spras/blob/master/test/ml/input/test-data-empty/empty.txt is already in the repo

PR #145, opened 2024-02-12
Add new SPRAS image and handle unpacked singularity images

jhiemstrawisc:
This PR aims to accomplish a few things:
1. The overarching goal is to make SPRAS something that we can run in distributed environments where containers don't necessarily have any privilege (like the OSPool). This requires handling singularity/apptainer images in a slightly different way, where instead of converting and running a `.sif` file directly, we first "unpack" the sif into a directory. Hence the new `unpack_singularity` config option.
2. A side effect of `1` is that we need a portable runtime environment that can run SPRAS in these distributed settings. The conda environment doesn't handle installations of apptainer/docker, so this adds a new container `docker-wrappers/SPRAS` that can be used to this end. 
3.  The next big step is going to be returning to the Snakemake executor for HTCondor, which will require a relatively new version of Snakemake (definitely in the `8.X.X` range). In updating the version here, I had to chase a few other version incompatibilities. I'm sure there are more to find.


jhiemstrawisc:
@agitter, I've tested this and it seems to work when I run it from an OSPool EP. Since I haven't yet set up the actual building of `reedcompbio/spras:v1`, you're welcome to use `jhiemstra/spras:final-v1` as the container in the HTCondor submit file if you want to test running in the CHTC pool (I haven't tested there, but I suspect it will work since this was working in what constitutes an even more locked-down environment). Otherwise, you can let me know if my container-building instructions were clear enough to follow along with!

I'm opening this as a draft for now, until we've had a chance to discuss further.

jhiemstrawisc:
One thing I'm unsure about is how garbage collecting unpacked sif images might work. I believe other images, both docker and apptainer, are added to some kind of cache that has its own garbage collection rules. However, these unpacked images are written out to disk, so they have the potential to clutter environments that aren't already ephemeral (as would be the case with container-in-container runs). Without some semi-significant refactoring, it's not obvious to me how we can return to delete them. 

Two things to consider:
- Is this our problem? Or are we fine leaving these unpacked containers around. Presumably someone using the tool might want to run things repeatedly, in which case image persistence prevents the user from re-pulling/unpacking things, which should offer a slight speedup. I'm not opposed to saying "if you use this feature, you're responsible for deleting unwanted byproducts." There aren't currently too many containers anyway, and most of them are under 1GB.
- Do we envision anyone using this feature outside of container-in-container settings? If not, then I'm even more okay not cleaning things up -- the outer container is most likely short lived, and so when it disappears so too do the inner unpacked images.

docker-wrappers/SPRAS/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+When you're ready to run SPRAS as an HTCondor workflow, navigate to the `spras/docker-wrappers/SPRAS` directory and run `condor_submit spras.sub`. This will
+submit SPRAS to HTCondor as a single job with as many cores as indicated by the `request_cpus` line in `spras.sub`, using `example_config.yaml` as the
+SPRAS configuration file. Note that you can alter the configuration file to test various workflows, but you should leave `unpack_singularity = true`,
+or it is likely the job will be unsuccessful. By default, the `example_config.yaml` runs everything except for `cytoscape`, which appears to fail periodically
+in HTCondor.
+
+To monitor the state of the job, you can run `condor_q` for a snapshot of how the job is doing, or you can run `condor_watch_q` if you want realtime updates (you
+should never run something like `watch condor_q`, because this places extraneous strain on the AP). Upon completion, the `output` directory from the workflow
```
agitter, line 48:
```suggestion
To monitor the state of the job, you can run `condor_q` for a snapshot of how the job is doing, or  `condor_watch_q` if you want realtime updates. Upon completion, the `output` directory from the workflow
```

I suggest removing the bad command. SPRAS users are highly unlikey to think of that and more likely to accidentally copy/paste the wrong command if it is here than think of it on their own.

docker-wrappers/SPRAS/README.md
```diff
@@ -0,0 +1,54 @@
...
+should be returned as `spras/docker-wrappers/SPRAS/output`, along with several files containing the workflows logging information (anything that matches `spras_*`
```
agitter, line 49:
```suggestion
should be returned as `spras/docker-wrappers/SPRAS/output`, along with several files containing the workflow's logging information (anything that matches `spras_*`
```

docker-wrappers/SPRAS/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Or to temporarily override your system's default by exporting the environment variable:
+```
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
+```
+(This environment variable can then be cleared by running `unset DOCKER_DEFAULT_PLATFORM` to return your system to its default)
+
+
+## Testing
```
agitter, line 33: I submitted a test job following these instructions. They were clear to me, though I am experienced with HTCondor. The job is still idle.

Update: I saw too late that I need to change the `container_image` in the sub file for testing. I'll try again later.

docker-wrappers/SPRAS/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The folder `docker-wrappers/SPRAS` also contains several files that can be used to test this container on HTCondor. To test the `spras` container
+in this environment, first login to an HTCondor Access Point (AP). Then, from the AP clone this repo:
+```
+git clone git@github.com:Reed-CompBio/spras.git
```
agitter, line 38: Is HTTPS cloning easier for novices?

docker-wrappers/SPRAS/README.md
```diff
@@ -0,0 +1,54 @@
+# SPRAS Docker image
+
+## Building
+A Docker image for SPRAS that is available on [Dockerhub]()
```
agitter, line 4:
```suggestion
A Docker image for SPRAS that is available on [DockerHub](https://hub.docker.com/repository/docker/reedcompbio/spras)
```

I built and pushed the image with the `latest` tag. I'll push the versioned verion once we decide what versioning scheme to use.

agitter (changes requested):
My initial pass focused on testing the HTCondor job, Docker image build, and conda environment. I haven't gone through the code changes in detail yet.

docker-wrappers/SPRAS/README.md
agitter: My updated job is still idle. This one uses `docker://reedcompbio/spras:latest`. `condor_q -better-analyze` reports

```
-- Schedd: ap2001.chtc.wisc.edu : <128.105.68.112:9618?...
The Requirements expression for job 130053.000 is

    (TARGET.PoolName == "CHTC") && ((HAS_SINGULARITY == true)) && TARGET.HasContainer && TARGET.HasDockerURL &&
    (TARGET.Disk >= RequestDisk) && (TARGET.Memory >= RequestMemory) && (TARGET.Cpus >= RequestCpus) && (TARGET.HasFileTransfer)

Job 130053.000 defines the following attributes:

    RequestCpus = 4
    RequestDisk = 16777216 (kb)
    RequestMemory = 4096 (mb)

The Requirements expression for job 130053.000 reduces to these conditions:

         Slots
Step    Matched  Condition
-----  --------  ---------
[1]           0  HAS_SINGULARITY == true

No successful match recorded.
Last failed match: Wed Mar 20 14:22:11 2024

Reason for last match failure: no match found

130053.000:  Run analysis summary ignoring user priority.  Of 447 machines,
    447 are rejected by your job's requirements
      0 reject your job because of their own requirements
      0 match and are already running your jobs
      0 match but are serving other users
      0 are able to run your job

WARNING:  Be advised:
   No machines matched the jobs's constraints
```

@jhiemstrawisc have you been able to run this in the CHTC pool?

jhiemstrawisc: Interesting, I've only run on the PATh facility at this point (which was close enough to the true OSPool for my purposes, that I could tell). It's interesting that CHTC machines don't advertise singularity. I'll look into this.

jhiemstrawisc: Sigh. It turns out the workflow can handle apptainer-in-apptainer, but apptainer doesn't work unless the launching user has an entry in `/etc/passwd`. CHTC launches this job under the Docker runtime with a random UID but doesn't the corresponding entry in `/etc/passwd`, so apptainer bricks. Without having to build an extra `.sif` for SPRAS itself, I'm trying to find a way to tell condor "yes, I'm giving you a docker URL, but I actually want you to launch the container with apptainer"

jhiemstrawisc:
@agitter Updates for this:

As it stands, this submit file is working for me from AP2002, and submits jobs to the OSPool. However, there's some container funkiness that I'm still looking into. In particular, running the submit file that points to an image on Dockerhub very frequently causes the job to fail with error `/srv//spras.sh: line 10: snakemake: command not found`. In the OSPool, most (if not all) Docker images are immediately converted to Apptainer, and something about this process is overwriting `/usr/local/bin` in the container. Unfortunately, that's where things like snakemake are supposed to be installed! This shouldn't be happening, so I'd like to keep the bits about using `container_image = docker://reedcompbio/spras:v0.1.0` for now, with the note about how to work around this error in the README.

I've also observed that some snakemake jobs intermittently fail on the EPs, but often succeed after a re-run. That's why I'm using `--retries 3` in the `spras.sh` wrapper. Hopefully that leads to more reliable execution.

To test things yourself, follow the container build instructions from `spras/docker-wrappers/SPRAS/README.md`. Until I can figure out what's going on with OSPool Docker containers, it might be best to immediately convert to `.sif` using the provided instructions, and then run the workflow.

jhiemstrawisc:
On a side note, when we merge this, we should make sure we actually build/publish `reedcompbio/spras:v0.1.0`.

spras/analysis/ml.py
```diff
@@ -90,7 +90,7 @@ def create_palette(column_names):
     """
     # TODO: could add a way for the user to customize the color palette?
     custom_palette = sns.color_palette("husl", len(column_names))
-    label_color_map = {label: color for label, color in zip(column_names, custom_palette)}
+    label_color_map = {label: color for label, color in zip(column_names, custom_palette, strict=False)}
```
agitter, line 93:
```suggestion
    label_color_map = {label: color for label, color in zip(column_names, custom_palette, strict=True)}
```
Something would have gone wrong if these don't match.

spras/config.py
```diff
@@ -179,7 +189,7 @@ def process_config(self, raw_config):
                 run_list_tuples = list(it.product(*all_runs))
                 param_name_tuple = tuple(param_name_list)
                 for r in run_list_tuples:
-                    run_dict = dict(zip(param_name_tuple, r))
+                    run_dict = dict(zip(param_name_tuple, r, strict=False))
```
agitter, line 192:
```suggestion
                    run_dict = dict(zip(param_name_tuple, r, strict=True))
```

agitter (changes requested):
I built and pushed `reedcompbio/spras:v0.1.0`. The test job in `spras.sub` worked for me as well when submitted from `ap2001`.

The overall design, readme, and test instructions look great. My remaining comments here are pretty minor.

There are a couple things to discuss with the broader team. One is the environment updates. The other is whether to make this an initial release in the GitHub repo. I'm in favor of that now that we'll have tagged Docker images.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "spras"
-version = "0.0.1"
+version = "0.1.0"
```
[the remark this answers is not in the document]
jhiemstrawisc: Let's double check with the rest of the group during our next meeting, but I'm in favor of a real release.

agitter:
We believe the latest test failure is due to Docker pull rate limits. The workflow and tests run for me locally.

PR #149, opened 2024-02-27
Remove r and g from OI1 params

### LINKED ISSUE #146: Fix Omics Integrator 1 parameters in config.yaml
The example Omics Integrator 1 parameter values for `r` and `g` in `config.yaml` are set to bad values. If someone were to copy those for a real dataset, it would likely give poor results. It would be better to delete them and use the defaults from https://github.com/fraenkel-lab/OmicsIntegrator#forest-input-files-and-parameters.

agitter:
Closes #146

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -45,10 +45,8 @@ algorithms:
         params:
               include: true
               run1:
-                  r: [5]
                   b: [5, 6]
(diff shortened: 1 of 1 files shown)


ntalluri (commented):
everything looks good

PR #160, opened 2024-06-10
Post processing / ML Code  - Aggregate per algorithm

ntalluri:


ntalluri:
Failed due to ML Bug https://github.com/Reed-CompBio/spras/pull/143

config/config.yaml
```diff
@@ -134,7 +134,10 @@ analysis:
         include: true
       # Machine learning analysis (e.g. clustering) of the pathway output files for each dataset
       ml:
+        # ml analysis per dataset
         include: true
+        # required; adds ml analysis per algorithm output
```
agitter, line 139: Maybe this should be optional and default to false?

I haven't looked through the code line-by-line yet. Is the behavior that if this is true, we will get the overall summaries as before and then additionally compute these algorithm-specified summaries? That is what I would expect.

agitter (commented):
I merged the changes from #143 and left some high level comments before reviewing the code in detail.

agitter:
We may need to think more about the errors we added in #143 when there are not enough pathways for the ML analysis. Now that we can aggregate by algorithm, that will be a common occurrence. Some algorithms don't have parameters so there will always be a single pathway generated. Should we inspect the number of parameter combinations in the config file and only try running ML analysis for those algorithms with multiple parameter combinations?

ntalluri:
Would the snake make command --keep-going be sufficient enough and allow the error to be printed out to the command line?

agitter:
> Would the snake make command --keep-going be sufficient enough

I would rather not require users to add another argument to Snakemake at the command line in order to successfully execute fairly common configurations. It seems like it would lead to a lot of confusion if forgetting that option causes the workflow to crash, so if we can figure out a strategy for writing default outputs or avoiding calling the per-algorithm aggregation when it won't work, that would be better.

ntalluri:
I updated the Snakemake workflow to only run for algorithms that are using multiple parameters when it comes to ml-aggregate

config/config.yaml
```diff
@@ -134,7 +134,11 @@ analysis:
         include: true
       # Machine learning analysis (e.g. clustering) of the pathway output files for each dataset
       ml:
+        # ml analysis per dataset
         include: true
+        # adds ml analysis per algorithm output
+        # works only on algorithms with multiple parameter combinations chosen
```
agitter, line 140:
```suggestion
        # only runs for algorithms with multiple parameter combinations chosen
        # default value false if not set
```

agitter (changes requested):
There are two types of end-to-end integration tests that would be nice to have, but we aren't set up well to make those kinds of tests currently. They would require running the entire workflow. Those tests:
- given the config file `config.yaml`, confirm that only the expected algorithms produce ensemble files
- confirm that one of the expected algorithm-specific ensembles (e.g. `data1-ml/omicsintegrator1-ensemble-pathway.txt`) is correct

Are these tests too hard in our current framework?

This design is great. I confirmed that if I add another parameter combination to the config file for MEO and then rerun Snakemake, then an ensemble pathway and ML outputs are created for MEO.

Snakefile
```diff
@@ -268,21 +284,44 @@ rule ml_analysis:
...
+        hac_clusters_vertical = SEP.join([out_dir, '{dataset}-ml', '{algorithm}-hac-clusters-vertical.txt']),
+        hac_image_horizontal = SEP.join([out_dir, '{dataset}-ml', '{algorithm}-hac-horizontal.png']),
+        hac_clusters_horizontal = SEP.join([out_dir, '{dataset}-ml', '{algorithm}-hac-clusters-horizontal.txt']),
+        ensemble_network_file = SEP.join([out_dir,'{dataset}-ml', '{algorithm}-ensemble-pathway.txt'])
+    run:
+        summary_df = ml.summarize_networks(input.pathways)
+        ml.pca(summary_df, output.pca_image, output.pca_variance, output.pca_coordinates, **pca_params)
+        ml.hac_vertical(summary_df, output.hac_image_vertical, output.hac_clusters_vertical, **hac_params)
+        ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
+        ml.ensemble_network(summary_df, output.ensemble_network_file)
```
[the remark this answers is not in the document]
agitter: We're saving this for #165

agitter (approved):

Issue #130, opened 2023-09-21
Support dsub as an alternative container backend


agitter:
[dsub](https://github.com/databiosphere/dsub) is a command line tool for running commands in Google Cloud inside a container. @nisthapanda is investigating whether we could use it to run containers in SPRAS because that may be one path toward using SPRAS in the restricted NIH All of US platform. The solution might be a further generalization of our `run_container` utility function.

This is related to #124 and Podman, another possible backend for containers.

agitter:
Nistha has a branch that successfully runs Omics Integrator 1 using dsub in the All of Us environment: https://github.com/nisthapanda/spras/tree/dsub

We don't have immediate plans to generalize this and merge this back into SPRAS. It relies on user-specific information about the environment and location of storage buckets.

agitter:
The changes in #132 that make the registry configuration make this easier to support now.

agitter:
@nisthapanda discovered that dsub requires `bash` to run containers, which isn't [available](https://stackoverflow.com/questions/40944479/docker-how-to-use-bash-with-an-alpine-based-docker-image) in Alpine Linux. I'm going to test adding `bash` to PathLinker to see if that resolves her problem.

agitter:
Closed by #195

PR #164, opened 2024-06-21
Add bash to PathLinker image

agitter:
This updates the PathLinker Docker image to include bash as described in https://github.com/Reed-CompBio/spras/issues/130#issuecomment-2182842994

Before making the change, I versioned the prior image as v1 and pushed it to DockerHub. This update is now v2.


agitter:
Expanding this pull request to also update AllPairs, which is Alpine based and has the same problem.

PR #169, opened 2024-07-08
Add Dataset Label Checking

ntalluri:


ntalluri:
I'm not sure how to efficiently and correctly test this change. Should I test every potential way a user could make the dataset label?

test/test_config.py
```diff
@@ -102,3 +102,6 @@ def test_config_container_registry(self):
         test_config["container_registry"]["owner"] = ""
         config.init_global(test_config)
         assert (config.config.container_prefix == config.DEFAULT_CONTAINER_PREFIX)
+
+    def test_dataset_label(self):
```
agitter, line 106: We don't need to test everything. Test coverage should make sure that ~2 valid labels pass and 2-3 invalid labels raise ValueErrors. Example invalid labels can contain `-`, another special character like `$`, and a Unicode symbol like `✉`.

We'll need to think about the most efficient way to test these. Do we need to write separate .yml files for each, or can we use a `sed` like operation to create temp copies of the `config.yml` file with modified dataset labels and then parse those?

agitter (commented):
Testing ideas

test/test_config.py
ntalluri: I tried something similar to the functions above in the file. It seems like I can just overwrite the temp_config and initialize the config to trigger the error.

agitter (commented):
The testing approach you came up with is very clean. These are great.

It is a little unusual that `_` would be allowed as a dataset label, but let's leave that behavior in order to keep the regex simple.

We should still update `config.yml` to document the label name requirements. Besides that, I'll do some local testing and it will be ready.

ntalluri:
Is ready to review

PR #170, opened 2024-07-09
Evaluation Implementation

ntalluri:


Snakefile
```diff
@@ -153,6 +156,10 @@ rule merge_input:
         dataset_dict = get_dataset(_config.config.datasets, wildcards.dataset)
         runner.merge_input(dataset_dict, output.dataset_file)
 
+# TODO: add a merge input for gold standard data?
```
agitter, line 159: Yes, we will need an equivalent of merge input to gather all of the text files per gold standard dataset and create the pickled gold standard object.

config/config.yaml
```diff
@@ -115,6 +115,13 @@ datasets:
       other_files: []
       # Relative path from the spras directory
       data_dir: "input"
+    
+gold_standard:
+    -
+      label: gs
```
agitter, line 121: Will we have the same constraints on labels here in terms of special characters? Let's remember to add a comment about that as documentation and check validity when parsing.

spras/evaluation.py
```diff
@@ -0,0 +1,46 @@
...
+
+        single_node_table = pd.read_table(os.path.join(data_loc, node_file))
+        self.node_table = single_node_table
+
+        # self.node_table = pd.DataFrame(node_set, columns=[self.NODE_ID])
+        # for loop? and read in node dataset into a pandas df
+
+def percision_recall():
+    None
+    # https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html
```
agitter, line 46: This is the best metric to use if we have node ranks. Should we start with the simplest thing and evaluate the entire set of nodes on the predicted pathway because we usually don't have node ranks? Then we could use metrics like precision, recall, F1, and MCC.

agitter (commented):
A few initial thoughts.

spras/evaluation.py
ntalluri: I'll try using precision first

agitter: Noting here that @sumedhars is also prototying an implementation of node ranking and precision-recall curves in her project fairly soon. We can coordinate because if it works for her, she can contribute the code here or in a follow up pull request.

ntalluri:
The first pass works for running all datasets over all gold standard data. Next pass will allow for a user to set which datasets to run over gold standard

ntalluri:
A third pass: allow for evaluation code to truly be optional

Snakefile
```diff
@@ -27,13 +28,15 @@ hac_params = _config.config.hac_params
 FRAMEWORK = _config.config.container_framework
 print(f"Running {FRAMEWORK} containers")
 
-# Return the dataset dictionary from the config file given the label
-def get_dataset(_datasets, label):
+# Return the dataset or goldstandard dictionary from the config file given the label
+def get_dict(_datasets, label):
```
agitter, line 32: It makes sense to reuse the same function for both datasets and gold standards. However, this function name is quite broad and should still have something to do with data.

Snakefile
```diff
@@ -150,9 +156,24 @@ rule merge_input:
     output: dataset_file = SEP.join([out_dir, '{dataset}-merged.pickle'])
     run:
         # Pass the dataset to PRRunner where the files will be merged and written to disk (i.e. pickled)
-        dataset_dict = get_dataset(_config.config.datasets, wildcards.dataset)
+        dataset_dict = get_dict(_config.config.datasets, wildcards.dataset)
         runner.merge_input(dataset_dict, output.dataset_file)
 
+# Return all files used in the gold standard
+def get_goldstandard_dependencies(wildcards):
```
agitter, line 163: Sometimes functions and variables names have `_` between `gold` and `standard` and sometimes they don't. Consistency may help avoid future mistakes.

spras/config.py
```diff
@@ -69,6 +70,8 @@ def __init__(self, raw_config):
         self.unpack_singularity = False
         # A dictionary to store configured datasets against which SPRAS will be run
         self.datasets = None
+        # A dictionary to store configured gold standard data against ouptut of SPRAS runs
+        self.gold_standard = None
```
agitter, line 74:
```suggestion
        self.gold_standards = None
```

Does this need to be plural? There can be multiple gold standards?

spras/config.py
```diff
@@ -140,6 +146,13 @@ def process_config(self, raw_config):
         # Convert to dicts to simplify the yaml logging
         self.datasets = {dataset["label"]: dict(dataset) for dataset in raw_config["datasets"]}
 
+        # TODO: turn into try except
+        self.gold_standard = {goldstandard["label"]: dict(goldstandard) for goldstandard in raw_config["gold_standard"]}
+        for key in self.gold_standard:
+            pattern = r'^\w+$'
+            if not bool(re.match(pattern, key)):
+                raise ValueError(f"Gold Standard label \'{key}\' contains invalid values. Gold Standard labels can only contain letters, numbers, or underscores.")
```
agitter, line 154:
```suggestion
                raise ValueError(f"Gold standard label \'{key}\' contains invalid values. Gold standard labels can only contain letters, numbers, or underscores.")
```

spras/runner.py
```diff
@@ -42,6 +43,15 @@ def merge_input(dataset_dict, dataset_file):
     dataset = Dataset(dataset_dict)
     dataset.to_file(dataset_file)
 
+def merge_gold_standard_input(gs_dict, gs_file):
```
agitter, line 46: This doesn't depend on `algorithm` as an input, so I don't think it needs to go in `runner.py`. Should it be in `evaluation.py`?

agitter (commented):
I'm leaving a few initial comments but skipped `evaluation.py` for now. I'm still looking for the best source of EGFR evaluation data we can use for a test case on pathways created with the `egfr.yml` config file.

ntalluri:
Ready for review

config/config.yaml
```diff
@@ -117,6 +117,20 @@ datasets:
       # Relative path from the spras directory
       data_dir: "input"
 
+gold_standard:
+    -
+      label: gs0
```
agitter, line 122: Do we need to copy the comment "# Labels can only contain letters, numbers, or underscores" as a reminder for the gold standard label?

spras/config.py
```diff
@@ -70,6 +70,8 @@ def __init__(self, raw_config):
         self.unpack_singularity = False
         # A dictionary to store configured datasets against which SPRAS will be run
         self.datasets = None
+        # A dictionary to store configured gold standard data against ouptut of SPRAS runs
```
agitter, line 73:
```suggestion
        # A dictionary to store configured gold standard data against output of SPRAS runs
```

spras/config.py
```diff
@@ -225,6 +240,13 @@ def process_config(self, raw_config):
         self.analysis_include_graphspace = raw_config["analysis"]["graphspace"]["include"]
         self.analysis_include_cytoscape = raw_config["analysis"]["cytoscape"]["include"]
         self.analysis_include_ml = raw_config["analysis"]["ml"]["include"]
+        self.analysis_include_evalution = raw_config["analysis"]["evaluation"]["include"]
+
+        # COMMENT: the code will run correctly without this section below due to empty dict in try except above
+        # TODO: decide if this part is needed
+        if self.gold_standards == {} and self.analysis_include_evalution == True:
+            print("Gold standard data not provided. Evaluation analysis cannot run.")
```
agitter, line 248: Are you saying the workflow would still run without errors because we set a default dict? If a user sets evaluation true and doesn't provide a gold standard, I think we should raise an error here with this friendly error message. It means they have a mistake in their config file.

spras/config.py
```diff
@@ -225,6 +240,13 @@ def process_config(self, raw_config):
...
+            self.analysis_include_evalution = False
```
agitter, line 250: Do we also need to do error-checking to see if all the dataset labels in the gold standards are existing datasets labels?

spras/evaluation.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Loads gold standard files from gold_standard_dict, which is one gold standard dataset
+        dictionary from the list in the config file with the fields in the config file.
+        Populates node_table.
+
+        node_table is a single column of nodes pandas table.
+
+        returns: none
+        """
+        self.label = gold_standard_dict["label"] # COMMENT: cannot be empty, will break with a NoneType exception
```
agitter, line 57: You can remove `COMMENT:` from these, that is implied

spras/evaluation.py
```diff
@@ -0,0 +1,98 @@
...
+        @param file_paths: file paths of pathway reconstruction algorithm outputs
+        @param node_table: the gold standard nodes
+        @param output_file: the filename to save the precision of each pathway
+        """
+        y_true = node_table['NODEID'].tolist()
+        results = []
+
+        for file in file_paths:
+            df = pd.read_table(file, sep="\t", header = 0, usecols=["Node1", "Node2"])
+            y_pred = list(set(df['Node1']).union(set(df['Node2'])))
```
agitter, line 88: It looks like it would be better to keep y_pred a set here. It gets converted back to a set in the next line.

spras/evaluation.py
```diff
@@ -0,0 +1,98 @@
...
+    def precision(file_paths: Iterable[Path], node_table: pd.DataFrame, output_file: str):
+        """
+        Takes in file paths for a specific dataset and an associated gold standard node table.
+        Calculates precision for each pathway file
+        Returns output back to output_file
+        @param file_paths: file paths of pathway reconstruction algorithm outputs
+        @param node_table: the gold standard nodes
+        @param output_file: the filename to save the precision of each pathway
+        """
+        y_true = node_table['NODEID'].tolist()
```
agitter, line 83: Make this a set for faster lookups later?

spras/runner.py
```diff
@@ -42,7 +42,6 @@ def merge_input(dataset_dict, dataset_file):
     dataset = Dataset(dataset_dict)
     dataset.to_file(dataset_file)
 
-
```
agitter, line 45: Add back the deleted line

agitter (changes requested):
Initially I was going to have an EGFR-based test case be part of this initial pull request. Now, I think we should work to merge it soon and add that test case in a follow up pull request to test the actual precision calcuation.

spras/config.py
ntalluri: From that comment, it will not run the evaluation code due to the empty dict created, even if evaluation is set as True.

spras/config.py
ntalluri: The code will break in the snakemake rule merge_input with error `InputFunctionException in rule merge_input in file /Users/nehatalluri/Desktop/research/spras/Snakefile, line 152: Error: KeyError: <the dataset that does not exist>`

There error isn't nice, but it does work as expected when the dataset label does not exist in the config file.

config/config.yaml
```diff
@@ -117,6 +117,20 @@ datasets:
...
+      node_files: ["gs_nodes0.txt"]
+      # edge_files: [] TODO: later iteration
+      data_dir: "input"
+      # Set of datasets (dataset labels) to compare with the specific gold standard dataset
+      datasets: ["data0"]
+    -
+      label: gs1
+      node_files: ["gs_nodes1.txt"]
+      data_dir: "input"
+      datasets: ["data1", "data0"]
```
ntalluri, line 132: @agitter  what would you suggest the behavior should be if `datasets: []`.  Right now it can be empty and snakemake will not run evaluation.

agitter: A warning or nothing is fine

agitter (approved):
I pushed a couple final changes. This is ready to merge once the tests pass.

PR #172, opened 2024-07-19
Enable spras in CHTC with new executor and profile

jhiemstrawisc:
@agitter I think this gets us to the point where we can run a full SPRAS workflow in both the CHTC pool and the OSPool. At least, I've done testing in both and it works for me 😉 

When you've got a chance, can you test out my new instructions and let me know how it goes?


docker-wrappers/SPRAS/spras_profile/config.yaml
```diff
@@ -0,0 +1,11 @@
+jobs: 30
```
agitter, line 1: What does this control? Does it need to align with the number of rules in the Snakemake DAG? Or will Snakemake decide how to allocate rules to this number of jobs?

docker-wrappers/SPRAS/spras_profile/config.yaml
```diff
@@ -0,0 +1,11 @@
...
+executor: htcondor
+configfile: example_config.yaml
+shared-fs-usage: none
+default-resources:
+  job_wrapper: "spras.sh"
+  # If running in CHTC, this only works with apptainer images
+  container_image: "spras.sif"
+  universe: "container"
+  request_disk: "16GB"
+  request_memory: "8GB"
```
agitter, line 11: Per our conversations, we don't know of any way to increase resources for a single job at this point, right? We could modify the Snakefile to modify resources for a single rule, but that would override resources of all instances of that rule. In SPRAS, we may only want the override to apply for certain wildcard values when that rule is run (e.g. one pathway reconstruction algorithm).

docker-wrappers/SPRAS/README.md
```diff
@@ -48,16 +48,66 @@ DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t reedcompbio/spras:v0.1.0 -f
...
+
+Parallelizing SPRAS workflows with HTCondor currently requires an experimental executor for HTCondor that has been forked from the upstream [HTCondor Snakemake executor](https://github.com/jhiemstrawisc/snakemake-executor-plugin-htcondor/tree/spras-feature-dev).
+To get this executor, clone the forked repository using the following:
+```bash
+git clone -b spras-feature-dev https://github.com/jhiemstrawisc/snakemake-executor-plugin-htcondor.git
+```
+
+Then, from your activated `spras` conda environment (important), run:
+```bash
+pip install snakemake-executor-plugin-htcondor/
```
agitter, line 79: Should we add this directly to the conda environment file? It is possible to pip install directly from GitHub, and I believe we could also specify the branch. That would simplify the setup steps.

docker-wrappers/SPRAS/README.md
```diff
@@ -48,16 +48,66 @@ DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t reedcompbio/spras:v0.1.0 -f
...
+
+Resource requirements can be adjusted as needed in `spras_profile/config.yaml`, and HTCondor logs for this workflow can be found in `.snakemake/htcondor`.
+You can set a different log directory by adding `htcondor-jobdir: /path/to/dir` to the profile's configuration.
 
+To run this same workflow in the OSPool, add the following to the profile's default-resources block:
+```
+  classad_WantGlideIn: true
+  requirements: |
+    '(HAS_SINGULARITY == True) && (Poolname =!= "CHTC")'
+```
```
agitter, line 105: I did not test any of these optional configurations.

docker-wrappers/SPRAS/README.md
```diff
@@ -48,16 +48,66 @@ DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t reedcompbio/spras:v0.1.0 -f
...
+
+**Note**: This workflow requires that the terminal session responsible for running snakemake stays active. Closing the terminal will suspend jobs,
```
agitter, line 107: This would be an issue for full workflows with thousands of jobs. We may want to talk to CHTC facilitators about preferred ways to run this. I'm assuming they would not want us to use tmux or screen to keep an active session, but I'm not sure what is preferable. Probably something that runs the Snakemake process with HTCondor itself. That seems too complicated.

docker-wrappers/SPRAS/README.md
```diff
@@ -67,9 +117,11 @@ contain useful debugging clues about what may have gone wrong.
...
-**Note**: In some cases, especially if you're encountering an error like `/srv//spras.sh: line 10: snakemake: command not found`, it may be necessary to convert
+## Creating an Apptainer image for SPRAS
+
+In some cases, especially if you're encountering an error like `/srv//spras.sh: line 10: snakemake: command not found`, it may be necessary to convert
 the SPRAS image to a `.sif` container image before running someplace like the OSPool. To do this, run:
-```
+```bash
 apptainer build spras.sif docker://reedcompbio/spras:v0.1.0
 ```
 to produce the file `spras.sif`. Then, substitute this value as the `container_image` in the submit file.
```
agitter, line 127: We could add `spras.sif` to `.gitignore`

agitter (changes requested):
I was able to follow the instructions and run the example SPRAS workflow in CHTC. It's incredible to see this working in parallel on multiple EPs. I did not carefully scrutinize all of the output and logs. The SPRAS output files I checked did look as expected, however.

_Edit: some of my initial comments were hidden when I finished reviewing, so I rewrote them. Then I noticed and deleted the duplicates._

agitter:
One more suggestion: can you please add a small section to the main SPRAS readme about how it can be run with HTCondor and refer to the instructions here?

docker-wrappers/SPRAS/README.md
```diff
@@ -48,16 +48,66 @@ DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t reedcompbio/spras:v0.1.0 -f
...
+```
+  classad_WantGlideIn: true
+  requirements: |
+    '(HAS_SINGULARITY == True) && (Poolname =!= "CHTC")'
+```
+
+**Note**: This workflow requires that the terminal session responsible for running snakemake stays active. Closing the terminal will suspend jobs,
+but the workflow can use Snakemakes checkpointing to pick up any jobs where they left off.
+
+### Job Monitoring
```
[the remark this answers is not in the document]
jhiemstrawisc: Running `condor_q` requires a second terminal. I'll make a note of that.

docker-wrappers/SPRAS/spras_profile/config.yaml
jhiemstrawisc: I'll add a few comments to the profile, but this controls the maximum number of concurrent jobs allowed into the HTCondor queue. This number isn't hit when running the example config, but might be more useful for larger workflows across multiple data sets. That being said, I think its safe to make this number pretty big.

docker-wrappers/SPRAS/spras_profile/config.yaml
jhiemstrawisc: Unfortunately, that's correct. I haven't figured out a way to apply a set of resource requirements to a job without scoping it to the rule level.

docker-wrappers/SPRAS/README.md
jhiemstrawisc: I've given some thought to this, and I believe the answer is to wrap the workflow in an HTCondor "local universe" job. Local universe allows you to run an executable from the AP as an HTCondor job. I'd like to make implementation of the wrapper HTCondor job a second PR, because it adds a few layers of complexity and feels like scope creep for this PR.

docker-wrappers/SPRAS/README.md
jhiemstrawisc: I tried doing that in another commit to this PR, but it doesn't seem like the `htcondor` python bindings needed by the executor play well with arm64/mac or Windows. If we wanted this to work reliably, we'd need to figure out how to install this feature conditioned on the host OS. As it currently stands, I can't see anyone trying to use this outside of an AP which should pretty much always be a compatible version of Linux, so I'm in favor of punting on this step for the time being.

docker-wrappers/SPRAS/README.md
agitter: Agreed that we'll make this a follow up pull request

docker-wrappers/SPRAS/spras_profile/config.yaml
agitter: We discussed how an advanced user could turn off the resource intensive algorithms and post-processing during an initial run, then enable the remaining algorithms after increasing resources, then enable post-processing. That is complex and we won't document it at this point.

docker-wrappers/SPRAS/README.md
agitter: I confirmed I could not install this on my Windows machine so we should leave the installation outside of the spras environment.
```
$ pip install snakemake-executor-plugin-htcondor/
Processing c:\users\agitter\desktop\madison\collaborators\ritz\snakemake-executor-plugin-htcondor
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
ERROR: Could not find a version that satisfies the requirement htcondor<24.0.0,>=23.4.0 (from snakemake-executor-plugin-htcondor) (from versions: none)
ERROR: No matching distribution found for htcondor<24.0.0,>=23.4.0
```

agitter (approved):
The changes address all my feedback.

@ntalluri did you want to continue testing this before merging or can we merge and you test after?

jhiemstrawisc:
@agitter, Neha and I worked through this together, and I think I came away with a few small improvements. The main issue was that the spras container I point to (v0.1.0) is old enough that using the Snakefile from the current master branch in the workflow causes errors in the container's config module. I'll update the workflow to use the new version you released.

agitter:
> The main issue was that the spras container I point to (v0.1.0) is old enough that using the Snakefile from the current master branch in the workflow causes errors in the container's config module.

That makes me curious how I ran it successfully? I must have been using some old version of SPRAS somewhere in my tests.

jhiemstrawisc:
@agitter I just included a few small updates that are worth looking over. Notably, I changed the default behavior I steer the user toward when selecting an image to use with HTCondor. Previously I pointed them to Docker and told them "use Apptainer if you have issues." The issues seem prevalent, however, so now I just tell them to use Apptainer in the first place.

The second change encourages them to build their own containers to make sure the version of SPRAS installed in the container is compatible with the version they've cloned onto the AP. This was the issue Neha ran into mentioned above.

docker-wrappers/SPRAS/README.md
```diff
@@ -38,27 +38,97 @@ Or to temporarily override your system's default during the build, prepend your
...
+```bash
+apptainer build <new image name>.sif docker://<name of container on DockerHub>
+```
 
+For example, creating an Apptainer image for the `v0.2.0` SPRAS image might look like:
+```bash
+apptainer build spras-v0.2.0.sif docker://reedcompbio/spras:v0.2.0
 ```
+
+After running this command, a new file called `spras-v0.2.0` will exist in the directory where the command was run.
```
agitter, line 58: Is the file called `spras-v0.2.0.sif`?

agitter (commented):
The updated instructions look good. There is one possible typo in a filename, and then we can merge.

docker-wrappers/SPRAS/README.md
jhiemstrawisc: Should be resolved. Thanks for catching it!

Issue #163, opened 2024-06-20
Decouple Ensemble Step from ML Step


ntalluri:
With the updates to the code in https://github.com/Reed-CompBio/spras/pull/160, algorithms with multiple parameter combinations will now run aggregate per-algorithm ML code. As a result, we won't be able to run ensemble code for algorithms with one or fewer parameter combinations. To ensure we have ensemble networks for all algorithms, we need to create a new rule that uses the ensembling code per algorithm, rather than running all the ML code for each algorithm.

agitter:
Closed by #175

PR #175, opened 2024-07-29
Decouple Ensembling

ntalluri:
I got distracted and decoupled the ensembling code https://github.com/Reed-CompBio/spras/issues/163


ntalluri:
(Something to think about) Should I consider modularizing each part of the ML code? I can't recall if I tested each ML function individually with edge cases. There's a possibility that while HAC might work even though PCA fails when dealing with those edge cases. Right now the code will not run HAC after PCA fails assuming that HAC will also fail.

spras/analysis/ml.py
```diff
@@ -86,18 +86,17 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
     concated_df = concated_df.fillna(0)
     concated_df = concated_df.astype('int64')
 
-    # don't do ml post-processing if there is an empty dataframe or the number of samples is <= 1
-    if concated_df.empty:
+    return concated_df
+
+def df_error(dataframe: pd.DataFrame):
```
sumedhars, line 91: an alternative approach to this could be creating a custom exception class for validation errors and having seperate methods that check for empty dfs/incorrect shapes? could help with modularity (but could also be overkill?)

Snakefile
```diff
@@ -93,14 +93,14 @@ def make_final_input(wildcards):
...
-        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-hac-vertical.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-hac-clusters-vertical.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-hac-horizontal.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-hac-clusters-horizontal.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos,algorithm_params=algorithms_with_params))
-        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-ensemble-pathway.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos,algorithm_params=algorithms_with_params))
```
agitter, line 103: It's hard to tell with the long lines. Is the change that `algorithm_params=algorithms_with_params` is not actually needed?

Snakefile
```diff
@@ -291,18 +291,28 @@ rule ml_analysis:
         hac_clusters_vertical = SEP.join([out_dir, '{dataset}-ml', 'hac-clusters-vertical.txt']),
         hac_image_horizontal = SEP.join([out_dir, '{dataset}-ml', 'hac-horizontal.png']),
         hac_clusters_horizontal = SEP.join([out_dir, '{dataset}-ml', 'hac-clusters-horizontal.txt']),
-        ensemble_network_file = SEP.join([out_dir,'{dataset}-ml', 'ensemble-pathway.txt'])
```
agitter, line 294: Can we keep the ability to create a single ensemble pathway out of all pathways from all algorithms. It looks like there is a Snakemake rule for that, but we no longer tell Snakemake to create the output file here.

Snakefile
```diff
@@ -291,18 +291,28 @@ rule ml_analysis:
...
         ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
+
+# Ensemble the output pathways for each dataset
+rule ensemble: 
+    input:
+        pathways = expand('{out_dir}{sep}{{dataset}}-{algorithm_params}{sep}pathway.txt', out_dir=out_dir, sep=SEP, algorithm_params=algorithms_with_params)
+    output:
+        ensemble_network_file = SEP.join([out_dir,'{dataset}-ml', 'ensemble-pathway.txt'])
+    run:
+        summary_df = ml.summarize_networks(input.pathways)
```
agitter, line 307: If running `summariz_networks` twice (for ensembling and ML) ever becomes a runtime bottleneck, we could write its output to disk and resuse it. We should keep this initial version simple though.

spras/analysis/ml.py
agitter: Now that I've read all the code, I think it is okay to keep this as a function. However, I would rename it to `validate_df` or something similar to indicate its role is to help with input validation. It isn't a custom error.

test/ml/test_ml.py
```diff
@@ -83,3 +95,25 @@ def test_ensemble_network(self):
...
+        expected = expected.round(5)
+
+        assert en.equals(expected)
+
+    def test_ensemble_network_empty(self):
+        dataframe = ml.summarize_networks([INPUT_DIR + 'test-data-empty/empty.txt'])
+        ml.ensemble_network(dataframe, OUT_DIR + 'ensemble-network-empty.tsv')
+
+        en = pd.read_table(OUT_DIR + 'ensemble-network-empty.tsv')
+        en = en.round(5)
```
agitter, line 115: The dataframe is empty so you could probably skip the rounding in this test case.

agitter (changes requested):
The overall design is good. I only have minor suggestions.

> Should I consider modularizing each part of the ML code?

This is low priority for me given how many other things we have to work on. If a user ever requests that or has a problem with it, we could implement it.

Snakefile
ntalluri: Yes that is correct. Only algorithm_params=algorithms_with_params is used ml_aggregate_algo because of the function `collect_pathways_per_algo` is grabbing the correct pathways to use. algorithm_params isn't being used anywhere in ensemble_per_algo or ml_aggregate_algo.

ntalluri: Instead, what is sent is a list of algorithms with multiple parameter combinations, which is then assigned to the algorithms parameter. However, for the ensemble part of analysis_include_ml_aggregate_algo, all algorithms are required, which include those with a single parameter combination.

Snakefile
ntalluri: I separated that into its own snakemake rule. This is because ensemble won't be run if any of the functions above it fail.

Snakefile
```diff
@@ -291,18 +291,28 @@ rule ml_analysis:
...
         ml.ensemble_network(summary_df, output.ensemble_network_file)
```
ntalluri, line 308: this is where the single ensemble pathway out of all pathways from all algorithms is created

spras/config.py
```diff
@@ -254,7 +254,7 @@ def process_config(self, raw_config):
             raise ValueError("Evaluation analysis cannot run as gold standard data not provided. "
                              "Please set evaluation include to false or provide gold standard data.")
 
-        if 'aggregate_per_algorithm' not in self.ml_params:
-            self.analysis_include_ml_aggregate_algo = False
-        else:
+        if 'aggregate_per_algorithm' in self.ml_params and self.analysis_include_ml == True:
```
ntalluri, line 257: There was a bug I found when I was running the code that 'aggregate_per_algorithm'  will still run when ML include is set to false.

ntalluri: Now thinking about this behavior, would we want aggregate_per_algorithm to still run if ml include is set to false. Really the include is more of a ml_analysis_per_dataset.

agitter: My interpretation of the config file values is that ml include True/False enables or disables everything. If it is False, then we do not care what `aggregate_per_algorithm` is set to because we are not running any of the ML analyses.

The current code looks like it matches that. However, `self.analysis_include_ml == True` can be simplified to `self.analysis_include_ml`.

Snakefile
agitter: I was mistaken and thought this deleted line was line 95 where the final input list is constructed.

agitter (commented):
I don't have any major comments remaining. I'll test locally and merge if everything looks good.

agitter:
The tests passed locally. However, when I ran with the config file, I was confused about the the data1 outputs. Should we have ML outputs (e.g. PCA output) for Omics Integrator 1? We run it with 4 parameter combinations and the summary table shows:
```
output/data1-omicsintegrator1-params-E3LSEZQ/pathway.txt	3	2	1	2	2	1	1
output/data1-omicsintegrator1-params-NFIPHUX/pathway.txt	0	0	0	0	0	0	0
output/data1-omicsintegrator1-params-SU2S63Y/pathway.txt	3	2	1	2	2	1	1
output/data1-omicsintegrator1-params-V26JBGX/pathway.txt	0	0	0	0	0	0	0
```
I didn't get those outputs.

ntalluri:
When I run OI1 locally, I am getting back outputs for the ml step. 

Are you suggesting that there shouldn't be any ml outputs to start with?  Or that locally you are not getting any ml outputs?

agitter:
When I ran it, I thought I did not get those OI1 ML outputs. However, I tested it again and I did. I could have been mistaken or had some temporary error.

PR #176, opened 2024-07-30
Responsenet Implementation into SPRAS

gabeah:
The wrapper implementation for ResponseNet.

Note: This is actively working, there are many issues that need to be ironed out. This is mainly for easier access for other maintainers to see the code.


gabeah:
Last update from me RE: ResponseNet wrapping.

As it stands, there is still an error with running the docker container. Anna and I were able to run interactively both my Docker image and the ReedCompBio image, but when trying to run it through SPRAS, the following error appeared:
image

Also as it stands, the wrapping is not yet complete. I started the wrapping process and its likely finished, but I was unable to do any testing of the wrapper due to previous errors. Thus, the things that need to be done are as follows:

- [ ] Finish and test the wrapper functions for SPRAS
- [ ] Add ResponseNet as a full fledged algorithm for SPRAS
- [ ] Write Pytests

After creating this comment, I'm going to turn this PR into a draft PR for whenever it may be revived. I am also adding Anna, Tony, Altaf, and Oliver as contributors.

Have a good year y'all o7

tristan-f-r:
[closed in favor of #241]

PR #177, opened 2024-08-16
Modify GHA Docker build tests to clean up after themselves

jhiemstrawisc:
@ntalluri reports several failing CI tests caused by insufficient disk space in the GHA runner's host OS, and it looks like the available disk is being chewed through by the docker build tests.

This PR adds an additional step to each container build test that explicitly deletes the image after it has been built. Hopefully this leads to more efficient use of space.

### DIFF
--- a/.github/workflows/test-spras.yml
+++ b/.github/workflows/test-spras.yml
@@ -94,6 +94,12 @@ jobs:
         tags: latest
         cache_froms: reedcompbio/omics-integrator-1:latest
         push: false
+    - name: Remove Omics Integrator 1 Docker image
+      # Remove the image to prevent the cache from being used. Here we use
+      # `|| true` to prevent the job from failing if the image doesn't exist or
+      # can't be removed for some reason
+      run: docker rmi reedcompbio/omics-integrator-1:latest || true
+
     - name: Build Omics Integrator 2 Docker image
       uses: docker/build-push-action@v1
       with:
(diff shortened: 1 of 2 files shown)


agitter:
We're still failing the "Build Docker images" test with
```
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```

The SPRAS image isn't that large, so I'm puzzled.

jhiemstrawisc:
@agitter, I think I've slimmed this down enough to pass during the build process. I'm still not 100% sure exactly what the source of the sudden failures were, but it's clearly related to some intermediary files created in the `dnf` installations. I don't think there's any harm in leaving the container cleanup between Docker build runs, but I can remove that if you'd like.

agitter:
We can leave the cleanup steps in between the Docker builds. The GitHub Actions builds are working again, so I'm happy with these changes.

However, should we bump spras to v0.1.1 before merging now that we've changed the Dockerfile?

jhiemstrawisc:
Good point, I don't think the current setup gives us a way to create a new image without ingesting all of the changes merged to master since the last image was cut, so this is probably a requirement. We might think about teasing these apart so we can make packaging changes without having to cut a new version every time.

agitter:
We agreed to Justin's proposed workflow for merge, update verison, tag, release, build and push the Docker image.

In a follow up pull request, he'll work on automating some of those steps.

agitter:
I pushed v0.2.0 to DockerHub https://hub.docker.com/repository/docker/reedcompbio/spras/general

Issue #133, opened 2023-10-04
Omics Integrator 2 Testing Error


ntalluri:
When I was writing the tests for parse outputs, I kept getting an error on this line of the code for oi2: df = df[df['in_solution'] == True]  # Check whether this column can be empty before revising this line

I was taking the output from the oi2 test suite, but it was failing because it was missing the 'in_solution' column, yet the column 'in_solution' appears in the raw-pathways for oi2 when SPRAS is run over the datasets. 

There might be an issue with the test suite or merely the result. I haven't checked it out, but it's something we should look at because it's not a current test.

agitter:
This may not be a problem with the Omics Integrator 2 container or run function but rather have to do with its input data and parameters. I modified the test function to use the parameters (b=4, g=0) and input data from the workflow. I was able to get a complete output file with the in_solution column when running the `test_oi2_required` test:
```
protein1	protein2	cost	in_solution
B	A	0.52	True
B	C	0.73	True
```

I'll need to run more tests or read through its source code to understand why this column is sometimes missing.

agitter:
I pushed the branch `oi2-misssing-col` with my initial test dataset changes and explorations and some documentation updates.

I had started looking at where Omics Integrator 2 sets `in_solution` in its source code, but didn't set find anything that helps explain this behavior: https://github.com/agitter/OmicsIntegrator2/blob/command-line/src/graph.py#L335

ntalluri:
When I was grid searching the parameters for EGFR, oi2 kept breaking due to this error. For parameter tuning, it will be helpful to find a solution for this issue soon.

ntalluri:
Examples of headers from the raw pathway file 

`protein1	protein2	cost`

`protein1	protein2	in_solution	cost`

`protein1	protein2	cost	in_solution`

ntalluri:
quick fix, if the first 2 column headers are there, then write an empty file and say it is corrupted

then try to figure what the code is doing

ntalluri:
https://github.com/Reed-CompBio/spras/pull/182

ntalluri:
The cost is always missing when the raw pathway is empty. However when it isn't empty, also happens to be nondeterministic, so it will occasionally not be present at all in the raw pathway.

ntalluri:
[...]

Issue #117, opened 2023-08-15
ML code raises ValueError when run with one empty pathway


agitter:
I tried running ML analysis with a single empty pathway from Omics Integrator 1 and received this error
```
RuleException:
ValueError in file Snakefile, line 265:
Found array with 0 feature(s) (shape=(1, 0)) while a minimum of 1 is required by StandardScaler.
  File "Snakefile", line 265, in __rule_ml_analysis
  File "spras\src\analysis\ml.py", line 115, in pca
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\sklearn\preprocessing\_data.py", line 809, in fit
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\sklearn\preprocessing\_data.py", line 844, in partial_fit
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\sklearn\base.py", line 577, in _validate_data
  File "C:\Users\agitter\.conda\envs\spras\lib\site-packages\sklearn\utils\validation.py", line 918, in check_array
  File "C:\Users\agitter\.conda\envs\spras\lib\concurrent\futures\thread.py", line 57, in run
```

@ntalluri this is a pretty unique corner case, but we should think about the desired behavior.

PR #182, opened 2024-09-03
Omics Integrator 2 Testing Error

ntalluri:


ntalluri:
There is now an error with Aggregate per algorithm for the ml step. It only runs when for algorithms with multiple parameter combinations chosen; however, some of those parameter combinations can lead to outputs that are still nothing, which can lead to the bug error from https://github.com/Reed-CompBio/spras/issues/117

ntalluri:
The issue with the different headers in OI2 is nondeterministic. Although the code functions as intended, the corrupted data happens randomly, causing the ML option `aggregate_per_algorithm` to fail unpredictably if we keep the default as true.

agitter:
Can you tell whether the non-determisitic behavior is the order of 'cost' and 'in_solution' or the presence of those columns? In https://github.com/Reed-CompBio/spras/issues/133#issuecomment-2321590234 I speculated that perhpas the order of those two columns is allowed to be in either order, in which case our quick fix here could sort the columns as long as we get the correct number of columns.

ntalluri:
I can check for the presence of the columns and order. I agree with the idea of sorting the columns as long as the correct number of columns and correct header names exist.

test/parse-outputs/test_parse_outputs.py
```diff
@@ -37,3 +38,17 @@ def test_empty_file(self):
 
             runner.parse_output(algo, test_file, out_file)
             assert filecmp.cmp(OUTDIR + f"{algo}-empty-pathway.txt", EXPDIR + f"empty-pathway-expected.txt", shallow=False)
+
+    def test_oi2_miss_insolution(self):
+        test_file = OI2_EDGE_CASES_INDIR + f"omicsintegrator2-miss-insolution-raw-pathway.txt"
+        out_file = OUTDIR + f"omicsintegrator2-miss-insolution-pathway.txt"
+
+        runner.parse_output('omicsintegrator2', test_file, out_file)
+        assert filecmp.cmp(out_file, EXPDIR + f"empty-pathway-expected.txt", shallow=False)
```
agitter, line 47: Do we need to commit the two new expected output files used in this test case?

For some reason, it looks like GitHub Actions isn't running these new test functions.

spras/omicsintegrator2.py
```diff
@@ -148,14 +148,20 @@ def parse_output(raw_pathway_file, standardized_pathway_file):
...
         else:
             df = pd.read_csv(raw_pathway_file, sep='\t', header=0)
-            df = df[df['in_solution'] == True]  # Check whether this column can be empty before revising this line
-            df = df.take([0, 1], axis=1)
-            df = add_rank_column(df)
-            df = reinsert_direction_col_undirected(df)
-            df.columns = ['Node1', 'Node2', 'Rank', "Direction"]
+            if sorted(df.columns) == sorted_correct_column_names: # if column header names are all correct
+                df = df[df['in_solution'] == True]  # Check whether this column can be empty before revising this line
+                df = df.take([0, 1], axis=1)
```
agitter, line 160: This will have to change now that the columns are in a new order. We should use the column names anyway instead of `take`, which is less readable.

agitter (changes requested):

test/parse-outputs/test_parse_outputs.py
ntalluri: The pathways from expected are files that already existed prior to these new tests. 

I can look over why their not running in GIthub Actions, but they're running locally.

spras/omicsintegrator2.py
agitter: In our meeting discussion we agreed this approach works. However, we may want a comment about what `take` on the first two column is the right thing to do. We can also update the column about what the above line is doing now that we've studied `in_solution` more and know what it is doing.

test/parse-outputs/test_parse_outputs.py
agitter: I was wrong about GitHub Actions. I checked the raw log at https://productionresultssa4.blob.core.windows.net/actions-results/5976fa12-4dd4-4ef0-89ce-d3656415ccc0/workflow-job-run-78bf5d3d-47df-5be6-a66b-bcf82a771e6e/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-09-12T18%3A18%3A53Z&sig=pKSCvzCmaJD%2BeAxZSu7pHjnZ8%2FiZ7m59GJJ9P7MhZ1o%3D&ske=2024-09-13T03%3A38%3A36Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-09-12T15%3A38%3A36Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-05-04&sp=r&spr=https&sr=b&st=2024-09-12T18%3A08%3A48Z&sv=2024-05-04 and searched that to find
```
2024-09-04T23:12:03.2930468Z test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_parse_outputs PASSED
2024-09-04T23:12:03.3004279Z test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_empty_file PASSED
2024-09-04T23:12:03.3027440Z test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_oi2_miss_insolution PASSED
2024-09-04T23:12:03.3053933Z test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_oi2_wrong_order PASSED
```

ntalluri:
Ready for final review

agitter (approved):

PR #183, opened 2024-09-04
Enable long Snakemake workflows in HTCondor

jhiemstrawisc:
To avoid tying Snakemake execution to the current terminal (which is bad if
you ever want to log out of the AP or let your computer fall asleep), the new
script `snakemake_long.py` wraps Snakemake execution in an HTCondor local
universe job. This job is submitted like a regular HTCondor job, but it runs
on the AP in the context of the submit directory.
    
Usage:
`snakemake_long.py --snakefile (OPTIONAL) </path/to/snakefile> --profile (REQUIRED) </path/to/profile> --htcondor-jobdir (OPTIONAL) </path/to/logs>`

The only CLI option that might feel new here is `htcondor-jobdir`. This is
actually the same CLI used by the HTCondor executor, and it specifies a directory
in which logs are placed. I chose to keep names the same with this script so it
feels more familiar to Snakemake users.

### DIFF
--- a/docker-wrappers/SPRAS/snakemake_long.py
+++ b/docker-wrappers/SPRAS/snakemake_long.py
@@ -83,14 +83,12 @@ def top_main():
         cwd = os.getcwd()
         args.snakefile = pathlib.Path(cwd) / "Snakefile"
     if not os.path.exists(args.snakefile):
-        print(f"Error: The Snakefile {args.snakefile} does not exist.")
-        return 1
+        raise FileNotFoundError(f"Error: The Snakefile {args.snakefile} does not exist.")
 
     # Make sure the profile directory exists. It's harder to check if it's a valid profile at this level
     # so that will be left to Snakemake.
     if not os.path.exists(args.profile):
-        print(f"Error: The profile directory {args.profile} does not exist.")
-        return 1
+        raise FileNotFoundError(f"Error: The profile directory {args.profile} does not exist.")
 
     # Make sure we have a value for the log directory and that the directory exists.
     if args.htcondor_jobdir is None:
@@ -101,9 +99,11 @@ def top_main():
         if not os.path.exists(args.htcondor_jobdir):
             os.makedirs(args.htcondor_jobdir)
 
-
-    submit_local(args.snakefile, args.profile, args.htcondor_jobdir)
-    return 0
+    try:
+        submit_local(args.snakefile, args.profile, args.htcondor_jobdir)
+    except Exception as e:
+        print(f"Error: Could not submit local universe job. {e}")
+        raise
 
 """
 Command to activate conda environment and run Snakemake. This is run by the local universe job, not the user.
@@ -121,18 +121,29 @@ def long_main():
     snakemake -s {args.snakefile} --profile {args.profile} --htcondor-jobdir {args.htcondor_jobdir}
     """
 
-    # Run the command in a single shell session
-    result = subprocess.run(command, shell=True, executable='/bin/bash')
-
-    # Return 0 for success and 1 for failure
-    return 0 if result.returncode == 0 else 1
+    try:
+        subprocess.run(command, shell=True, executable='/bin/bash', check=True)
+        return 0
+    except subprocess.CalledProcessError as e:
+        print(f"Error: Command '{e.cmd}' returned non-zero exit status {e.returncode}.")
+        raise
+    except Exception as e:
+        print(f"Unexpected error: {e}")
+        raise
 
 def main():
-    if len(sys.argv) > 1:
-        if sys.argv[1] in ["long"]:
+    try:
+        if len(sys.argv) > 1 and sys.argv[1] in ["long"]:
             return long_main()
+        else:
+            top_main()
 
-    return top_main()
+    except subprocess.CalledProcessError as e:
+        print(f"Error: Snakemake failed with return code {e.returncode}.")
+        sys.exit(e.returncode)
+    except Exception as e:
+        print(f"Unexpected error: {e}")
+        sys.exit(1)
 
 if __name__ == '__main__':
-    sys.exit(main())
+    main()
--- a/docker-wrappers/SPRAS/snakemake_long.py
+++ b/docker-wrappers/SPRAS/snakemake_long.py
@@ -20,28 +20,33 @@ Parse various arguments for the script. Note that this script has two "modes" of
 need different arguments. The "top" mode is for submitting the HTCondor wrapper, and the "long" mode
 is for running the Snakemake command itself.
 """
-def parseArgs(isLocal=False):
+def parse_args(isLocal=False):
     parser = argparse.ArgumentParser(description="A tool for long-running Snakemake jobs with HTCondor.")
-    # We add a special command to trigger allowing this script to execute the long-running Snakemake command.
     if isLocal:
+        # We add a special command that allows this singular executable to serve two purposes. The executable
+        # is first run by the user with their args to submit the local universe job. Then, the local universe
+        # job runs `snakemake_long.py long <user args>` to indicate to the script that it's time to submit the
+        # long-running Snakemake process instead of submitting another local universe job.
         parser.add_argument("command", help="Helper command to run", choices=["long"])
-    parser.add_argument("--snakefile", help="The Snakefile to run.", required=False)
+    parser.add_argument("--snakefile", help="The Snakefile to run. If omitted, the Snakefile is assumed to be in the current directory.", required=False)
     parser.add_argument("--profile", help="A path to a directory containing the desired Snakemake profile.", required=True)
     # I'd love to change this to "logdir", but using the same name as Snakemake for consistency of feeling between this script
     # and Snakemake proper.
-    parser.add_argument("--htcondor-jobdir", help="The directory Snakemake will write logs to.", required=False)
+    parser.add_argument("--htcondor-jobdir", help="The directory Snakemake will write logs to. If omitted, a 'logs` directory will be created in the current directory", required=False)
     return parser.parse_args()
 
 """
 Given a Snakefile, profile, and HTCondor job directory, submit a local universe job that runs
 Snakemake from the context of the submission directory.
 """
-def submitLocal(snakefile, profile, htcondor_jobdir):
+def submit_local(snakefile, profile, htcondor_jobdir):
     # Get the location of this script, which also serves as the executable for the condor job.
     script_location = pathlib.Path(__file__).resolve()
 
     submit_description = htcondor.Submit({
         "executable":              script_location,
+        # We use the "long" command to indicate to the script that it should run the Snakemake command instead of submitting another job.
+        # See comment in parse_args for more information.
         "arguments":               f"long --snakefile {snakefile} --profile {profile} --htcondor-jobdir {htcondor_jobdir}",
         "universe":                "local",
         "request_disk":            "512MB",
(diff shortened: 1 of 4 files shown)


jhiemstrawisc:
@agitter, this branch is based on the other CHTC executor work in #172. That should be merged first (and doing so will dramatically reduce the noise in this PR by hiding most of the commits that came along with it from the other branch).

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+is for running the Snakemake command itself.
+"""
+def parseArgs(isLocal=False):
+    parser = argparse.ArgumentParser(description="A tool for long-running Snakemake jobs with HTCondor.")
+    # We add a special command to trigger allowing this script to execute the long-running Snakemake command.
+    if isLocal:
+        parser.add_argument("command", help="Helper command to run", choices=["long"])
+    parser.add_argument("--snakefile", help="The Snakefile to run.", required=False)
+    parser.add_argument("--profile", help="A path to a directory containing the desired Snakemake profile.", required=True)
+    # I'd love to change this to "logdir", but using the same name as Snakemake for consistency of feeling between this script
```
agitter, line 30: That seems like the right call even though "logdir" does sound better

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+Parse various arguments for the script. Note that this script has two "modes" of operation which
+need different arguments. The "top" mode is for submitting the HTCondor wrapper, and the "long" mode
+is for running the Snakemake command itself.
+"""
+def parseArgs(isLocal=False):
+    parser = argparse.ArgumentParser(description="A tool for long-running Snakemake jobs with HTCondor.")
+    # We add a special command to trigger allowing this script to execute the long-running Snakemake command.
+    if isLocal:
+        parser.add_argument("command", help="Helper command to run", choices=["long"])
+    parser.add_argument("--snakefile", help="The Snakefile to run.", required=False)
```
agitter, line 28: Should we say anything about default behavior when this is not provided?

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+import time
+
+import htcondor
+
+"""
+Parse various arguments for the script. Note that this script has two "modes" of operation which
+need different arguments. The "top" mode is for submitting the HTCondor wrapper, and the "long" mode
+is for running the Snakemake command itself.
+"""
+def parseArgs(isLocal=False):
```
agitter, line 23: The camel case function names in this file are different from the naming conventions used in the rest of the project.

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+Given a Snakefile, profile, and HTCondor job directory, submit a local universe job that runs
+Snakemake from the context of the submission directory.
+"""
+def submitLocal(snakefile, profile, htcondor_jobdir):
+    # Get the location of this script, which also serves as the executable for the condor job.
+    script_location = pathlib.Path(__file__).resolve()
+
+    submit_description = htcondor.Submit({
+        "executable":              script_location,
+        "arguments":               f"long --snakefile {snakefile} --profile {profile} --htcondor-jobdir {htcondor_jobdir}",
```
agitter, line 45: Does `long` here imply there is a max runtime on the local universe job?

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+        "universe":                "local",
+        "request_disk":            "512MB",
+        "request_cpus":            1,
+        "request_memory":          512,
+
+        # Set up logging
+        "log":                     f"{htcondor_jobdir}/snakemake-long.log",
+        "output":                  f"{htcondor_jobdir}/snakemake-long.out",
+        "error":                   f"{htcondor_jobdir}/snakemake-long.err",
```
agitter, line 54: The `long` in these filenames may not be intuitive

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Specify `getenv` so that our script uses the appropriate environment
+        # when it runs in local universe. This allows the job to access
+        # modules we've installed in the submission environment (notably spras).
+        "getenv":                  "true",
+
+        "JobBatchName":            f"spras-long-{time.strftime('%Y%m%d-%H%M%S')}",
```
agitter, line 61: Same comment here, calling this `long` may not be intuitive. Should it be `local`? `manage` to correspond with the print statement below?

docker-wrappers/SPRAS/snakemake_long.py
```diff
@@ -0,0 +1,134 @@
...
+def topMain():
+    args = parseArgs()
+
+    # Check if the snakefile is provided. If not, assume it's in the current directory.
+    if args.snakefile is None:
+        cwd = os.getcwd()
+        args.snakefile = pathlib.Path(cwd) / "Snakefile"
+    if not os.path.exists(args.snakefile):
+        print(f"Error: The Snakefile {args.snakefile} does not exist.")
+        return 1
```
agitter, line 82: This approach of returning error codes is a bit atypical to me. Is this better than raising errors?

agitter (changes requested):
My initial testing started off well, but then all 6 of my jobs were put on hold `condor_q -l` showed
```
HoldReason = "Transfer input files failure at access point ap2001 while sending files to execution point slot1_4@e2354.chtc.wisc.edu. Details: reading from file /var/lib/condor/execute/slot1/dir_20697/_condor_stdout: (errno 2) No such file or directory"
HoldReasonCode = 13
HoldReasonSubCode = 2
```

docker-wrappers/SPRAS/snakemake_long.py
jhiemstrawisc: Nope, it's my way of using the same executable to serve two purposes -- `snakemake_long.py` is first run by the user with their args to submit the local universe job. Then, the local universe job runs `snakemake_long.py long <user args>` to indicate to itself that it's time to submit the long-running Snakemake process instead of submitting another local universe job.

I'll add a comment for now to better explain the behavior, and if you don't like the name I can change it all together.

docker-wrappers/SPRAS/snakemake_long.py
jhiemstrawisc: I'll rename them "snakemake.<suffix>` unless you have another preference then. The `.log` file will contain the HTCondor job log for the local universe wrapper, and the `.out/.err` files do actually contain output from the Snakemake process running in the wrapper.

docker-wrappers/SPRAS/snakemake_long.py
jhiemstrawisc: Since this is all spras-specific (at this point), how about just calling it `spras-<timestamp>`?

docker-wrappers/SPRAS/snakemake_long.py
jhiemstrawisc: This is mostly to play better with HTCondor, which expects job executables to return 0 for success, and something else for failure. I guess raising an error _also_ returns non-zero, but this felt more explicit to me. I'll let you make the call.

jhiemstrawisc:
> My initial testing started off well, but then all 6 of my jobs were put on hold `condor_q -l` showed
> 
> ```
> HoldReason = "Transfer input files failure at access point ap2001 while sending files to execution point slot1_4@e2354.chtc.wisc.edu. Details: reading from file /var/lib/condor/execute/slot1/dir_20697/_condor_stdout: (errno 2) No such file or directory"
> HoldReasonCode = 13
> HoldReasonSubCode = 2
> ```

Hmm, I'm not totally sure what to make of that error at first glance. It seems to imply that there was a failure sending files from the AP to the EP, but the file it's looking to send seems to come from the EP...

Can you share the content of the generated log files `snakemake-long.out/err`?

agitter:
I tried running this again and encountered the same error:
```
HoldReason = "Transfer input files failure at access point ap2001 while sending files to execution point slot2_2@txie-dsigpu4000.chtc.wisc.edu. Details: reading from file /var/lib/condor/execute/slot2/dir_530230/_condor_stdout: (errno 2) No such file or directory"
```

497726.0 is an example of a held job. I left it in the queue for now.

agitter:
We discovered some of my errors were because my .sif file was named incorrectly. I returned to that submission later, and condor_q shows no more running jobs. However, the output dir does not show all expected files.

In `snakemake-long.err` I see:
```
Error in rule reconstruct:
    message: Job 18 with HTCondor Cluster ID 498174 has  status Completed, but failed with ExitCode 1.For further error details see the cluster/cloud log and the log files of the involved rule(s).
    jobid: 18
    input: output/prepared/data0-meo-inputs/sources.txt, output/prepared/data0-meo-inputs/targets.txt, output/prepared/data0-meo-inputs/edges.txt
    output: output/data0-meo-params-GKEDDFZ/raw-pathway.txt
    external_jobid: 498174

[Mon Sep 23 14:58:24 2024]
Error in rule reconstruct:
    message: Job 3 with HTCondor Cluster ID 498179 has  status Completed, but failed with ExitCode 1.For further error details see the cluster/cloud log and the log files of the involved rule(s).
    jobid: 3
    input: output/prepared/data0-omicsintegrator1-inputs/prizes.txt, output/prepared/data0-omicsintegrator1-inputs/edges.txt
    output: output/data0-omicsintegrator1-params-PU62FNV/raw-pathway.txt
    external_jobid: 498179

Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: .snakemake/log/2024-09-23T141203.356730.snakemake.log
WorkflowError:
At least one job did not complete successfully.
``

**Comment by jhiemstrawisc** [comment]
Adding my thoughts from our Slack conversation for git posterity -- I believe the errors in your run were related to the luck of the draw in a distributed, heterogeneous computing environment. When I investigated, everything succeeded with a retry that allowed Snakemake to pick up where it left off. I was even able to generate a few new errors related to a pokey HTCondor SchedD, but again a retry picked things back up and carried the workflow to completion.

It's less clear to me how to communicate this to users in the docs -- some errors are definitely not retryable, but enumerating all the cases for "rerun if you encounter error XXX but don't rerun if you encounter error YYY" isn't feasible due to the unknowably-many error modes. Maybe we add a snippet to the README here saying something like "Some errors Snakemake might encounter while executing rules in the workflow boil down to bad luck in a distributed, heterogeneous computational environment, and it's expected that _some_ errors can be solved simply by rerunning. If you encounter a Snakemake error, try restarting the workflow to see if the same error is generated in the same rule a second time -- repeatable, identical failures are more likely to indicate a more fundamental issue that might require user intervention to fix."

**Comment by agitter** [comment]
I tested again, and this time it ran to completing confirming what you said above. Adding a message similar to
> Some errors Snakemake might encounter while executing rules in the workflow boil down to bad luck in a distributed, heterogeneous computational environment, and it's expected that some errors can be solved simply by rerunning. If you encounter a Snakemake error, try restarting the workflow to see if the same error is generated in the same rule a second time -- repeatable, identical failures are more likely to indicate a more fundamental issue that might require user intervention to fix.

does seem valuable because I initially thought something was more fundamentally broken.

I saw in the snakemake long err log
```
The job argument contains a single quote. Removing it to avoid issues with HTCondor.
```
Does that mean something we should pay attention to?

agitter (approved):

Issue #181, opened 2024-08-29
Check robustness of PCA representation


agitter:
Based on a discussion with @brycejoh16 we realized we should confirm that our PCA representation of a collection of graphs is robust to the order in which the graphs and the edges are presented. We can try randomly permuting the rows and columns of the graph X edge matrix before running PCA and confirm the outputs still pass our test cases.

ntalluri:
I checked the robustness of PCA. Shuffling the columns (the algorithms) and the rows (the edges) do not change the output of PCA.

ntalluri:
https://github.com/Reed-CompBio/spras/pull/185

added pca robustness test to test_ml

agitter:
Closed by #185

PR #185, opened 2024-09-10
Added test to check robustness of PCA representation

ntalluri:


agitter (approved):
Great test. Closes #181.

Issue #189, opened 2024-10-10
MEO Node ID Error


ntalluri:
If a node is in the format [name]_[info], MEO ignores the [info] part and only uses the [name] during reconstructions.

This happened when doing parameter tuning egfr dataset with MEO, the node names were missing the “_HUMAN” suffix. In the gold standard dataset, nodes such as EGF_HUMAN and EDGR_HUMAN are expected, but in the MEO output, they appear without the suffix. This error led to zero matching nodes to the gold standard nodes during evaluation. 

input prize file
```
NODEID	prize	sources	targets	active
1433Z_HUMAN	1.041379133		True	True
41_HUMAN	3.389112802		True	True
4ET_HUMAN	2.569973509		True	True
```

raw pathway:
```
Source	Type	Target	Oriented	Weight
ERBB2	pd	CTNB1	true	0.553333
ERBB2	pd	CDK1	true	0.773333
EGFR	pd	SHC1	true	0.666667
EGFR	pd	41	true	0.553333
```

agitter:
MEO may not support node IDs with underscores based on Neha's EGFR dataset testing and my initial inspection of its Vertex class: https://github.com/agitter/meo/blob/master/src/alg/Vertex.java#L45-L50

ntalluri:
Doing something similar that domino does would help with fixing this issue 

Ex.replacing the "_" with a "-" and then in post processing, replacing the  "-" with "_"

agitter:
The relevant DOMINO code is:
https://github.com/Reed-CompBio/spras/blob/a26f4d0ea86e4bf8e3fb89fcca5c33e44103d7c5/spras/domino.py#L215-L231

I don't believe replacing "\_" with "-" will work because then "A_B" and "A-B" would be the same. Those IDs are unlikely to both be in the network, but they could be. We'll need to replace "_" with another string that is very unlikely to appear in real protein names. If MEO supports Unicode, using some combination of strange Unicode characters would be safe.

PR #190, opened 2024-10-25
Update MEO Node ID Error

ntalluri:
https://github.com/Reed-CompBio/spras/issues/189 fixing this issue


ntalluri:
The way I made generate-inputs test suite would require me to change the format of data1 and data0 to test MEO's update.

spras/meo.py
```diff
@@ -10,6 +10,8 @@
 
 __all__ = ['MEO', 'write_properties']
 
+# replaces all underscores in the node names with unicode seperator
```
agitter, line 13:
```suggestion
# replaces all underscores in the node names with unicode seperator
# MEO keeps only the substring up to the first underscore when parsing node names
# https://github.com/agitter/meo/blob/1fe57e8ff3952c494e2b14dfdc563a84596e2fcd/src/alg/Vertex.java#L56-L71
```
Let's document why we need this

agitter (commented):
The solution looks good to me.

> The way I made generate-inputs test suite would require me to change the format of data1 and data0 to test MEO's update.

Does this still need to be done?

ntalluri:
> The solution looks good to me.
> 
> > The way I made generate-inputs test suite would require me to change the format of data1 and data0 to test MEO's update.
> 
> Does this still need to be done?

I just added an idea, but the only thing I wasn't able to change is the parse_outputs domino case

ntalluri:
localized the testing to be done only in generate inputs instead of it being coupled to the changes in inputs and the config.yaml file used as the examples.

Issue #188, opened 2024-10-07
`InvalidIndexError` when running `ml_analysis`


ntalluri:
I encountered a `InvalidIndexError` running the `ml_analysis` code running individual runs  of parameter sweeps on mincostflow on the EGFR dataset. The issue happens in the `summarize_networks` function, related to reindexing, where non-unique index values are causing problems with pandas concatenation. 

 ```
# initially construct separate dataframes per algorithm
    edge_dataframes = []
    # the dataframe is set up per algorithm and a 1 is set for the edge pair that exists in the algorithm
    for tup in edge_tuples:
        dataframe = pd.DataFrame(
            {
                str(tup[0]): 1,
            }, index=tup[1]
        )
        edge_dataframes.append(dataframe)

    # concatenating all the algorithm-specific dataframes together
    # (0 is set for all the edge pairs that don't exist per algorithm)
    concated_df = pd.concat(edge_dataframes, axis=1, join='outer')
    concated_df = concated_df.fillna(0)
    concated_df = concated_df.astype('int64')
```

Error Trace:

```
RuleException:
InvalidIndexError in file /Users/nehatalluri/Desktop/research/spras/Snakefile, line 315:
Reindexing only valid with uniquely valued Index objects
  File "/Users/nehatalluri/Desktop/research/spras/Snakefile", line 315, in __rule_ml_analysis
  File "/Users/nehatalluri/Desktop/research/spras/spras/analysis/ml.py", line 85, in summarize_networks
  File "/Users/nehatalluri/anaconda3/envs/spras/lib/python3.11/site-packages/pandas/util/_decorators.py", line 331, in wrapper
  File "/Users/nehatalluri/anaconda3/envs/spras/lib/python3.11/site-packages/pandas/core/reshape/concat.py", line 381, in concat
  File "/Users/nehatalluri/anaconda3/envs/spras/lib/python3.11/site-packages/pandas/core/reshape/concat.py", line 612, in get_result
  File "/Users/nehatalluri/anaconda3/envs/spras/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3904, in get_indexer
```

ntalluri:
The issue is because the pathways generated by mincostflow contain duplicate edges in the pathway.txt file. This seems to occur in many of the output pathways from mincostflow.

Here is an example of one the pathway.txt files with duplicate edges

Node1	Node2	Rank	Direction
_**EGFR_HUMAN	EGF_HUMAN	1	U**_
[...]

PR #191, opened 2024-10-25
summarize_networks InvalidIndexError

ntalluri:
PR for https://github.com/Reed-CompBio/spras/issues/188

Currently created a test case to showcase the error and how it is happening


test/ml/test_ml.py
```diff
@@ -42,6 +42,9 @@ def test_summarize_networks_wrong_direction(self):
         with pytest.raises(ValueError):
             ml.summarize_networks([INPUT_DIR + 'test-data-wrong-direction/wrong-direction.txt'])
 
+    def test_repeat_edges(self):
```
ntalluri, line 45: Look at `run tests` in checks. The error is the expected behavior I was expecting to see.

spras/analysis/ml.py
```diff
@@ -78,6 +78,7 @@ def summarize_networks(file_paths: Iterable[Union[str, PathLike]]) -> pd.DataFra
                 str(tup[0]): 1,
             }, index=tup[1]
         )
+        dataframe = dataframe[~dataframe.index.duplicated(keep='first')]
```
ntalluri, line 81: https://pandas.pydata.org/docs/reference/api/pandas.Index.duplicated.html

ntalluri, line 81: add a warning that edges are being dropped from a file

agitter:
```suggestion
        # Mark the first occurrence of a duplicate as True and all others as False
        # Non-duplicates are also marked as True
        dataframe = dataframe[~dataframe.index.duplicated(keep='first')]
```

agitter: If this is hard to do here, we can think about whether it should be handled somewhere else. For instance, should the `parse_output` functions check for duplicate edges and raise a warning or error?

agitter (commented):
What you have looks good.

If you didn't already open an issue about MinCostFlow giving the duplicate issues that trigged this problem, please do that and link it to https://github.com/Reed-CompBio/spras/issues/188#issuecomment-2438619614. We may want to follow up on that potential bug later.

test/parse-outputs/input/duplicate-edges/omicsintegrator1-raw-pathway.txt
```diff
@@ -0,0 +1,4 @@
+A	pd	C
+C	pp	D
+A	pd	C
+C	pp	D
```
agitter, line 4: This test case makes it seem like the util function may be too simplistic. What if the edges were instead:
```
A	pd	C
C	pp	D
A	pp	C
D	pp	C
```
What would the current behavior be? What would the desired behavior be?

spras/util.py
```diff
@@ -76,3 +76,14 @@ def raw_pathway_df(raw_pathway_file: str, sep: str = '\t', header: int = None) -
         df = pd.DataFrame(columns=['Node1', 'Node2', 'Rank', 'Direction'])
 
     return df
+
+def duplicate_edges(df: pd.DataFrame):
```
agitter, line 81: I like this new design of checking for duplicate edges in parse outputs. However, see the comment on the Omics Integrator test case about whether we need to have more complex duplicate checking logic.

agitter (commented):
I pushed a note to the contributing guide about this new post-processing step.

test/parse-outputs/input/duplicate-edges/omicsintegrator1-raw-pathway.txt
ntalluri: With that, the code in util.py doesn't take that my thoughts into consideration. It will remove anything in the form A C without taking into account the directionality.

spras/util.py
ntalluri: I updated it to remove duplicate rows now; treats undirected and directed edges coming between 2 nodes as separate edges that the original algorithm should have dealt with.

ntalluri:
> I pushed a note to the contributing guide about this new post-processing step.
> 
> > If you didn't already open an issue about MinCostFlow giving the duplicate issues that trigged this problem, please do that and link it to #188 (comment). We may want to follow up on that potential bug later.
> 
> Do we still want a new issue to track this?

I think we should track that there is still a problem with MCF, but we solve the problem for now with this PR.

ntalluri:
keep the smallest rank instead of keep first; sort by rank first
need to sort the undirected edges ( B - A changed to A - B)  before removing duplicates

ntalluri:
@agitter Ready for another review

test/parse-outputs/input/duplicate-edges/omicsintegrator1-raw-pathway.txt
```diff
@@ -2,3 +2,6 @@ A	pd	C
 C	pp	D
 A	pd	C
 C	pp	D
+A	pp	C
```
agitter, line 5: This is a good test case for all of the corner cases we want to cover.

agitter (approved):

PR #193, opened 2024-11-04
Parameter Tuning Code Integration: update evaluation and add egfr gold standard

ntalluri:


config/egfr.yaml
ntalluri, line 1: should I add the gold standard to this?

test/evaluate/test_evaluate.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        Path(OUT_DIR).mkdir(parents=True, exist_ok=True)
+
+    def test_node_ensemble(self):
+        ensemble_file = INPUT_DIR + 'ensemble-network.tsv'
+        edge_freq = Evaluation.edge_frequency_node_ensemble(ensemble_file)
+        edge_freq.to_csv(OUT_DIR + 'node-ensemble.csv', sep="\t", index=False)
+        assert filecmp.cmp(OUT_DIR + 'node-ensemble.csv', EXPECT_DIR + 'expected-node-ensemble.csv', shallow=False)
+
+    def test_precision_recal_curve_ensemble_nodes(self):
```
ntalluri, line 28: I don't know how else to test the ensemble node outputs other than looking at the image

config/config.yaml
```diff
@@ -173,4 +173,14 @@ analysis:
         # 'euclidean', 'manhattan', 'cosine'
         metric: 'euclidean'
       evaluation:
+        # evaluation per dataset-goldstandard pair
+        # evalution will not run unless ml include is set to true
         include: true
+        # adds evaluation per algorithm per dataset-goldstandard pair
+        # evalution per algortihm will not run unless ml include and ml aggregate_per_algorithm is set to true
+        aggregate_per_algorithm: true
+        # TODO: should we decouple parts of eval that involve ml
```
ntalluri, line 182: Lots of coupling happening now. I put in a solution for now in config.py, but is it worth separating the functions into their own true/ false?

spras/analysis/ml.py
```diff
@@ -142,8 +142,14 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_var: str, output_coord:
     if not isinstance(labels, bool):
         raise ValueError(f"labels={labels} must be True or False")
 
-    scaler = StandardScaler()
+    #TODO: MinMaxScaler changes nothing about the data
```
ntalluri, line 145: I don't know if it is better to use StandardScalar or MinMaxScalar for the binary data

spras/evaluation.py
```diff
@@ -72,29 +79,161 @@ def load_files_from_dict(self, gold_standard_dict: Dict):
...
+        @param algorithms: list of algorithms used in current run of SPRAS
+        @param output_file: the filename to save the precision and recall of each pathway
+        @param output_png (optional): the filename to plot the precision and recall of each pathway (not a PRC)
         """
         y_true = set(node_table['NODEID'])
         results = []
-
         for file in file_paths:
             df = pd.read_table(file, sep="\t", header=0, usecols=["Node1", "Node2"])
+            # TODO: do we want to include the pathways that are empty for evaluation / in the pr_df?
```
ntalluri, line 97: Currently the code will add a precision and recall for empty pathways. Is that something we shouldn't include?

Snakefile
```diff
@@ -105,8 +104,17 @@ def make_final_input(wildcards):
...
-        final_input.extend(expand('{out_dir}{sep}{dataset_gold_standard_pair}-evaluation.txt',out_dir=out_dir,sep=SEP,dataset_gold_standard_pair=dataset_gold_standard_pairs,algorithm_params=algorithms_with_params))
-    
+        final_input.extend(expand('{out_dir}{sep}{dataset_gold_standard_pair}-eval{sep}precision-recall-per-pathway.txt',out_dir=out_dir,sep=SEP,dataset_gold_standard_pair=dataset_gold_standard_pairs,algorithm_params=algorithms_with_params))
+        final_input.extend(expand('{out_dir}{sep}{dataset_gold_standard_pair}-eval{sep}precision-recall-per-pathway.png',out_dir=out_dir,sep=SEP,dataset_gold_standard_pair=dataset_gold_standard_pairs)) 
+        final_input.extend(expand('{out_dir}{sep}{dataset_gold_standard_pair}-eval{sep}precision-recall-pca-chosen-pathway.txt',out_dir=out_dir,sep=SEP,dataset_gold_standard_pair=dataset_gold_standard_pairs))
+        final_input.extend(expand('{out_dir}{sep}{dataset_gold_standard_pair}-eval{sep}precision-recall-curve-ensemble-nodes.png',out_dir=out_dir,sep=SEP,dataset_gold_standard_pair=dataset_gold_standard_pairs,algorithm_params=algorithms_with_params))
+        # TODO: should we provide the node ensemble frequencies
```
ntalluri, line 111: Since we are already calculating the node ensembles, should we give it to the user?

config/config.yaml
ntalluri: Maybe deal with some of the coupling by giving warnings and stopping the flow rather than silently shutting things off

ntalluri:
Note for self: Try seeing if it makes more sense to separate the parameter tuning and evaluation code into their classes for organization

ntalluri:
I think I need to redo the idea for using ensembling for parameter tuning. 

Currently the code will take in the ensemble file and build a node ensemble file by processing an ensemble of edge frequencies to identify the highest frequency associated with each node (y_scores). This is then compared to the node gold standard (y_trues). Then these are plotted (with no point labels) to a graph showing the PRC between the nodes in the output pathways and the gold standard, as well as the average PR between all the nodes. 

I don’t think this can be used to do ensemble parameter tuning the way I think I would do it. It could be used to help parameter tune though to get a better grid search or for evaluation in general.

- also the visualization for this could be way better (it lacks label information making it hard to fully understand)

input/gs-egfr.txt
agitter, line 1: Can you please add some notes about how you prepared this file and link it back to the related issue? We'll want to be able to track that later.

agitter (commented):
> I think I need to redo the idea for using ensembling for parameter tuning.

Is there a way to break this huge pull request into smaller parts? Currently it combines evaluation, parameter tuning, ensembling, etc. That is making it hard to coordinate all of those decisions and also review the big pull request. I'm wondering if we can start small and merge in a subset of the file changes to lock in some progress, like the evaluation code.

If it is all interdependent, we'll have to deal with that and proceed as is.

agitter:
> Is there a way to break this huge pull request into smaller parts?

In response to this, we now have separate follow up pull requests #207, #208, #209.

ntalluri:
Start with this PR first.

Included in this PR:
1) updated code to decouple the ML and evaluation code from each other in the config file
2) new test cases for the config file to test different coupling situations
3) the new egfr gold standard dataset
4) a new egfr parameter tuned config.yaml file

config/egfr-param-tuning.yaml
agitter, line 1: What is the overall goal of this huge config file? Why does it use a long list of individual runs instead of a list or other syntax that enumerates parameter combinations? Some context would be helpful.

If this is being added to support the SPRAS benchmarking, does it belong in the spras-benchmarking repo and not here?

input/gs-egfr.txt
```diff
@@ -0,0 +1,324 @@
...
+ATF2_HUMAN
+ATX1_HUMAN
+B2CL1_HUMAN
+BAD_HUMAN
+BCAR1_HUMAN
+BCL2_HUMAN
+BDNF_HUMAN
+BRAF_HUMAN
+BTC_HUMAN
+Ca++_PSEUDONODE
```
agitter, line 38: We should remove all nodes with the `_PSEUDONODE` suffix

input/gs-egfr.txt
agitter, line 1: A reminder to document how this file was prepared.

spras/config.py
```diff
@@ -101,6 +101,10 @@ def __init__(self, raw_config):
         self.analysis_include_ml = None
         # A Boolean specifying whether to run the Evaluation analysis
         self.analysis_include_evaluation = None
+        # A Boolean specifying whether to run the ML per algorithm analysis
+        self.analysis_include_ml_aggregate_algo = None
+        # A Boolean specifying whether to run the Evaluation per algorithm aanalysis
```
agitter, line 106:
```suggestion
        # A Boolean specifying whether to run the evaluation per algorithm analysis
```

spras/config.py
```diff
@@ -252,11 +257,27 @@ def process_config(self, raw_config):
...
+
+        # Raises an error if Evaluation is enabled but no gold standard data is provided
         if self.gold_standards == {} and self.analysis_include_evaluation:
             raise ValueError("Evaluation analysis cannot run as gold standard data not provided. "
                              "Please set evaluation include to false or provide gold standard data.")
 
-        if 'aggregate_per_algorithm' in self.ml_params and self.analysis_include_ml:
-            self.analysis_include_ml_aggregate_algo = raw_config["analysis"]["ml"]["aggregate_per_algorithm"]
+         # Only run Evaluation if ML is set to True
+        if not self.analysis_include_ml and self.analysis_include_evaluation:
```
agitter, line 272: Why do we need to check the value of `self.analysis_include_evaluation` here?

spras/config.py
```diff
@@ -252,11 +257,27 @@ def process_config(self, raw_config):
...
+
+        # Only run Evaluation aggregate per algorithm if analysis include ML is set to True
+        if 'aggregate_per_algorithm' in self.evaluation_params and self.analysis_include_evaluation:
+            self.analysis_include_evaluation_aggregate_algo = raw_config["analysis"]["evaluation"]["aggregate_per_algorithm"]
         else:
-            self.analysis_include_ml_aggregate_algo = False
+            self.analysis_include_evaluation_aggregate_algo = False
+
+        # Only run Evaluation per algorithm if ML per algorithm is set to True
+        if not self.analysis_include_ml_aggregate_algo and self.analysis_include_evaluation_aggregate_algo:
```
agitter, line 282: Same question for `self.analysis_include_evaluation_aggregate_algo` here

test/test_config.py
```diff
@@ -142,3 +144,132 @@ def test_error_gs_dataset_mismatch(self):
 
         with pytest.raises(ValueError):
             config.init_global(test_config)
+
+    def test_eval_ml_coupling(self):
```
agitter, line 148: You can use `@pytest.mark.parametrize` to remove the redundant code. You will create test function arugments and then provide a list of values the test case should take. See https://github.com/Reed-CompBio/spras/blob/master/test/test_util.py#L36-L42 for an example.

test/test_config.py
```diff
@@ -142,3 +144,132 @@ def test_error_gs_dataset_mismatch(self):
...
+        test_config = get_test_config()
+        include_combos = [(True, True), (True, False), (False, True), (False, False)]
+
+        # ml: True evaluation: True
+        test_config["analysis"]["ml"]["include"] = include_combos[0][0]
+        test_config["analysis"]["evaluation"]["include"] = include_combos[0][1]
+        config.init_global(test_config)
+        assert config.config.analysis_include_ml == True and config.config.analysis_include_evaluation == True
```
agitter, line 156: These are boolean so they should simplify to assert or assert not (no ==).

test/test_config.py
```diff
@@ -142,3 +144,132 @@ def test_error_gs_dataset_mismatch(self):
...
+        test_config["analysis"]["ml"]["include"] = include_combos[3][0]
+        test_config["analysis"]["evaluation"]["include"] = include_combos[3][1]
+        config.init_global(test_config)
+        assert config.config.analysis_include_ml == False and config.config.analysis_include_evaluation == False
+
+
+    def test_ml_agg_algo_coupling(self):
+
+        test_config = get_test_config()
+        include_combos = [(True, True), (True, False), (False, True), (False, False)]
```
agitter, line 180: Can parameterize these as well

test/test_config.py
```diff
@@ -142,3 +144,132 @@ def test_error_gs_dataset_mismatch(self):
...
+        config.init_global(test_config)
+        assert config.config.analysis_include_ml == False and config.config.analysis_include_ml_aggregate_algo == False
+
+    def test_eval_agg_algo_coupling(self):
+
+        test_config = get_test_config()
+        test_config["analysis"]["ml"]["include"] = True
+        test_config["analysis"]["ml"]["aggregate_per_algorithm"] = True
+
+        include_combos = [(True, True), (True, False), (False, True), (False, False)]
```
agitter, line 211: Can parameterize these as well

agitter (commented):
There is a lot of complexity involved just in this config file parsing. It is good you adding so many test cases.

config/egfr-param-tuning.yaml
ntalluri: I think it is huge because I used [yaml prettify](https://onlineyamltools.com/prettify-yaml). 

This was for benchmarking the egfr dataset. We can keep it in the spras-benchmarking repo, but a while back you mentioned how it would be an example we would want to keep in the SPRAS repo.

spras/config.py
ntalluri: I was being very explicit about the use case when I made this. We don't actually need self.analysis_include_evaluation_aggregate_algo.

spras/config.py
ntalluri: I was being very explicit about the use case when I made this. We don't actually need self.analysis_include_evaluation_aggregate_algo.

input/README.md
```diff
@@ -39,6 +39,22 @@ A	B	0.98
...
+
+For example:
+```
+A
+B
+C
+```
+
+#### Edges
+TBD
```
agitter, line 56: @sumedhars once you have the EGFR pathway edge gold standard ready, we can update this documentation in a follow up pull request.

agitter (approved):

config/egfr-param-tuning.yaml
ntalluri: I need to go through and clean this up. I think when I was tuning before, I had code that made separate runs based on my chosen heuristics; this isn't really needed and everything can be in a list.

ntalluri: > I don't mind the YAML syntax and whether we use a more compact or expanded format. I was referring to how many separate runs there are, >300 I believe. When I previously suggested that we put the parameter tuning config in this repo, I may have anticipated we would have far fewer parameter combinations. This now seems like a heavy duty analysis instead of an example of how to use the SPRAS software. My perspective is that config files demonstrating how to use (or test) SPRAS can say here, and those used to do analysis with SPRAS can go in the benchmarking repo. Does that seem like a good division?

Yes this is a good division.

PR #194, opened 2024-11-11
Update summary.py to include parameter combinations

cgsze:


spras/analysis/summary.py
```diff
@@ -44,28 +47,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame) ->
...
         for node_list in nodes_by_col:
             num_nodes = len(set(nw).intersection(node_list))
             cur_nw_info.append(num_nodes)
+
+        # String split name to access algorithm and hashcode from filepath
+        # Name of filepath follows format "output/data#-algo-params-hashcode/pathway.txt"
+        # algorithm parameters have format { algo : { hashcode : { parameter combos } } }
+
+        filepath = nw_name.split("/")
+        filename = filepath[1].split("-")
```
ntalluri, line 64: Will this work for outputs stored in subfolders within the main output folder?

example `output/example/data#-algo-params-hashcode/pathway.txt"`

agitter: Can get the list of algorithm-params-hashcode from `algorithms_with_params` in the Snakefile and then pass it to the function so you don't have to extract it

agitter:
For the test cases, we discussed dropping the toy summarization example because it does not have corresponding algorithm-parameter combinations that generated those graphs. We will summarize only the example data and EGFR data that correspond to actual SPRAS runs.

spras/analysis/summary.py
```diff
@@ -44,28 +47,43 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame) ->
...
+        hashcode = filename[2]
+        index = index + 1
+
+        param_combo = algo_params[algo][hashcode]
+        params = json.dumps(param_combo)
+        params = params.replace("\"", "") #removes extra double quotes from string
+        cur_nw_info.append(params)
+
+        # Prepare column names
+        col_names = ["Name", "Number of nodes", "Number of undirected edges", "Number of connected components"]
```
ntalluri, line 73: Should the header be updated to “Number of edges" instead of "Number of undirected edges"?

Even though we’re treating all graphs as undirected (on line 43), directionality doesn’t seem to be considered in any of the statistics being calculated in summary.py at the moment.

agitter: I believe the original intention of this header was to convey that even a directed or mixed graph would be parsed and summarized as an undirected graph here. I can see how that could instead be interpreted as only counting the undirected edges in a mixed graph though. Which do you think is more precise?

ntalluri: I was looking at the summary file and got confused on why were only counting the undirected edges. So I think renaming it to "Number of edges" is more intuitive and putting a comment about line 43 to say that "directed or mixed graph are parsed and summarized as an undirected graph"

test/analysis/test_cytoscape.py
```diff
@@ -6,17 +6,23 @@
 import spras.config as config
 from spras.analysis.cytoscape import run_cytoscape
 
-config.init_from_file("config/config.yaml")
+config.init_from_file("input/config.yaml")
```
ntalluri, line 9: if using the config file in the input folder in analysis, make sure to set cytoscape include to true incase

agitter:
While doing a final code review, I made some implementation changes. I now quote strings in the parameter combinations in the summary table. However, reading those tables meant that checking the dataframes for equality failed for some reason. Instead, I compare the two versions of the summary table on disk, which works.

This code still assumes that the order of the file paths and the algorithms list is the same, which is a bit dangerous. We could parse the file paths to get the algorithm parameter hash codes instead. I won't do that now. It would work for SPRAS workflows but complicate testing.

agitter:
This adds a second copy of the PPI network `test/analysis/input/phosphosite-irefindex13.0-uniprot.txt` to the repo, which may not be a great idea. We can reconsider that later.

tristan-f-r:
> This adds a second copy of the PPI network `test/analysis/input/phosphosite-irefindex13.0-uniprot.txt` to the repo, which may not be a great idea. We can reconsider that later.

This was dropped in https://github.com/Reed-CompBio/spras/pull/320.

PR #198, opened 2024-12-09
Container workflow overhaul

jhiemstrawisc:
This PR overhauls some of the GHA workflows designed to build spras's docker containers. In particular, it defines a new template in `.github/workflows/build-and-remove-template.yml` that tries to accomplish the following:
- Only build containers when they've been modified -- no need to build every container if their definitions haven't changed
- Split out the container build workflow so it can happen if needed in parallel to other tests.
- Update the docker build/push action to use the @v6 tag.


.github/workflows/build-and-remove-template.yml
```diff
@@ -0,0 +1,75 @@
...
+        if: ${{ inputs.always_build }}
+        run: |
+          echo "Input for job 'always_build' is true -- building regardless of detected changes."
+          echo "changed=true" >> $GITHUB_ENV
+
+      - name: Build Docker image
+        if: env.changed == 'true'
+        uses: docker/build-push-action@v6
+        with:
+          context: ${{ inputs.context || inputs.path }}/.
```
agitter, line 63: I see, is this why we don't need a default context input?

.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,69 @@
+name: Build SPRAS Containers
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
agitter, line 5: I checked options at https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request and these three look good. I'm still not sure what synchronize is even after looking at the docs.

.github/workflows/test-spras.yml
```diff
@@ -61,141 +61,6 @@ jobs:
...
-        path: docker-wrappers/OmicsIntegrator1/.
-        dockerfile: docker-wrappers/OmicsIntegrator1/Dockerfile
-        repository: reedcompbio/omics-integrator-1
-        tags: latest
-        cache_froms: reedcompbio/omics-integrator-1:latest
-        push: false
-    - name: Remove Omics Integrator 1 Docker image
-      # Remove the image to prevent the cache from being used. Here we use
-      # `|| true` to prevent the job from failing if the image doesn't exist or
-      # can't be removed for some reason
```
agitter, line 100: Could transfer this comment to the new `docker rmi`

.github/workflows/build-containers.yml
agitter, line 1: We will need to update Step 5 of the contribution guide to modify this workflow instead of `test-spras.yml`. Specifically the paragraph
> Extend .github/workflows/test-spras.yml to pull and build the new Docker image.

.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,69 @@
...
+    with:
+      path: docker-wrappers/OmicsIntegrator2
+      container: reedcompbio/omics-integrator-2
+      tag: latest
+  build-and-remove-pathlinker:
+    uses: "./.github/workflows/build-and-remove-template.yml"
+    with:
+      path: docker-wrappers/PathLinker
+      container: reedcompbio/pathlinker
+      tag: v2
```
agitter, line 28: I haven't been very thoughtful about which tags to test. Should we go back to testing latest for all of these?

agitter (changes requested):
This new design is excellent. The only non-trivial comment is about updating the contributing guide.

.github/workflows/build-and-remove-template.yml
```diff
@@ -0,0 +1,75 @@
...
+        required: true
+        type: string
+      tag:
+        required: true
+        type: string
+      always_build:
+        required: false
+        type: boolean
+        default: false
+      context:
```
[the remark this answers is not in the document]
jhiemstrawisc: The context lets Docker answer the question "where should I be scoped when if I need to start copying files from relative paths, as defined in the Dockerfile". For all the PRMs, the context should be set to the the same directory that holds the Dockerfile. However, building the SPRAS container requires that the context is set to the root of the repo so that the Dockerfile can copy everything it needs to.

The context isn't set as required because none of the PRM containers need to provide it (the default is simply the path to the PRM Dockerfile), but the SPRAS container passes the repo root as a context. See the `build-and-remove-spras` step from `build-containers.yml`.

.github/workflows/build-containers.yml
jhiemstrawisc: Agreed, they do a poor job of documenting it, but it basically means "whenever someone updates the PR":
https://github.com/orgs/community/discussions/24567

.github/workflows/test-spras.yml
jhiemstrawisc: Addressed in a new commit.

.github/workflows/build-containers.yml
jhiemstrawisc: These tags are arbitrary -- they're set for the image that gets created in the action, but the image is immediately deleted and never pushed because all we're testing is that it builds. We could set the tags to anything we want! After reflecting a bit, I think I'll just remove the tag field as an input and always set things to latest.

.github/workflows/build-containers.yml
jhiemstrawisc: Addressed in a recent commit.

.github/workflows/build-containers.yml
jhiemstrawisc: I removed the input in a recent commit.

.github/workflows/build-and-remove-template.yml
jhiemstrawisc: There's no sensible global default, since each context needs to be scoped to a specific directory. Every test must provide an input path, so I use that as a test-specific default.

.github/workflows/build-and-remove-template.yml
```diff
@@ -0,0 +1,74 @@
+name: Build and Remove Docker Image
+# A workflow template that handles building+removing SPRAS containers.
+
+# Required inputs are:
+# - path: The path to a container's Dockerfile directory, relative to the repo root.
+# - container: The registry/container to be built, without a tag.
+# - tag: The tag to be built.
```
agitter, line 7:
```suggestion
```

agitter (commented):
One more tag-related removal and then I'll merge

agitter (approved):

PR #200, opened 2024-12-09
Add auto container release action

jhiemstrawisc:
Now whenever a new tag of SPRAS is pushed to github, the corresponding container from that tag is build and pushed to Dockerhub.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,43 @@
+name: Release SPRAS on new tags
+# This action runs whenever a new tag matching vX.Y.Z is pushed to the repo.
+# Its job is to build the `spras` container and push it to DockerHub with the
+# 'latest' tag, and a tag matching the one supplied to GitHub.
+
+on:
+  push:
+    tags:
+      # run release on any version tag vX.Y.Z, e.g. v0.2.1
+      - v[0-9]+\.[0-9]+\.[0-9]+
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ vars.DOCKER_USERNAME }}
+          # Passwords are set up through the GitHub control pannel. The docs for
+          # the login action are explicit about never using a real password, and
+          # instead setting up a DockerHub personal access token:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ vars.DOCKER_USERNAME }}
```
agitter, line 25: It turns out I added a DockerHub personal access token to this repository in 2021 and forgot about it
image

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
+name: Release SPRAS on new tags
+# This action runs whenever a new tag matching vX.Y.Z is pushed to the repo.
+# Its job is to build the `spras` container and push it to DockerHub with the
+# 'latest' tag, and a tag matching the one supplied to GitHub.
+
+on:
+  push:
+    tags:
+      # run release on any version tag vX.Y.Z, e.g. v0.2.1
+      - v[0-9]+\.[0-9]+\.[0-9]+
```
agitter, line 10: Our [tags](https://github.com/Reed-CompBio/spras/tags) so far have not had a leading v. Should we keep that convention and change the pattern here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
+name: Release SPRAS on new tags
+# This action runs whenever a new tag matching vX.Y.Z is pushed to the repo.
```
agitter, line 2: Adjust depending on what we decide below regarding tag pattern

agitter (commented):
We need to decide on the tagging convention. Then we can merge and test it by making a release.

.github/workflows/release.yml
jhiemstrawisc: Fixed in the latest commit.

.github/workflows/release.yml
jhiemstrawisc: Fixed in the latest commit.

.github/workflows/release.yml
jhiemstrawisc: Nice!

agitter (approved):

agitter:
This is merged. Should I make a tag and release to test it?

agitter:
I made release [`0.3.0`](https://github.com/Reed-CompBio/spras/releases/tag/0.3.0) to test this and the action [failed](https://github.com/Reed-CompBio/spras/actions/runs/13081768261/job/36506644826#step:4:10). I'll try a successful login before releasing again.

PR #202, opened 2025-01-30
Set up Read The Docs (RTD) Scaffolding and Sketch Docs Website

jhiemstrawisc:
This sets up some of the basics the SPRAS team has discussed w.r.t. building an official documentation website. I tried to focus on organization/scaffolding for the docs, and not so much the content (which I'm less qualified to write).

I also spent some time putting together some meta docs to explain to others how to write/build/visualize any documentation changes they may choose to make. My hope is that I can help the group with formatting, structure, and hosting, and others can work on the biology

While the changes in this PR function to _generate_ documentation, this doesn't yet address the actual hosting of any documentation, which is outside the scope of what I can accomplish in a PR because it requires RTD account creation and the generation of github secrets.

Here's a screenshot of the landing page these changes currently produce:
Screenshot 2025-01-30 at 11 52 49


jhiemstrawisc:
A few questions I have at this point:
1. What should we do about the existing `doc/` folder, that holds a few markdown files and an image that's linked into the root README? It'll be confusing to have both a `doc/` and a `docs/` folder, but the contents of `doc/` really don't belong in `docs/` because they're not built into the website.
2. Funny enough, we already had a lot of the sphinx and read the docs dependencies baked into our `pyproject.toml` and `environment.yml`. I suspect these really aren't dependencies unless you're planning to work on documentation. Does anyone foresee any issues moving these into the "optional" sections of those dependency declarations?

.pre-commit-config.yaml
```diff
@@ -24,3 +24,14 @@ repos:
     rev: 'v0.0.269'
     hooks:
       - id: ruff
+  # Hook to enforce a consistent markdown linewrap policy
```
agitter, line 27: I actually dislike line wrapped markdown. The preference comes from our Manubot project for writing collaborative manuscripts with markdown, where we [recommend](https://github.com/manubot/rootstock/blob/main/USAGE.md#manubot-markdown) one sentence per line. I would be okay enforcing that with a hook.

docs/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+This guide explains how to work with SPRAS documentation using `sphinx` and Read
+The Docs.
+
+## Background Information
+
+[Read The Docs](https://docs.readthedocs.com/platform/stable/), or RTD, is the
+documentation hosting platform the SPRAS team has chosen to use for hosting its
+official documentation. As a platform, it automatically builds and versions docs
+based on the contents of the `docs/` file in this repository by periodically
```
agitter, line 11:
```suggestion
based on the contents of the `docs/` directory in this repository by periodically
```

docs/README.md
```diff
@@ -0,0 +1,92 @@
...
+polling for changes. When a new SPRAS version is detected, Read The Docs pulls
+that tag, uses it to build the relevant HTML content, and then hosts the
+newly-versioned HTML content on its website as `https://spras.readthedocs.io`.
+
+While Read the Docs is the platform that builds and hosts documentation, it is
+distinct from the actual tool/package responsible for generating the HTML
+content. That tool is [`sphinx`](https://www.sphinx-doc.org/en/master/), a
+popular python python package that tries to make documentation easy to write,
```
agitter, line 19:
```suggestion
popular python package that tries to make documentation easy to write,
```

docs/README.md
```diff
@@ -0,0 +1,92 @@
...
+- [Read The Docs Tutorial](https://docs.readthedocs.com/platform/stable/tutorial/index.html):
+  This tutorial shows you how a new project could be set up with RTD, and
+  highlights the steps SPRAS has already taken to connect the SPRAS repository
+  with the RTD workflow. It may be useful to better understand how RTD functions
+  as a platform.
+- [Sphinx Tutorial](https://sphinx-tutorial.readthedocs.io/): This tutorial (and
+  in general, this website) walks you through the basics of using sphinx to
+  create documentation. In particular, it helps you figure out how to achieve
+  the information layouts you may be after, such as certain content appearing in
+  a ToC.
```
agitter, line 39:
```suggestion
  a table of contents.
```

docs/README.md
```diff
@@ -0,0 +1,92 @@
...
+```bash
+make html
+```
+
+The output of this command will indicate whether the build was successful, and
+whether there are any warnings you should consider. It's generally bad to ignore
+these warnings, so if you see any, try to understand what they mean and how you
+can fix them.
+
+Once you have a successful build, the [\_build](_build) directory will contain
```
agitter, line 70: Is this going to be committed? The text below suggests no, should should we remove the link? We already ignore it.

docs/about.rst
```diff
@@ -0,0 +1,10 @@
+About SPRAS
+============
+
+SPRAS, short for "Signalling Pathway Reconstruction Streamliner Analysis" is a work-in-progress dockerized library
```
agitter, line 4:
```suggestion
SPRAS, short for "Signaling Pathway Reconstruction Analysis Streamliner" is a work-in-progress dockerized library
```

docs/conf.py
```diff
@@ -0,0 +1,39 @@
...
+
+sys.path.insert(0, os.path.abspath('..'))
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'SPRAS'
+copyright = '2025, Anthony Gitter & Anna Ritz'
+author = 'Anthony Gitter & Anna Ritz'
+release = '0.2.0'
```
agitter, line 20: Can we pull this version from somewhere else? I have now bumped to `0.3.0` and will soon bump again to test the Docker action.

docs/contributing.rst
```diff
@@ -0,0 +1,10 @@
+Contributing to SPRAS
+=====================
+
+We welcome contributions to SPRAS! If you would like to contribute, please follow the guidelines below.
+
+Adding New PRMs
```
agitter, line 6: I'd like to avoid the PRM algorithm in the docs. We could call these "Algorithms" generically if space is limited or "Network Algorithims" if we have a little more space.

@annaritz what do you think?

docs/fordevs/spras.rst
```diff
@@ -0,0 +1,82 @@
+SPRAS Python Modules
+====================
+
+.. automodule:: spras
```
agitter, line 4: We'll have to add these for each new module manually?

docs/prms/allpairs.rst
```diff
@@ -0,0 +1,4 @@
+All Pairs
+=========
+
+Here's a description of the PRM.
```
agitter, line 4: Can we link this to the auto-generated API documentation for the module?

docs/prms/prms.rst
```diff
@@ -0,0 +1,44 @@
+Pathway Reconstruction Methods
```
agitter, line 1: This is another area where we'll need to update the contribution guide if the list of algorithms needs to be extended manually.

environment.yml
```diff
@@ -24,6 +24,6 @@ dependencies:
   - mock=4.0
   - recommonmark=0.7
   - sphinx=6.0
+  - sphinx-rtd-theme=3.0.1
   - pip:
      - graphspace_python==1.3.1
```
agitter, line 29: I believe the sphinx dependencies came from [graphspace](https://github.com/adbharadwaj/graphspace-python/blob/master/requirements.txt). That is currently a required packaged for one of our pathway output file formats, so all of its dependencies may also be required.

agitter (changes requested):
The initially structure looks really great. I understand most of the pieces but not necessarily how every single file fits together.

> What should we do about the existing doc/ folder

I believe we can convert all the markdown files there into something that would be part of the SPRAS docs. Would that involve converting to rst?

docs/conf.py
agitter: One related project https://github.com/search?q=repo%3Amanubot%2Fmanubot%20__version__&type=code

agitter:
I logged in to readthedocs to see if I could provide authorization. I found that "Read the Docs Community (readthedocs.org)" has already been granted organization access by Reed-CompBio because, as Anna told us, they've used it for other projects.

Then I added a project and found the Reed-CompBio/spras GitHub repo. The setup wants us to add `.readthedocs.yaml` to the root of this repo. Their example file is
```
# Read the Docs configuration file
# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details

# Required
version: 2

# Set the OS, Python version, and other tools you might need
build:
  os: ubuntu-24.04
  tools:
    python: "3.13"

# Build documentation in the "docs/" directory with Sphinx
sphinx:
   configuration: docs/conf.py

# Optionally, but recommended,
# declare the Python requirements required to build your documentation
# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
# python:
#    install:
#    - requirements: docs/requirements.txt
```

Then it tried to run a build, which failed because we are missing that config file.
```
Config file not found at default path The required .readthedocs.yaml configuration file was not found at repository's root. Learn how to use this file in our [configuration file tutorial](https://docs.readthedocs.io/en/stable/config-file/index.html).
```

docs/README.md
jhiemstrawisc: You're right, we won't be committing it (in fact, it's existed in the `.gitignore` since before this PR). I'll remove the hyperlink.

docs/conf.py
jhiemstrawisc: I did some poking around, and it looks like the way `manubot` does has [fallen slightly out of flavor](https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module). Instead, I'll use the approach suggested in the top SO answer, where I set it in `pyproject.toml` and use that to populate other places (including setting the canonical `__version__` dunder).

docs/contributing.rst
jhiemstrawisc: I'll update the template for now. Once we merge this scaffolding, anyone is welcome to adapt the placeholder text I'm using throughout.

docs/fordevs/spras.rst
jhiemstrawisc: I think I found a cleaner way to do this using the `sphinx-apidoc` tool. See the "Auto generate all API module rst files" commit.

docs/conf.py
agitter: We discussed your new solution, and I like it. We may need a try/except in case the package is not installed. This also makes stronger assumptions that spras runs as an installed package, so we plan to update the readme and test case accordingly so the test case installs the local source into the activated conda environment.

agitter (approved):

PR #207, opened 2025-03-03
Param tuning code integration: ensembling

ntalluri:

PR #209, opened 2025-03-03
Param tuning code integration: pca chosen

ntalluri:


spras/analysis/ml.py
```diff
@@ -142,8 +142,14 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_var: str, output_coord:
     if not isinstance(labels, bool):
         raise ValueError(f"labels={labels} must be True or False")
 
-    scaler = StandardScaler()
+    #TODO: MinMaxScaler changes nothing about the data
+    # scaler = MinMaxScaler()
```
ntalluri, line 146: How to do PCA on Binary Data

ntalluri: https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html
https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html (data is not in a gaussian distribution; does not make sense to use standard scalar)

ntalluri:
@agitter Do this PR Last

Will need to merge with updated master after https://github.com/Reed-CompBio/spras/pull/193, https://github.com/Reed-CompBio/spras/pull/207 is merged, and https://github.com/Reed-CompBio/spras/pull/208. (hopefully this will remove the repeated files through out the PRs)
- there will be merge conflicts with the Snakefile, evaluation.py and the test suite.

Included in this PR:

1. update to evaluation.py that updates the code to include precision_and_recall and pca chosen pathway (precision_and_recall is also used in https://github.com/Reed-CompBio/spras/pull/208)
2. a new test suite evaluate for pca chosen pathway
3. updates to Snakemake file that will run evaluation per dataset and per algortihm-dataset pair for pca chosen pathway
4. update to ml code for PCA to add centroids
5. still need to figure out how to rescale the binary data
6. update to ml test suite for expected pca coordinates and ml test code

spras/analysis/ml.py
ntalluri: https://stackoverflow.com/questions/40795141/pca-for-categorical-features

spras/analysis/ml.py
ntalluri, line 1: for the create palette function, update the code to have a sorted list of unique column headers

unique_column_names = list(sorted(set(column_names)))
    custom_palette = sns.color_palette(palette = "tab20c", n_colors = len(unique_column_names))
    label_color_map = {label: color for label, color in zip(unique_column_names, custom_palette, strict=True)}
    return label_color_map

ntalluri:
fix in this pr #231

test/evaluate/expected/expected-pr-per-pathway-pca-chosen.txt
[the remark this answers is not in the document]
ntalluri: # calculating the centroid
 centroid = np.mean(X_pca, axis=0) # mean of each principal component across all samples

spras/evaluation.py
```diff
@@ -72,29 +77,88 @@ def load_files_from_dict(self, gold_standard_dict: Dict):
...
+        for coordinates_file in coordinates_files:
+            coord_df = pd.read_csv(coordinates_file, delimiter="\t", header=0)
+
+            centroid_row = coord_df[coord_df['datapoint_labels'] == 'centroid']
+            centroid = centroid_row.iloc[0, 1:].tolist()
+            coord_df = coord_df[coord_df['datapoint_labels'] != 'centroid']
+
+            pc_columns = [col for col in coord_df.columns if col.startswith('PC')]
+            coord_df['Distance To Centroid'] = np.sqrt(sum((coord_df[pc] - centroid[i]) ** 2 for i, pc in enumerate(pc_columns)))
+            closest_to_centroid = coord_df.sort_values(by='Distance To Centroid').iloc[0]
```
ntalluri, line 159: how to deal with tiebreaker?

ntalluri: If tied add all ties to the rep_pathway?

test/evaluate/expected/expected-pr-per-pathway-pca-chosen.txt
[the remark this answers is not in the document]
agitter: Why did you expect a different pathway to be selected as closest to the centroid?

Do we want to eliminate empty pathways from the closest-to-centroid selection if there exist any non-empty alternatives?

spras/analysis/ml.py
agitter: The background section of https://arxiv.org/abs/1510.06112 gives some good explanation and earlier work. I'm doubtful sklearn supports any of those options directly. A previous graduate student worked on general matrix factorization with multiple types of data and wrote custom code to support different data types, e.g. binary.

If we don't find correct PCA implementations, we could switch to a different visualization like multidimensional scaling on the Jaccard similarities. Or use option 2 above and document in the code and manuscript why we chose that even though it is not what we really want.

ntalluri:
Screenshot 2025-06-12 at 11 36 08

*will update with comments on centroid problems

agitter:
Regarding empty pathway, I see them as having two different roles in the PCA plots. When we use PCA (or its replacement) to summarize output pathways in a 2D space, I am okay including empty pathways. That be informative across algorithms.

However, I'm inclined to agree that when we are using PCA representations to choose the typical algorithm output, we should exclude empty outputs. There would be different ways to do this in practice. One would be to run PCA on all pathways but remove the points from the 2D space that correspond to empty pathways before picking the centroid.

When picking the centroid, our current idea to workaround the problem in images A and B above is to use a kernel density estimator in the 2D space. Then we could find the point with the greatest smoothed density and finally the (non-empty) pathway with the parameter combination that gave a point closest to that point.

ntalluri:
For removing empty pathways options, should this also happen in all of the ml post processing? (HAC, Jaccard similarity, Ensembling)

agitter:
> For removing empty pathways options, should this also happen in all of the ml post processing?

My initial thought is no. We have done a lot of past ensembling with Omics Integrator 1 and left in empty pathways. It decreases the absolute frequencies, but that doesn't matter for edge ranking. Finding there different algorithms produce empty pathways that all co-cluster seems okay, if uninteresting.

Snakefile
```diff
@@ -95,6 +96,9 @@ def make_final_input(wildcards):
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-matrix.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-heatmap.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
     
+        if generate_kde:
```
ntalluri, line 99: I don't think we need to save the KDE plot, but I left temporary code in case we want to.

config/config.yaml
```diff
@@ -139,10 +139,9 @@ reconstruction_settings:
 
   #set where everything is saved
   locations:
-
     #place the save path here
     # TODO move to global
-    reconstruction_dir: "output"
+    reconstruction_dir: "output/config"
```
ntalluri, line 144:
```suggestion
    reconstruction_dir: "output"
```

ntalluri:
This idea will only choose a pathways rom pathway reconstruction algortihms if it has multiple parameter combinations. Allpairs and those with 1 chosen parameter combination will never be given a pca chosen pathway. Should we add the one pathway it makes to the list of representative pathways.

test/evaluate/expected/expected-pr-per-pathway-pca-chosen.txt
ntalluri, line 1: Locally this is what my expected pathway is, but I think on github actions, the expected is 789 pathway

ntalluri:
The visualization that sns.kde has looks better than the one I am using right now. (TODO: add images to show)

spras/analysis/ml.py
```diff
@@ -153,20 +169,94 @@ def pca(dataframe: pd.DataFrame, output_png: str, output_var: str, output_coord:
...
+
+        if len(max_rows) > 1:
+             # TODO: what if there are > 1 maxes? how do we deal with tiebreakers?
+                # 1) choose the one closest to the centroid or the kde center of mass (weighted average of coordinates, where the weights come from the KDE density at each point)
+                # 2) keep all of the maxes?
+                # 3) taking the mean of median of the maxes coordinates
+
+            # TODO: TEMP; still choosing the one maximum for now
+            max_row = max_rows.iloc[0]
+            kde_row = ['kde_peak', max_row["X_coordinate"], max_row["Y_coordinate"]]
```
ntalluri, line 251: closest one to 0.0 or some value

ntalluri:
How to deal with KDE Max tiebreakers:

When deciding which maximum KDE point to highlight in a 2D PCA plot, I think it makes much more sense to use the point closest to the centroid instead of the origin (0,0). The centroid actually reflects where the data points are concentrated of the data in PCA space. On the other hand, (0,0) is just a fixed mathematical point that doesn’t really tell me anything about how my data are distributed. Especially after PCA, the points often aren’t centered at the origin because of how the variance is spread across components. By using the centroid, I’m making sure that the selected KDE peak represents the overall structure and central tendency of my data, rather than just being arbitrarily close to (0,0).

The centroid is just the average location of all points in PCA space(I take the mean value of the top 2 PCA component across my points). It mathematically reflects the "center of mass" of my data, unlike (0,0), which is just a fixed point that doesn't consider where the data actually are.

ntalluri:
How to deal with multiple representative pathways tiebreakers:

The plan is to send summary.txt to help find pathways that are the smallest. Smallest in this case means smallest number of edges and nodes. If they are all equal, then do it based on name to keep it deterministic. 

1) choose smallest number of edges and smallest number of nodes
2) end all be all, choose the first one based on name

in the future, we can choose other stats to use for tiebreaking

ntalluri:
Due to different machines rounding differently, the test case for pca chosen kept failing. I tried rounding everything to the 8th decimal and this seemed to fix the issue. I will need to go back through and update it to be more careful on what rows it rounds on.

Snakefile
agitter: Per our meeting discussion, the plot is nice to have but if it introduces a great deal of additional complexity in terms of new code or the Snakemake rule, we can skip it.

config/config.yaml
```diff
@@ -172,6 +171,17 @@ analysis:
...
     metric: 'euclidean'
+    # adds kernel density to pca plots
+    # default metric used is euclidean for kernel_density
+    kernel_density: true
+    # hyperparameter to control the smoothness of estimated kernel density
+    # single float, 'scott', 'silverman'
+    bandwidth: 1.0
+    # hyperparameter kernel to use for kernel_density
+    # 'gaussian', 'tophat', 'epanechnikov', 'exponential', 'linear', 'cosine'
+    kernel: 'gaussian'
```
agitter, line 182: I noted this in a comment elsewhere already. If we don't save the PCA plots, I'm not sure how a user will know how to tune these, which may imply we don't need to expose them.

agitter:
There are new merge conflicts to resolve

agitter (commented):
I only have a few remaining comments about documentation and other open pull requests that will need updates after this is merged. We can finalize this during our next meeting.

test/ml/expected/expected-pca-coordinates-kde.tsv
```diff
@@ -0,0 +1,7 @@
+datapoint_labels	PC1	PC2
+test-data-s1	-1.01220906	0.05003395
```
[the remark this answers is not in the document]
ntalluri: This most likely will change with the update in #310.

spras/config.py
```diff
@@ -308,3 +315,8 @@ def process_config(self, raw_config):
         # Only run Evaluation per algorithm if ML per algorithm is set to True
         if not self.analysis_include_ml_aggregate_algo:
             self.analysis_include_evaluation_aggregate_algo = False
+
+        # Set kernel_density to True if Evaluation is set to True
```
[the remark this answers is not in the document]
ntalluri: This doesn't get preset (but also none of the other pca parameters are preset to a value).

config/config.yaml
```diff
@@ -189,6 +188,11 @@ analysis:
     linkage: 'ward'
     # 'euclidean', 'manhattan', 'cosine'
     metric: 'euclidean'
+    # controls whether kernel density estimation (KDE) is computed and visualized on top of PCA plots.
```
ntalluri, line 191: add in a comment why this needs to be set to true for evaluation

agitter (approved):
I made the final comments about KDE and will merge if tests pass

PR #213, opened 2025-04-09
Update Snakemake HTCondor Executor instructions with additional troubleshooting

jhiemstrawisc:
There's a new error floating around for people using the HTCondor/Snakemake executor that's the result of a breaking change in HTCondor config. The fix for this error has been merged in the upstream executor's main branch for ~1 week, but several people have reached out to me unsure of how to fix it.

This new section of documentation explains how to upgrade your spras conda env in a way that forces re-installation of the executor, which should get rid of the error.

Note that it wasn't enough to run `pip install --upgrade`, which is why this approach uses `pip install --force-reinstall`.


agitter (approved):
The instructions make sense so I'll merge this right away

PR #195, opened 2024-11-18
Integrate dsub

jhiemstrawisc:
Opening this PR in draft mode for now while @nisthapanda work through collaborating on documentation updates and other testing.

PR #220, opened 2025-05-22
chore: format docker errors

tristan-f-r:
This adds a note for context (what errors are coming from wrapping the docker daemon, what are coming from actually running the container), and prints the stacktrace instead of just the error message.

(Didn't notice that I didn't have docker-in-docker installed, and encountered a confusing HTTP error since I didn't have docker installed in my developer environment.)

### DIFF
--- a/spras/containers.py
+++ b/spras/containers.py
@@ -2,6 +2,7 @@ import os
 import platform
 import re
 import subprocess
+import traceback
 from pathlib import Path, PurePath, PurePosixPath
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -166,7 +167,11 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     out = None
     try:
         # Initialize a Docker client using environment variables
-        client = docker.from_env()
+        try:
+            client = docker.from_env()
+        except Exception as err:
+            err.add_note("An error occurred when fetching the docker daemon: is docker installed?")
+            raise err
         # Track the contents of the local directories that will be bound so that new files added can have their owner
         # changed
         pre_volume_contents = {}
@@ -233,7 +238,7 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
         client.close()
 
     except Exception as err:
-        print(err)
+        print(traceback.format_exc())
     # Removed the finally block to address bugbear B012
     # "`return` inside `finally` blocks cause exceptions to be silenced"
     # finally:


tristan-f-r:
It does!

tristan-f-r:
This broke all of the singularity test cases - I'll fix those now.

tristan-f-r:
Hm, it seems that these changes made some of the errors from Singularity that were usually just caught and printed bleed to the main python context - I'll have to install Singularity locally to properly fix these cases.

tristan-f-r:
Took a good while to get singularity installed, but that latest commit works locally :+1:

spras/containers.py
```diff
@@ -148,6 +150,42 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+                else:
+                    print("Note: This is an unknown message format - if you want this pretty printed, please file out an issue.")
+                    out = str(out)
+            elif not isinstance(out, str):
+                out = str(out, "utf-8")
+            print(indent(out))
+    except docker.errors.ContainerError as err:
+        print(f"(Command formatted as list: `{err.command}`)")
+        print(f"An unexpected non-zero exit status ({err.exit_status}) inside the docker image {err.image} occurred:")
+        err = str(err.stderr if err.stderr is not None else "", "utf-8")
```
tristan-f-r, line 185: The primary piece about this PR, as mentioned in slack, is that these `err: str` are constructed per-runtime. In another PR, we can have these errors be filtered for messages that we know are thrown by containers in certain situations, and we can handle them appropriately, while sufficiently raising exceptions of unknown errors to stop the Snakemake runtime and report those issues to SPRAS as a bug.

The current way that errors are handled have caused Aden and I a good amount of trouble when wrapping algorithms, since it's quite hard to tell that the output even is an error, and, if the algorithm that's being wrapped happens to touch the output file, it can sometimes make tests pass when they should fail.

spras/containers.py
```diff
@@ -148,6 +150,42 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+    Runs a command in the container using Singularity or Docker with associated pretty printed messages.
+    @param framework: singularity or docker
+    @param container_suffix: name of the DockerHub container without the 'docker://' prefix
+    @param command: command to run in the container
+    @param volumes: a list of volumes to mount where each item is a (source, destination) tuple
+    @param working_dir: the working directory in the container
+    @param environment: environment variables to set in the container
+    @return: output from Singularity execute or Docker run
+    """
+    print('Running {} on framework "{}" with command: {}'.format(name, framework, ' '.join(command)), flush=True)
```
jhiemstrawisc, line 164: I've always thought the word "framework" is a strange way to expose the name of the container runtime to users. Personally, I'd either explicitly say "with container framework {}" or maybe "with container runtime {}". Of course, it's also nice to use consistent language for developers, and switching "framework" to "runtime" may confuse people.

Just my two cents on this one -- up to you whether you want to change anything!

spras/containers.py
```diff
@@ -148,6 +150,42 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+    try:
+        out = run_container(framework=framework, container_suffix=container_suffix, command=command, volumes=volumes, working_dir=working_dir, environment=environment)
+        if out is not None:
+            if isinstance(out, list):
+                out = ''.join(out)
+            elif isinstance(out, dict):
+                if 'message' in out:
+                    # singularity message - lets print it.
```
jhiemstrawisc, line 172: It'd be useful drop a comment here that points to documentation describing whatever apptainer output format you're tapping into.

spras/containers.py
```diff
@@ -161,79 +199,79 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
-        # Raised on non-Unix systems
-        except AttributeError:
-            pass
-
-        # TODO: Not sure whether this is needed or where to close the client
-        client.close()
-
     except Exception as err:
-        print(err)
+        err.add_note("An error occurred when fetching the docker daemon: is docker installed and is dockerd running?")
```
jhiemstrawisc, line 209: I'm actually not totally sure of the answer to this... Does SPRAS function with the Docker API from other runtimes like Podman? If so, it may be worth providing a more generic error message, e.g.
```
err.add_note("An error occurred when connecting to the Docker API: is Docker (or a compatible container runtime) installed and running?")
```

At the same time, maybe this is too mountain-dewy and the vast majority of people running SPRAS on their personal machine.

spras/containers.py
```diff
@@ -148,6 +150,42 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+            if isinstance(out, list):
+                out = ''.join(out)
+            elif isinstance(out, dict):
+                if 'message' in out:
+                    # singularity message - lets print it.
+                    if 'return_code' in out and not out['return_code'] == 0:
+                        print(f"(Program exited with non-zero exit code '{out['return_code']}')")
+                    out = ''.join(out['message'])
+                else:
+                    print("Note: This is an unknown message format - if you want this pretty printed, please file out an issue.")
```
jhiemstrawisc, line 177: Since the SPRAS repo is unlikely to change, you might even provide a link to
https://github.com/Reed-CompBio/spras/issues/new

jhiemstrawisc (commented):
A few minor comments. You're free to take them or leave them!

Otherwise, looks good to me.

spras/containers.py
tristan-f-r: Unfortunately, when I was experimenting with Podman support, the docker-py API is not compatible with podman.

spras/containers.py
```diff
@@ -2,12 +2,14 @@
 import platform
 import re
 import subprocess
+import traceback
```
agitter, line 5: My IDE reports this isn't used

spras/analysis/cytoscape.py
```diff
@@ -46,14 +46,12 @@ def run_cytoscape(pathways: List[Union[str, PurePath]], output_file: str, contai
...
-
     container_suffix = "py4cytoscape:v3"
-    out = run_container(container_framework,
-                        container_suffix,
-                        command,
-                        volumes,
-                        work_dir,
-                        env)
-    print(out)
+    run_container_and_log('Cytoscape',
```
agitter, line 50: Does this new approach to calling the container helper function change anything in the contributing guide?

spras/containers.py
```diff
@@ -148,6 +150,43 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+                out = ''.join(out)
+            elif isinstance(out, dict):
+                if 'message' in out:
+                    # This is the format of a singularity message.
+                    # See https://singularityhub.github.io/singularity-cli/api/source/spython.main.html?highlight=execute#spython.main.execute.execute.
+                    if 'return_code' in out and not out['return_code'] == 0:
+                        print(f"(Program exited with non-zero exit code '{out['return_code']}')")
+                    out = ''.join(out['message'])
+                else:
+                    print("Note: This is an unknown message format - if you want this pretty printed, please file out an issue at https://github.com/Reed-CompBio/spras/issues/new.")
```
agitter, line 178: "file" or "fill out" instead of "file out"

spras/containers.py
```diff
@@ -148,6 +150,43 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
...
+                    out = str(out)
+            elif not isinstance(out, str):
+                out = str(out, "utf-8")
+            print(indent(out))
+    except docker.errors.ContainerError as err:
```
agitter, line 183: `errors` reports `Cannot find reference 'errors' in '__init__.py' ` in my IDE. Do you see that as well?

spras/containers.py
```diff
@@ -148,6 +150,43 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
     else:
         raise ValueError(f'{framework} is not a recognized container framework. Choose "docker" or "singularity".')
 
+def run_container_and_log(name: str, framework: str, container_suffix: str, command: List[str], volumes: List[Tuple[PurePath, PurePath]], working_dir: str, environment: str = 'SPRAS=True'):
+    """
+    Runs a command in the container using Singularity or Docker with associated pretty printed messages.
+    @param framework: singularity or docker
```
agitter, line 156: Add `name` to docstring

spras/containers.py
```diff
@@ -148,6 +150,43 @@ def run_container(framework: str, container_suffix: str, command: List[str], vol
     else:
         raise ValueError(f'{framework} is not a recognized container framework. Choose "docker" or "singularity".')
```
agitter, line 151: Not in scope, but I noticed this error message is wrong and should include "dsub" as a choice.

spras/containers.py
```diff
@@ -161,79 +200,79 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     @param volumes: a list of volumes to mount where each item is a (source, destination) tuple
     @param working_dir: the working directory in the container
     @param environment: environment variables to set in the container
-    @return: output from Docker run
+    @return: output from Docker run, or will error if the container errored.
     """
     out = None
```
agitter, line 205: This may no longer be needed given the new error handling and control flow

spras/containers.py
```diff
@@ -161,79 +200,79 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
-        # TODO: Not sure whether this is needed or where to close the client
-        client.close()
-
     except Exception as err:
-        print(err)
+        err.add_note("An error occurred when fetching the docker daemon: is docker installed and is dockerd running?")
+        raise err
+    # Track the contents of the local directories that will be bound so that new files added can have their owner
+    # changed
+    pre_volume_contents = {}
```
agitter, line 214: Is it correct to assume nothing changed here, only whitespace?

agitter (commented):
A few comments and questions. I have not yet tested it locally.

spras/containers.py
tristan-f-r: I don't see that.

agitter: There isn't a stack trace, it is a static warning from PyCharm:
image

If you don't see it, I'm not concerned.

spras/containers.py
agitter: If dsub is missing other places, it was an oversight in #195 and almost certainly not an intentional omission.

agitter (approved):

Issue #105, opened 2023-07-20
Add DeRegNet


agitter:
De novo identification of maximally deregulated subnetworks based on multi-omics data with DeRegNet
https://doi.org/10.1186/s12859-022-04670-6

https://github.com/KohlbacherLab/deregnet

Requires Docker and Gurobi to run.

PR #225, opened 2025-05-23
feat: netmix2

### LINKED ISSUE #157: NetMix2
NetMix2: A Principled Network Propagation Algorithm for Identifying Altered Subnetworks
https://doi.org/10.1089/cmb.2022.0336

https://github.com/raphael-group/netmix2

tristan-f-r:
Closes #157. We need a SPRAS gurobi server license. Once this is integrated, this lays the ground for #105.


config/config.yaml
ntalluri, line 1: please update the file to have include: true for all that were turned to false

spras/netmix2.py
```diff
@@ -0,0 +1,103 @@
...
+
+        command = ['python',
+                   '/NetMix2/run_netmix2.py',
+                   '-el',
+                   network_file,
+                   '-gs',
+                   scores_file,
+                   '-o', mapped_out_prefix + '/v-output.txt']
+
+        # Add optional arguments
```
ntalluri, line 75: These parameters are not currently in the example config.yaml file. Please add them as parameters a user can set.

spras/netmix2.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            raise ValueError("Node prizes are required for NetMix2.")
+        node_df.to_csv(filename_map['scores'], index=False, columns=['prize', 'NODEID'], header=False, sep='\t')
+
+        edges_df = data.get_interactome()
+        edges_df.to_csv(filename_map['network'], index=False, sep='\t', columns=['Interactor1', 'Interactor2'], header=False)
+    
+    @staticmethod
+    def run(network=None, scores=None, output_file=None, delta=None, num_edges=None, density=None,
+            time_limit=None,
```
ntalluri, line 35: Should we preset a time_limit?

spras/netmix2.py
```diff
@@ -0,0 +1,103 @@
...
+                            work_dir)
+        print(out)
+
+        # Rename the primary output file to match the desired output filename
+        # Currently PathLinker only writes one output file so we do not need to delete others
+        output_vertices = out_dir / 'out' / 'v-output.txt'
+        output_vertices.rename(output_file)
+
+    @staticmethod
+    def parse_output(raw_pathway_file, standardized_pathway_file):
```
ntalluri, line 101: Netmix2 will not work with the ML post processing and possibly some of the other post processing. I'm not sure how to fix this at the moment, but this is something that will need to be considered.

test/parse-outputs/test_parse_outputs.py
ntalluri, line 1: Please add the expected-pathway file and input raw-pathway file

test/generate-inputs/test_generate_inputs.py
ntalluri, line 1: Please add the expected input file

spras/netmix2.py
```diff
@@ -0,0 +1,103 @@
...
+from spras.dataset import Dataset
+from spras.prm import PRM
+
+__all__ = ['NetMix2']
+
+class NetMix2(PRM):
+    required_inputs = ['network', 'scores']
+
+    @staticmethod
+    def generate_inputs(data: Dataset, filename_map):
```
ntalluri, line 13: For directionality and edges, I just wanted to confirm that NetMix2 deals with only undirected inputs and can handle duplicate edges?

spras/netmix2.py
ntalluri: Is there a goal to allow for the other families other than the propagation family?

test/NetMix2/test_nm2.py
ntalluri, line 1: Please add test cases that includes the required parameters only, required + all optional parameters, missing any required parameters, and singularity test case.

config/config.yaml
tristan-f-r: Hi! Apologies for the late response on these reviews - this PR is a draft PR, so I didn't mean for anyone to do a deep review of its contents :sweat_smile:

spras/netmix2.py
tristan-f-r: I wasn't quite sure how to handle that, as `time_limit` feels like another option that should really be global to algorithms.

config/config.yaml
ntalluri: It's okay, I'm going through each PR in waiting.

spras/netmix2.py
tristan-f-r: Not at the moment.

spras/netmix2.py
tristan-f-r: Is there something bad about getting the induced subgraph that makes netmix2/RWR not work well with ML post processing?

spras/netmix2.py
tristan-f-r: [Undirected](https://github.com/raphael-group/netmix2/blob/110954e27ec1911b6fe68b6a34352fecca494922/src/common.py#L39-L40).

ntalluri:
This PR doesn't have the pydantic style. Reminder to update this PR.

config/config.yaml
```diff
@@ -147,6 +147,19 @@ algorithms:
       run1:
         n: 1
 
+  - name: "netmix2"
```
ntalluri, line 174: can we add some defaults here?

ntalluri:
getting this error when I try to run this locally: 

```
ImportError in file "/Users/nehatalluri/Desktop/research/spras/Snakefile", line 2:
cannot import name 'algorithms' from partially initialized module 'spras.runner' (most likely due to a circular import) (/Users/nehatalluri/Desktop/research/spras/spras/runner.py)
  File "/Users/nehatalluri/Desktop/research/spras/Snakefile", line 2, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/runner.py", line 11, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/netmix2.py", line 13, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/secrets.py", line 9, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/config/config.py", line 27, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/config/schema.py", line 17, in <module>
  File "/Users/nehatalluri/Desktop/research/spras/spras/config/algorithms.py", line 20, in <module>
```

spras/netmix2.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        for input_type in NetMix2.required_inputs:
+            if input_type not in filename_map:
+                raise ValueError(f"{input_type} filename is missing")
+
+        if data.contains_node_columns('prize'):
+            node_df = data.get_node_columns(['prize'])
+        else:
+            raise ValueError("Node prizes are required for NetMix2.")
+        node_df.to_csv(filename_map['scores'], index=False, columns=['prize', 'NODEID'], header=False, sep='\t')
```
akniyetnurbol: The order of the column should be changed, so it's node then prize, but also raw SPRAS prizes are random weights and can't be used as p-values, so np.exp(-prize) is needed to convert them to valid numbers from 0 to 1 as required by locfdr

import numpy as np
node_df['pval'] = np.exp(-node_df['prize'])
node_df.to_csv(filename_map['scores'], index=False, columns=['NODEID', 'pval'], header=False, sep='\t')

spras/netmix2.py
```diff
@@ -0,0 +1,120 @@
...
+
+        volumes = list()
+
+        bind_path, network_file = prepare_volume(inputs["network"], work_dir, container_settings)
+        volumes.append(bind_path)
+
+        bind_path, scores_file = prepare_volume(inputs["scores"], work_dir, container_settings)
+        volumes.append(bind_path)
+
+        bind_path, license_file = prepare_volume(inputs["gurobi_path"], work_dir, container_settings)
```
akniyetnurbol: the gurobi path is not in the inputs dict, so it causes an error when run() is used. Could use the gurobi_path variable from line 52

bind_path, scores_file = prepare_volume(gurobi_path, work_dir, container_settings)

spras/netmix2.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            raise ValueError("Node prizes are required for NetMix2.")
+        node_df.to_csv(filename_map['scores'], index=False, columns=['prize', 'NODEID'], header=False, sep='\t')
+
+        edges_df = data.get_interactome()
+        edges_df = convert_directed_to_undirected(edges_df)
+        edges_df.to_csv(filename_map['network'], index=False, sep='\t', columns=['Interactor1', 'Interactor2'], header=False)
+
+    @staticmethod
+    def run(inputs, output_file, args, container_settings=None):
```
akniyetnurbol, line 59: Check for no input cases in this function

test/NetMix2/test_nm2.py
```diff
@@ -0,0 +1,60 @@
+import shutil
+from pathlib import Path
+
+import pytest
+
+import spras.config as config
```
akniyetnurbol: import spras.config.config as config since init_from_file lives in spras/config/config.py

import spras.config.config as config

test/NetMix2/test_nm2.py
```diff
@@ -0,0 +1,60 @@
...
+config.init_from_file("config/config.yaml")
+
+TEST_DIR = Path('test', 'NetMix2/')
+OUT_FILE = Path(TEST_DIR, 'output', 'nm2-output.txt')
+
+
+class TestNetMix2:
+    def test_nm2_required(self):
+        OUT_FILE.unlink(missing_ok=True)
+        NetMix2.run(
```
akniyetnurbol, line 26: Inputs must be a dict, not keyword args (consistent with other SPRAS wrappers)
Also, if the container is set to None, then prepare_volume raises an error, so container_settings should be passed explicitly

NetMix2.run(
            {"network": TEST_DIR / 'input' / 'network-basic.txt',
             "scores": TEST_DIR / 'input' / 'scores-basic.txt'},
            output_file=OUT_FILE,
            args=NetMix2Params(num_edges=1),
            container_settings=config.config.container_settings
        )

docker-wrappers/NetMix2/requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy==1.19.5
+scipy==1.5.4
+matplotlib==3.3.4
+pandas==1.1.5
+statsmodels==0.12.2
+gurobipy==9.5.2
+networkx==2.5.1
```
akniyetnurbol, line 7: Should add packaging for patsy with a consistent version

packaging==24.0

test/NetMix2/test_nm2.py
```diff
@@ -0,0 +1,60 @@
...
+            num_edges=1,
+            delta=1,
+            density=5,
+            output_file=OUT_FILE
+        )
+        assert OUT_FILE.exists(), 'Output file was not written'
+
+    def test_nm2_missing(self):
+        with pytest.raises(ValueError):
+            NetMix2.run(
```
akniyetnurbol, line 52: TypeError because args was missing

NetMix2.run(
                {"network": TEST_DIR / 'input' / 'network-basic.txt'},
                output_file=OUT_FILE,
                args=NetMix2Params(),
                container_settings=config.config.container_settings #config.container_settings
            )

spras/netmix2.py
```diff
@@ -0,0 +1,120 @@
...
+        gurobi_path = gurobi()
```
akniyetnurbol, line 62: gurobi() function in spras/secrets.py is used to get the Gurobi license path, which is read from secrets.gurobi in config.yaml

When the user downloads the gurobi.lic file, they need to update the directory in config.yaml under "secrets:"
OR 
Place it at the location already indicated

akniyetnurbol (changes requested):
Comments describe the changes made to get NetMix2 running locally through SPRAS and the tests passing. Also, includes a note on how the Gurobi license is configured in config.yaml and accessed via secrets.py.

spras/netmix2.py
annaritz: I think we might want to add np.exp(-prize) as its own function in the same place we have the -log(p-val) function - that would keep these consistent since they transform scores into p-values and vice versa.

annaritz: Also, this doesn't exactly transform scores into p-values because there's no statistical test here - it's just converting them to numbers between 0 and 1 where lower is better.

annaritz:
Proposed edits look good, we'll check with @tristan-f-r to see whether you should do a PR to `tristan-f-r-:netmix2` to get the changes in.

ntalluri:
@akniyetnurbol I saw your email about the pval - prize prize - pval conversion. Would you be able to explain why in SPRAS we would need this conversion for Netmix2? In my opinion pval - prize conversions would happen before putting data into SPRAS.

spras/netmix2.py
ntalluri: Does Netmix require the values to be between 0 - 1 or is this a statistical test that the Netmix2 paper used to convert their data?

akniyetnurbol:
> @akniyetnurbol I saw your email about the pval - prize prize - pval conversion. Would you be able to explain why in SPRAS we would need this conversion for Netmix2? In my opinion pval - prize conversions would happen before putting data into SPRAS.

@ntalluri Yeah, when I was running snakemake there was an error because NetMix2 explicitly expects p-values as input (per its readme and paper), but spras stores node data as prizes using -log10(p-value). So by the time data enters spras it's already been converted from p-values to prizes. The generate_inputs wrapper converts back using 10^(-prize), which recovers the original p-value. I put it in the wrapper so the user doesn't have to handle this themselves just to keep things consistent with other algorithms in spras.

spras/netmix2.py
akniyetnurbol: By my understanding, it's less of a statistical choice and more of a hard requirement in the code based on how NetMix2 processes inputs. It converts scores to z-scores using norm.ppf, which only works for values in (0, 1], so the input just needs to be a number in that range where lower is better.

annaritz:
I think we are nearly ready to review this PR, after @akniyetnurbol's changes. I added the gurobi license to this repo as a secret.

There seems to be two unresolved conflicts in `spras/runner.py` related to how algorithms are loaded - @akniyetnurbol or @tristan-f-r can you resolve those?  The remaining tests will trigger once the conflict is resolved.

annaritz:
In talking with @akniyetnurbol, it seems like we might need to set a GitHub Environment for the gurobi license to work. Will wait to see if the tests fail before adding that.

.github/workflows/netmix2-license.yml
agitter, line 1: What is the intended testing sequence? Does this workflow run only if the main test workflow completes? Is there anything preventing NetMix2 tests from running in the main test workflow, or is that why it fails?

.github/workflows/netmix2-license.yml
```diff
@@ -0,0 +1,45 @@
...
+    types: [completed]
+
+permissions:
+  contents: read
+
+jobs:
+  approval_gate:
+    if: github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion == 'failure'
+    runs-on: ubuntu-latest
+    environment: safe-to-test
```
agitter, line 15: What does this do?

.github/workflows/netmix2-license.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.workflow_run.head_sha }}
+      - name: Write Gurobi license
+        env:
+          GUROBI_LICENSE: ${{ secrets.GUROBI_LICENSE }}
+        run: |
+          mkdir -p licenses
+          printf '%s\n' "$GUROBI_LICENSE" > licenses/gurobi.lic
```
agitter, line 32: Can this leak the secret?

Issue #78, opened 2023-04-20
Omics Integrator 2 requires internet connection


agitter:
Omics Integrator 2 unexpected failed when I was running tests locally. It was because it requires an internet connection. We would have to modify the source code to change this behavior. For now, this issue serves to document the behavior, and I don't intend to change it.

The error was:
```
[...]

PR #226, opened 2025-05-26
fix: make omicsintegrator2 work offline

tristan-f-r:
Forces OI2 to work offline by disabling the unused networkx graph as interactive HTML feature (we already have workflows in SPRAS that enable this).

The network error was coming from `output_networkx_graph_as_interactive_html` dynamically fetching jQuery to insert it as an inline script. This could have also been fixed by providing a `local` folder with all of the scripts, but putting the several hundred-kilobyte scripts on this repository did not seem worth it for an unused feature.

This is tested - this also introduces a (docker-only) `network_disabled` parameter for running containers with a detached network.

Closes https://github.com/Reed-CompBio/spras/issues/78.

This is set as `P-medium` as this PR sets the precedent for patch files for both any regular patching, _and_ (more importantly) seeding algorithms.


spras/omicsintegrator2.py
```diff
@@ -125,7 +125,8 @@ def run(edges=None, prizes=None, output_file=None, w=None, b=None, g=None, noise
                             container_suffix,
                             command,
                             volumes,
-                            work_dir)
+                            work_dir,
+                            network_disabled=True)
```
ntalluri, line 129: The `network_disabled` parameter should be added into the config.yaml file and the run function arguments as a parameter a user can set.

ntalluri: Does this parameter apply to all algorithms or only oi2?

I'm assuming all, but for the scope of this PR I would keep it as an oi2 only parameter that can be used and set in the config file. If needed, a new PR can be created to enable this for all docker containers.

docker-wrappers/OmicsIntegrator2/Dockerfile
```diff
@@ -2,5 +2,14 @@
 # https://github.com/fraenkel-lab/OmicsIntegrator2
 FROM continuumio/miniconda3:4.9.2
 
+COPY 0001-disable-graph-exports.patch .
```
agitter, line 5: We'll have to make the usual updates elsewhere in SPRAS that accompany a Docker image bump: image version, note in the readme, etc.

docker-wrappers/OmicsIntegrator2/Dockerfile
```diff
@@ -2,5 +2,14 @@
...
+
+RUN git clone https://github.com/agitter/OmicsIntegrator2 && \
+    cd OmicsIntegrator2 && \
+    git reset --hard 568f170eae388e42e923c478ac9f3308b487760b && \
+    git config user.email "email@example.com" && \
+    git config user.name "Non-existent User" && \
```
agitter, line 11: git has to set a user before applying a patch? We don't have a SPRAS bot set up, so I can't think of a better default email and user to use. This is all inside an image and not committed so it shouldn't matter, right?

docker-wrappers/OmicsIntegrator2/Dockerfile
```diff
@@ -2,5 +2,14 @@
 # https://github.com/fraenkel-lab/OmicsIntegrator2
 FROM continuumio/miniconda3:4.9.2
 
+COPY 0001-disable-graph-exports.patch .
+
+RUN git clone https://github.com/agitter/OmicsIntegrator2 && \
```
agitter, line 7: Applying a patch can be a useful pattern, so a comment explaining what is happening here will help future maintainers.

Alternatively, we could use the new docs in #367 to write a short section on how to create and apply a patch, which would be even better.

agitter (commented):
Now that OI2 is patched, what is the goal of `network_disabled`? Is it for SPRAS developers to test algorithms to see if they fail without a network?

docker-wrappers/OmicsIntegrator2/Dockerfile
tristan-f-r: (This can be made easier with #314!)

docker-wrappers/OmicsIntegrator2/Dockerfile
tristan-f-r: Indeed. You can also just apply patch files directly, but this seemed like an okay nice tradeoff to get the niceties of `git patch`.

agitter:
> Yes. network_disabled is a fail-fast runtime assertion: any algorithm that doesn't need internet access shouldn't have it.

That helps. So it is a decision made by the SPRAS contributor who is wrapping the algorithm.

docker-wrappers/OmicsIntegrator2/Dockerfile
agitter: I don't know how soon we'll get to #314 so we may need to make these related changes manually to get this merged.

agitter:
I suggest we merge this before the larger #314. That means sticking with the existing versioning updates, readme update, etc. and manual DockerHub image push.

tristan-f-r:
[I don't have permission to update DockerHub, so someone else will have to do it / reconfigure my permissions]

agitter:
I will and push OI2. It is currently on `v2` so I will bump to `v3`.

Can you please make the readme version note and `run` function version updates as in #389?

PR #227, opened 2025-05-26
feat: experimental podman support

### LINKED ISSUE #124: Run SPRAS with Podman instead of Docker?
I'm trying to get Docker installed on my compbio lab machines, which needs to have a rootless version of Docker due to user permissions in the lab. This might be a possible scenario for other people as well (wanting to run SPRAS but not having root access). 

Our IT dept suggested MacOS + [Podman](https://podman.io/), a command line tool that uses a library that doesn't depend on a daemon and doesn't require root privileges.  It appears that you can run existing Docker images using Podman since they are compatible with the OCI standard (see, for example, [this tutorial](https://developers.redhat.com/blog/2020/11/19/transitioning-from-docker-to-podman)). 

If this would be a useful thing to pursue, we can get Podman installed on our machines and test SPRAS. I imagine this might be similar to the singularity wrappers @agitter wrote, so it would potentially be a larger change (but the code structure seems amenable to it).

tristan-f-r:
no formal test cases yet.

Closes #124

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -5,9 +5,10 @@ hash_length: 7
 
 # Specify the container framework used by each PRM wrapper. Valid options include:
 # - docker (default if not specified)
+# - podman - rootless version of docker, which should be a simple swap - if any algorithm doesn't work with podman, that is a bug.
 # - singularity -- Also known as apptainer, useful in HPC/HTC environments where docker isn't allowed
 # - dsub -- experimental with limited support, used for running on Google Cloud
-container_framework: docker
+container_framework: podman
 
 # Only used if container_framework is set to singularity, this will unpack the singularity containers
 # to the local filesystem. This is useful when PRM containers need to run inside another container,
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,7 @@ dependencies:
   - adjusttext=0.7.3.1
   - bioconda::snakemake-minimal=8.17.0
   - docker-py=5.0
+  - podman=5.4.0
   - matplotlib=3.6
   - networkx=2.8
   - pandas=1.5
--- a/spras/containers.py
+++ b/spras/containers.py
@@ -6,6 +6,7 @@ from pathlib import Path, PurePath, PurePosixPath
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import docker
+import podman
 
 import spras.config as config
 from spras.util import hash_filename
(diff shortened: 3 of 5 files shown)


config/config.yaml
```diff
@@ -5,9 +5,10 @@ hash_length: 7
 
 # Specify the container framework used by each PRM wrapper. Valid options include:
 # - docker (default if not specified)
+# - podman - rootless version of docker, which should be a simple swap - if any algorithm doesn't work with podman, that is a bug.
```
ntalluri, line 8: is it a bug with SPRAS or with the algorithm itself?

config/config.yaml
```diff
@@ -5,9 +5,10 @@ hash_length: 7
...
 # - singularity -- Also known as apptainer, useful in HPC/HTC environments where docker isn't allowed
 # - dsub -- experimental with limited support, used for running on Google Cloud
-container_framework: docker
```
ntalluri, line 10: change back to docker

environment.yml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - adjusttext=0.7.3.1
   - bioconda::snakemake-minimal=8.17.0
   - docker-py=5.0
+  - podman=5.4.0
```
ntalluri, line 8: should this also be 5.4.0.1?

spras/containers.py
```diff
@@ -239,6 +242,40 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     # finally:
     return out
 
+# TODO any issue with creating a new client each time inside this function?
+def run_container_podman(container: str, command: List[str], volumes: List[Tuple[PurePath, PurePath]], working_dir: str, environment: str = 'SPRAS=True'):
```
ntalluri, line 246: To test the Podman functionality, a test case should be added to each algorithm test (similar to the singularity test cases).

environment.yml
tristan-f-r: I ran into an issue when I pinned it to 5.4.0.1 and ignored it - I could look more into it, though I'll circle back to this PR since I figured out what was blocking me in the NetMix2 integration.

spras/containers.py
```diff
@@ -239,6 +242,40 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
+            out = str(out, 'utf-8')
+
+            if code != 0:
+                print(f"Podman container {container} exited with non-zero exit code {code}.")
+        else:
+            raise RuntimeError("podman socket is not responding to pings.")
+
+    # Removed the finally block to address bugbear B012
+    # "`return` inside `finally` blocks cause exceptions to be silenced"
+    # finally:
```
jhiemstrawisc, line 276: Leftover from a copy paste

config/config.yaml
jhiemstrawisc: Which error domain would you consider this to be a bug in? Does it signify a bug in SPRAS, or does it mean the algorithm/wrapper container itself is incompatible with Podman?

spras/containers.py
```diff
@@ -239,6 +242,40 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     # finally:
     return out
 
+# TODO any issue with creating a new client each time inside this function?
```
jhiemstrawisc, line 245: Any insight on whether this comment affects the podman client? I'd imagine this approach means that workflow runs with high concurrency will spawn many clients that each need to establish a connection with the podman socket. It may be useful to adjust the comment to explain why we're worried about this.

config/config.yaml
tristan-f-r: It would mean the container itself is incompatible with Podman, but this likely means that on our side, we did some really hacky trick to get that container to work in the first place.

tristan-f-r:
[I left this PR stale - it's still under draft because I never got Podman to work inside my devcontainer environment]

Issue #33, opened 2021-05-21
Make config file parameter parsing more robust


agitter:
There are a few ways we could improve how parameters are parsed from the config 
- [x] Support singleton values so they do not have to be wrapped in a list (e.g. `3`)
- [x] Restrict when we call `eval` and use pattern-matching to support only a few predetermined functions like `range`, `np.linspace`, `np.arange`, `np.logspace`
- [ ] Support running an algorithm multiple times with the same parameters, which should be a new special field instead of a parameter. This would be used to run a non-deterministic algorithm multiple times.

A useful way to check for and handle singletons could be
```python
from collections.abc import Iterable

if not isinstance(param, Iterable):
```

tristan-f-r:
Also noticing this (following the contributing guide, made a small err in my `config.yml`) - perhaps there is a nice validation library in python that allows us to parse with a schema first, so the config file parsing has nicer error messages that link to where the user got the configuration wrong?

(I got an indexing by string instead of number error [related to parsing `runN`] because I included a bad key, `owner`, thinking that I could specify the docker hub owner per-algorithm instead of overall.)

tristan-f-r:
(I'm going to actually close this in favor of #174 since #228 was merged.)

PR #228, opened 2025-05-26
fix: tighten & relax yaml parsing constraints

tristan-f-r:
This checks for object prefixes before using `eval`. Instead, the parsing goes as:
1. Is the param an int?
1. Is the param a float?
1. Is the param a bool?
1. Is the param one of our desired evaluation objects? (If this causes too many problems, we can probably just flag `np.` to always be allowed.)
1. Parse as a string inside a singleton array.

This does make config parsing perhaps more rigid, but this has the side bonus of allowing singleton values.

(I chose not to prefer `eval` here, as I do not want to see how the python `eval` system and the [already broad](https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell) YAML format interact with each other. I still do an extra `iterable` check so the "can't index X by string" error appears without a recognizable stacktrace.)

This addresses two out of the three wants from https://github.com/Reed-CompBio/spras/issues/33.


ntalluri:
Could you add test cases in test/config.py for this updated change?

tristan-f-r:
Thanks! I forgot we had a dedicated test file for configs.

test/test_config.py
```diff
@@ -21,7 +21,23 @@ def get_test_config():
...
+                "params": {
+                    "include": True,
+                    "run1": {"test": 1, "test2": [2, 3]}
+                }
+            },
+            {
+                "name": "objs",
+                "params": {
+                    "include": True,
+                    "run1": {"test": "str1", "test2": "np.linspace(0,5,2)"}
```
ntalluri, line 37: can you also test for list of strings?

test/test_config.py
```diff
@@ -21,7 +21,30 @@ def get_test_config():
         },
         "datasets": [{"label": "alg1"}, {"label": "alg2"}],
         "gold_standards": [{"label": "gs1", "dataset_labels": []}],
-        "algorithms": [{"params": ["param2", "param2"]}],
+        "algorithms": [
+            {"params": ["param2", "param2"]},
+            {
+                "name": "singleton",
```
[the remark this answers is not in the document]
tristan-f-r: Not a fan of not mixing up string and non-string types in the `obj` test - I agree that they should be reorganized, though. I'll find a layout :+1:

tristan-f-r:
Adding more tests for some more code coverage.

ntalluri:
Note: test egfr file

spras/config.py
```diff
@@ -222,16 +223,36 @@ def process_config(self, raw_config):
...
                     for p in cur_params[run_params]:
                         param_name_list.append(p)
-                        all_runs.append(eval(str(cur_params[run_params][p])))
+                        obj = str(cur_params[run_params][p])
+                        try:
+                            obj = [int(obj)]
+                        except ValueError:
+                            try:
+                                obj = [float(obj)]
+                            except ValueError:
```
agitter, line 232: Should we also check for `bool` along with `int` and `float` before moving to these ranges?

spras/config.py
```diff
@@ -222,16 +223,36 @@ def process_config(self, raw_config):
...
+                                if obj.startswith(("range", "np.linspace", "np.arange", "np.logspace", "[")):
+                                    obj = eval(obj)
+                                else:
+                                    obj = [obj]
+                            if not isinstance(obj, Iterable):
+                                raise ValueError(f"The object `{obj}` in algorithm {alg['name']} at key '{p}' in run '{run_params}' is not iterable!") from None
```
agitter, line 238: This behavior seems inconsistent with the numeric parameter behavior that is now supported. We can have a int or float that is not wrapped in a list, but a string still has to be wrapped in a list. Users won't understand that.

Should the case here treat the parameter as a single string? Are there downsides I'm missing?

spras/util.py
```diff
@@ -22,7 +34,7 @@ def hash_params_sha1_base32(params_dict: Dict[str, Any], length: Optional[int] =
     @param length: the length of the returned hash, which is ignored if it is None, < 1, or > the full hash length
     """
     params_hash = hashlib.sha1()
-    params_encoded = json.dumps(params_dict, sort_keys=True).encode()
+    params_encoded = json.dumps(params_dict, sort_keys=True, cls=cls).encode()
```
agitter, line 37: We should make a SPRAS release after merging this. I expect the consequence is that the sha1 hashes of parameter combinations will change, breaking output caching. That is, if a user ran PathLinker with k=5 before and they run it again with k=5 after merging this, the results would be regenerated with a new parameter hash.

spras/util.py
```diff
@@ -8,10 +8,22 @@
 from pathlib import Path, PurePath, PurePosixPath
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+import numpy as np
 import pandas as pd
 
 
-def hash_params_sha1_base32(params_dict: Dict[str, Any], length: Optional[int] = None) -> str:
+# https://stackoverflow.com/a/57915246/7589775
+class NpEncoder(json.JSONEncoder):
```
agitter, line 16: Will need to extend this for bool if you accept my other suggestion above.

test/test_config.py
```diff
@@ -21,7 +24,51 @@ def get_test_config():
...
         "datasets": [{"label": "alg1"}, {"label": "alg2"}],
         "gold_standards": [{"label": "gs1", "dataset_labels": []}],
-        "algorithms": [{"params": ["param2", "param2"]}],
+        "algorithms": [
+            {"params": ["param2", "param2"]},
+            {
+                "name": "strings",
+                "params": {
+                    "include": True,
+                    "run1": {"test": "str1", "test2": ["str2", "str3"]}
```
agitter, line 33: I must be misreading something. I thought I single string was not currently allowed, but it works in the test here.

test/test_config.py
```diff
@@ -21,7 +24,51 @@ def get_test_config():
...
+            },
+            {
+                "name": "str_array_np_logspace",
+                "params": {
+                    "include": True,
+                    "run1": {"test": ["a", "b"], "test2": "np.logspace(1,1)"}
+                }
+            },
+            {
+                "name": "int64artifact",
```
agitter, line 65: Is the Python `range` tested?

test/test_config.py
```diff
@@ -45,6 +92,20 @@ def get_test_config():
...
 
+def value_test_util(name: str, configurations: list):
+    assert name in config.config.algorithm_params, f"{name} isn't a present algorithm configuration!"
+
+    keys = config.config.algorithm_params[name]
+    values = [config.config.algorithm_params[name][key] for key in keys]
+
+    # https://stackoverflow.com/a/50486270/7589775
+    set_values = set(tuple(sorted(d.items())) for d in sorted(values, key=lambda x: pickle.dumps(x, protocol=3)))
+    set_configurations = set(tuple(sorted(d.items())) for d in sorted(configurations, key=lambda x: pickle.dumps(x, protocol=3)))
```
agitter, line 103: Why is pickle needed here?

agitter (commented):
This is a much needed update.

test/test_config.py
tristan-f-r: The `obj = [obj]` is a catch-all for non-evaluated objects - it handles strings. I've just documented that behavior :+1:

spras/util.py
tristan-f-r: I've accepted the above `bool` suggestion, which was a good catch, but this specifically relates to numpy encoding. I have added a comment that explains this.

tristan-f-r:
[Apologies for the confusing code - it seems that I thought of them as too natural, so this review was very much needed - explanations have been added to the code :+1:]

spras/config.py
```diff
@@ -222,16 +223,43 @@ def process_config(self, raw_config):
...
+                                    obj = eval(obj)
+                                elif obj.lower() == "true":
+                                    obj = [True]
+                                elif obj.lower() == "false":
+                                    obj = [False]
+                                else:
+                                    # Catch-all for strings
+                                    obj = [obj]
+                            if not isinstance(obj, Iterable):
+                                raise ValueError(f"The object `{obj}` in algorithm {alg['name']} at key '{p}' in run '{run_params}' is not iterable!") from None
```
tristan-f-r, line 245: I'm not a fan of the massive chain of try-catch statements - is there some parsing method that allows me to safely parse `int`s and `float`s with `Optional`s instead of this?

config/config.yaml
```diff
@@ -51,57 +51,50 @@ algorithms:
...
+      include: true
+      run1:
+        flow: 1 # The flow must be an int
+        capacity: 1
+
+  - name: "allpairs"
+    params:
+      include: true
 
   - name: "domino"
```
agitter, line 92: The DOMINO algorithm was dropped? Was that an accident?

spras/config.py
```diff
@@ -230,9 +230,16 @@ def process_config(self, raw_config):
                             try:
                                 obj = [float(obj)]
                             except ValueError:
+                                # Handles arrays and special evaluation types
+                                # TODO: do we want to explicitly bar `eval` if we may use untrusted user inputs later?
```
agitter, line 234: Disallowing `eval` seems like a good idea now that you have this more flexible parsing in place.

spras/util.py
```diff
@@ -13,15 +13,25 @@
 
 
 # https://stackoverflow.com/a/57915246/7589775
-class NpEncoder(json.JSONEncoder):
+# numpy variables are not, by default, encodable by python's JSONEncoder.
+# Thus, we need to wrap the encoder to reduce np-objects down to regular floats and ints.
+# To preserve precision, we stringify the objects instead,
+# which is okay, as this is specifically for hashing.
+# TODO: this can still have a hashing conflict if someone uses `np.integer` for one paramater combination,
```
agitter, line 20:
```suggestion
# TODO: this can still have a hashing conflict if someone uses `np.integer` for one parameter combination,
```

Would the conflict come if something does, for instance:
```
k: [5, "5"]
```

That would be an intentionally unusual corner case, so I'm okay leaving the comment but removing the TODO.

agitter (commented):
All of my major comments have been addressed. If someone else wants to do the final review and approval while I am unavailable, go for it.

test/test_config.py
agitter: The commend about dicts explained things.

spras/util.py
ntalluri: There are some cases for packages that allow Ints and strings for input. Most recently I saw it for the KDE Scikit learn. It’s not a pra, but this case could happen.

agitter: Can you please give an example of the arguments it could take to help illustrate?

config/config.yaml
tristan-f-r: [I'm seeing it on my fork](https://github.com/Reed-CompBio/spras/blob/7ccae7a567255942ceb029a5daf30efb19047f1e/config/config.yaml#L92-L97) - are we desynced in some way? I seem to be up to date.

test/test_config.py
```diff
@@ -145,6 +215,22 @@ def test_error_gs_dataset_mismatch(self):
         with pytest.raises(ValueError):
             config.init_global(test_config)
 
+    def test_config_values(self):
+        test_config = get_test_config()
```
ntalluri, line 219: Are lines 219-220 needed if the function value_test_util is being used?

tristan-f-r: This test does work without `get_test_config`, but I suspect that that is because proper global config cleanup isn't done properly in this entire test suite - `value_test_util` doesn't do any `get_test_config` setup.

spras/config.py
ntalluri: @tristan-f-r  is this something you are planning on doing? Otherwise I was going to approve this PR.

ntalluri, line 245: I was trying to get this error called, but I can't find a combination that would cause this to happen. Do you have a test case where this error is called?

tristan-f-r: This branch is essentially never supposed to get called - any case that a user should reasonably type in should have already been caught by the previous branches. However, it is still possible to get this to be called. For example, try using `range` as a value: since `range` is in the allowlist of objects to be evaluated [for now], `range` is evaluated, but the output of `range` is a function, not an iterable object.

Ideally, we'll have complete parsing support later on that gets rid of `eval` entirely, which we can either handle in this PR or a secondary follow-up one - then, in that world, this case should (ideally) never happen.

ntalluri: Ok that makes more sense. Would you be able to add that as a test case in test_config.py?

tristan-f-r: I could, but I don't think it would be much help as a test case - eventually, that behavior won't be desired.

PR #239, opened 2025-05-28
style: yamlfmt

### LINKED ISSUE #48: Lint YAML config file
Our example config file has a mixture of different spacing and a single space in front of every row.  We should run it through a linter and standardize the whitespace.

tristan-f-r:
Closes #48. this uses the yamlfmt default config, which uses 2 spaces instead of four - a convention that seems to be recommended.

[If any style opinion that deviates from yamlfmt's default is preferred I will immediately add it. I do not want bikeshedding.]


.pre-commit-config.yaml
```diff
@@ -24,3 +24,7 @@ repos:
     rev: 'v0.0.269'
     hooks:
       - id: ruff
+  - repo: https://github.com/google/yamlfmt
+    rev: v0.14.0
```
agitter, line 28: Is there a reason to use v0.14.0 instead of [v0.17.0](https://github.com/google/yamlfmt/releases)?

agitter (commented):
I only looked at the `.pre-commit-config.yaml` changes so far and like the idea of adding yamlfmt. One of us should manually review the automatically updates applied to the .yml/.yaml files before merging to make sure all of them look okay.

.pre-commit-config.yaml
tristan-f-r: Nope! That was taken from the pre-commit docs from yamlfmt. I've just sent a PR to them just to update those docs :+1:

tristan-f-r:
Since reviewing all of these YAML files is quite a tedious task, I've removed `easy` from this PR.

tristan-f-r:
Restricted `yamlfmt` to allow single newlines.

agitter (approved):

agitter:
Now that this initially formatting has been merged, if you want to make a small follow up that sets `retain_line_breaks_single: false` that would be easier to review instead of doing both at once.

PR #241, opened 2025-05-29
Responsenet recreate

tristan-f-r:
Finishes #176. [I couldn't push to Gabe's branch - sorry!]


tristan-f-r:
Accidentally undrafted this - I need to add test cases for generate inputs and process outputs, but I'm unable to replicate the core issue that was being experienced in #176 - the current changes do actually make the tests related to ResponseNet succeed.

ntalluri:
(Copying information from directionality PR that explains why ResponseNet outputs are undirected explained by @agitter)

The original paper for ResponseNet on biological networks states that the direction of flow is not meaningful in the output networks. Therefore we treat all the outputs of ResponseNet as undirected.

The information is in the supplement https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2733244/bin/NIHMS109939-supplement-supp.pdf and the section is "Directionality of ResponseNet output". Please add a reference to this in the comments to explain why we treat the output as undirected. It states:
> 
> The flow algorithm we employ provides a directed network. However the directionality
> of the interactions in the network is determined by the fact that we have connected all
> genetic hits to the source of flow and all differentially expressed genes to the flow sink.
> Therefore, except for the interactions between transcription factors and their targets, the
> flow does not necessarily reflect a causal order of events

tristan-f-r:
Thanks! I got a little confused since the original PR (#176) marked all the output edges as directed, despite leaving an explanation why they should be undirected.

config/config.yaml
```diff
@@ -45,6 +45,15 @@ container_registry:
 # careful: too many parameters might make your runs take a long time.
 
 algorithms:
+      - name: "responsenet"
+        params:
+              include: true
+              run1:
+                  gamma: [10]
+                  _include_st: [false]
```
annaritz, line 53: remove developer params from config file.

spras/responsenet.py
```diff
@@ -0,0 +1,159 @@
...
+        # responsenet should be recieving a directed graph
+        edges = data.get_interactome()
+        edges = convert_undirected_to_directed(edges)
+
+        # creates the edges files that contains the head and tail nodes and the weights after them
+        edges.to_csv(filename_map['edges'], sep='\t', index=False, columns=["Interactor1", "Interactor2", "Weight"],
+                     header=False)
+
+    @staticmethod
+    def run(sources=None, targets=None, edges=None, output_file=None, gamma=10, _include_st=False, _verbose=False ,_output_log=False, container_framework="docker"):
```
annaritz, line 61: Remove developer variables here too.

spras/responsenet.py
```diff
@@ -0,0 +1,159 @@
...
+        # Makes the Python command to run within in the container
+        command = ['python',
+                    'responsenet.py',
+                    '--edges_file', edges_file,
+                    '--sources_file', sources_file,
+                    '--targets_file', targets_file,
+                    '--output', str(mapped_out_prefix / 'output'),
+                    '--gamma', str(gamma)]
+
+        # add optional flags, value can be changed in config/config.yaml
```
annaritz, line 109: REmove optional flags.

annaritz (commented):
We talked as a group and decided to remove the optional flags.

Issue #246, opened 2025-05-30
Understand the licensing implications of different containerization strategies


agitter:
If some of the wrapped pathway reconstruction algorithms have restrictive licenses, we need to carefully think through our options for preparing and distributing containers. This Linux Foundation [whitepaper](https://www.linuxfoundation.org/resources/publications/docker-containers-what-are-the-open-source-licensing-considerations) provides an initial resource to consult.

tristan-f-r:
From slack, it seems that we generally decided that the least permissive licenses that we can use are in the GPL family (AGPL, LGPL, GPL). Issues that have more restrictive licenses (e.g. non-commercial licenses) are flagged with the `license-conflict` label on GitHub.

Issue #109, opened 2023-08-07
Add ROBUST


agitter:
Robust disease module mining via enumeration of diverse prize-collecting Steiner trees
https://doi.org/10.1093/bioinformatics/btab876

Original Python script
https://github.com/bionetslab/robust

Updated Python script
https://github.com/bionetslab/robust_bias_aware

ntalluri:
Maybe instead of making our own, could we potentially use the dockerized ROBUST method from https://academic.oup.com/bioinformatics/article/42/Supplement_1/btag223/8726292

agitter:
Yes, that would be a great first step for #507

Issue #88, opened 2023-06-07
Standardized Nodes Output File?


annaritz:
As @Lyce24 implements a random walk with restarts algorithm, it occurs to us that there may be algorithms that contain useful information for the nodes (in RWR's case, the node visitation probability). Right now, SPRAS standardizes a pathway as an edge list file. Some options we could discuss:
- Essentially ignore node information in the SPRAS outputs. Intermediate files may contain this information for different algorithms.
- Add a standardized node file, which is written if there is useful node information. Would often require developers to do a bit of extra work when there's no useful node information.
- Include node information as extra columns in the `pathway.txt` standardized output file. The information would be redundant - the node info would be written for EVERY edge that the node is incident on - but it would keep the current file format for pathways.

agitter:
The standardized node file seems like the right choice. We could create a helpful util function for the algorithms that don't generate node information that takes in the pathway file and automatically writes a default node output file so that developers don't have to think about it too much. Like `write_default_node_scores("pathway.txt", "nodes.txt")`.

Before we proceed, can we outline what other methods (existing or on the SPRAS roadmap) also generate node information so we can think through how it can be used? And how do we envision using node information in downstream tasks? For instance, it would be great to load into our Cytoscape visualizations. Would we load it as a generic "score", or do we need to also track a label saying this score is a node visitation probability.

PR #247, opened 2025-05-30
feat: DIAMOnD

### LINKED ISSUE #244: Add DIAMOnD
A DIseAse MOdule Detection (DIAMOnD) Algorithm Derived from a Systematic Analysis of Connectivity Patterns of Disease Proteins in the Human Interactome
https://doi.org/10.1371/journal.pcbi.1004120 

https://github.com/barabasilab/DIAMOnD

While it can't take into account scores, it was used as a benchmarking reference for ROBUST (#109)

tristan-f-r:
Another disease module mining algorithm to complement #240 (#109). Closes #244.

Note that, in compliance with #246, [DIAMOnD is MIT licensed](https://github.com/dinaghiassian/DIAMOnD/blob/master/LICENSE).

### DIFF
--- a/spras/diamond.py
+++ b/spras/diamond.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from pydantic import BaseModel, ConfigDict
 
 from spras.config.container_schema import ProcessedContainerSettings
-from spras.containers import prepare_volume, run_container_and_log
+from spras.containers import ContainerError, prepare_volume, run_container_and_log
 from spras.dataset import Dataset
 from spras.interactome import (
     convert_directed_to_undirected,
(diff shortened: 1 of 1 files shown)


spras/diamond.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @staticmethod
+    def parse_output(raw_pathway_file, standardized_pathway_file, original_dataset: Dataset):
+        df = raw_pathway_df(raw_pathway_file, sep='\t', header=0)
+        if not df.empty:
+            # preprocessing - drop [useful] p_hyper information, and rename columns to Rank and Node
+            df = df.drop(columns=["p_hyper"])
+            df.columns = ["Rank", "Node"]
+
+            # TODO: we throw away rank information - we want this.
```
tristan-f-r, line 105: Actually, how do we want to handle this? Take the edges with vertex rankings 1-4 and 2-3. Should we have this as 'decorative vertex data' (suggested in #88) or do we want some user input about what ranking algorithm to use on edges given vertex rankings?

tristan-f-r: Perhaps, since we do have `Rank 1` constant ranks, we can have these ranks to be the same (edge rank = vertex 1 rank + vertex 2 rank), and add the original vertex ranks as decorative vertex data from #88.

tristan-f-r:
This will be updated after #261 gets merged.

agitter:
Whose help is needed to support DIAMOnD integration? @nisthapanda noticed that we are working on this because she may be able to use it for her project on genetic variant interpretation.

tristan-f-r:
@agitter this only needs further review on some pandas calls in `parse_output` and its correctness, but the rest of this is a straightforward PRM implementation from what I remember. (It turns node rankings into edge rankings.)

annaritz:
Still a little unclear how seed weights are incorporated (should they be prizes, then?) 

Also, do you know why tests are failing?

tristan-f-r:
There is an actual CI failure! I am not sure exactly, but I believe DIAMOnD fails reconstruction if the provided `n` is one less than the network size. The last two commits address this and wrap the error with a nice display message wrapping the original opaque error.

annaritz (approved):
Ok, after thinking about the alpha parameter I get it now (it's something _passed_ to DIAMOnD, not something that requires input file modification). That's where my confusion was - the code and the comments are all correct.

All tests pass with the `linux/amd64` images! I'm approving for merging.

PR #204, opened 2025-02-07
Wrapped BTB algorithm for SPRAS integration

oliverfanderson:
Finish wrapping BowTieBuilder pathway reconstruction algorithm into SPRAS environment.

BTB connects multiple source nodes to multiple target nodes in a "BowTie" shape. It can consider weighted edges and works on both directed and undirected graphs.

PR #168, opened 2024-07-03
BowTieBuilder Code Review

gabeah:

PR #150, opened 2024-02-29
Draft: BowtieBuilder status

ctrlaltaf:
Things to do:
- Work on `generate_inputs`, `run`, and `parse_output` methods in `btb.py`
- Create tests for btb

PR #250, opened 2025-06-03
feat: BTB

tristan-f-r:
supersedes #150, #168, and #204 - [link to algorithm](https://github.com/Reed-CompBio/BowTieBuilder-Algorithm)


.github/workflows/test-spras.yml
annaritz, line 1: Remove this.

spras/btb.py
```diff
@@ -0,0 +1,163 @@
...
+        # Tests for pytest (docker container also runs this)
+        # Testing out here avoids the trouble that container errors provide
+
+        if not sources or not targets or not edges or not output_file:
+            raise ValueError('Required BowTieBuilder arguments are missing')
+
+        if not Path(sources).exists() or not Path(targets).exists() or not Path(edges).exists():
+            raise ValueError('Missing input file')
+
+        # Testing for btb index errors
```
annaritz, line 91: This might not be relevant anymore (if it is, it should be in `prepare_inputs()`). Try commenting out and checking.

spras/btb.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Create network file
+        edges = data.get_interactome()
+
+        # Format into directed graph (BTB uses the nx.DiGraph constructor internally)
+        edges = convert_undirected_to_directed(edges)
+
+        edges.to_csv(filename_map["edges"], sep="\t", index=False,
+                                      columns=["Interactor1", "Interactor2", "Weight"],
+                                      header=False)
```
annaritz, line 66: This should be where the "btb index check" code might go.

annaritz (commented):
Some minor things to check; a few larger issues raised but we'll make new PRs/Issues for them.

spras/btb.py
annaritz: Did anyone check if this check (for btb index errors) needs to be here, or whether it should be in `prepare_inputs()`? Does it matter?

annaritz:
RWR is now merged, we can make the modifications here and wrap this up soon.

spras/btb.py
annaritz: I talked through this with @tristan-f-r - let's note this issue in error handling (that this test isn't checking something that actually will ever happen) and keep the code in for now so that test passes for BTB until we make a decision at a later time.

PR #254, opened 2025-06-03
fix: remove support for indicator files

### LINKED ISSUE #192: Remove support for secondary node that uses the filename as the node attribute
We currently support two node file formats (https://github.com/Reed-CompBio/spras/tree/master/input#node-file)
- A node table
- A file listing node IDs that uses the filename as the node attribute

The second format causes more problems than benefits. It may be time to drop support.

tristan-f-r:
these used the filenames as the column names, which can cause headaches (e.g. case sensitivity, sources (1), ...)

Closes #192.

### DIFF
--- a/input/sources.txt
+++ b/input/sources.txt
@@ -1,1 +1,2 @@
-A
+NODEID	sources
+A	True
--- a/spras/dataset.py
+++ b/spras/dataset.py
@@ -107,14 +107,9 @@ class Dataset:
         self.node_table = pd.DataFrame(node_set, columns=[self.NODE_ID])
         for node_file in node_data_files:
             single_node_table = pd.read_table(os.path.join(data_loc, node_file))
-            # If we have only 1 column, assume this is an indicator variable
+            # If we have only 1 column, mention that indicator files are deprecated.
             if len(single_node_table.columns) == 1:
-                single_node_table = pd.read_table(
-                    os.path.join(data_loc, node_file), header=None
-                )
-                single_node_table.columns = [self.NODE_ID]
-                new_col_name = node_file.split(".")[0]
-                single_node_table[new_col_name] = True
+                raise ValueError("Indicator files are deprecated. See: https://github.com/Reed-CompBio/spras/issues/192.")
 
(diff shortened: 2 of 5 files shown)


input/sources.txt
ntalluri, line 1: Can you add in two more files called alternate_sources.txt which can be put in the config file data1. And for the sources and targets file, keep it for data0. This will make sure that we can have different names for this file.

spras/dataset.py
ntalluri, line 1: Can you update the input/README.md with this new update.

input/README.md
ntalluri, line 1: The node file also needs to be updated. I don’t think this code changes the way the gold standard is processed. That is done in evaluation.py.

tristan-f-r:
Since tests are likely desired for this, #255 should be merged first as it lays the ground for dataset tests.

input/alternative-sources.txt
ntalluri, line 1: why is this the same source as sources.txt

config/config.yaml
```diff
@@ -116,7 +116,7 @@ datasets:
     -
       label: data1
       # Reuse some of the same sources file as 'data0' but different network and targets
-      node_files: ["node-prizes.txt", "sources.txt", "alternative-targets.txt"]
+      node_files: ["node-prizes.txt", "alternative-sources.txt", "alternative-targets.txt"]
```
ntalluri, line 119: where is alternative-targets.txt?

input/alternative-sources.txt
agitter: I also like the secondary format with a header row followed by a list of nodes. I find that simple to prepare for input files that list sources, targets, etc.

I'm not pushing back on supporting that file format, I only want to make sure we all agree on what input format(s) should be supported after these updates. @ntalluri do you have a preference?

config/config.yaml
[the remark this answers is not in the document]
tristan-f-r: I believe that makes more sense for algorithms than it does datasets: nothing actually changes if you change the format of how the datasets are specified.

PR #255, opened 2025-06-03
fix: check if edge weights are between 0 and 1

### LINKED ISSUE #237: Check Prize/Score Values Between 0 and 1
When using experimental scores for the synthetic dataset, we initially forgot to normalize prize values between 0 and 1. This oversight caused confusion and downstream failures in SPRAS, because the scores were between 0 and 1000. To prevent this issue and confusion, we should add a validation step that checks whether all prize values fall within the [0, 1] range. If any values fall outside this range, the workflow should halt and return a clear error message indicating the actual minimum and maximum prize values detected (e.g., “Error: Prize values must be in [0, 1]. Detected range: [x, y]”).

tristan-f-r:
Closes #237.


tristan-f-r:
Actually, I do want to print out what the value is (don't want anyone to have to try to write a script or perhaps even a regex to try to find what the outlier value is)

tristan-f-r:
[Do we have an established way of testing dataset parsing? A quick look says there isn't any, but it could be useful to add for e.g. #254]

ntalluri:
This would be a good chance to add in a test for dataset parsing.

tristan-f-r:
That sounds good - @ntalluri should I merge this PR and #254 to add some encapsulating dataset test file for the two?

ntalluri:
I would keep them separate for now as they target different things. We can plan to merge this pr before 254. I would add the test case here first then once merged, pull it into 254.

spras/dataset.py
```diff
@@ -11,13 +12,20 @@
 Methods and intermediate state for loading data and putting it into pandas tables for use by pathway reconstruction algorithms.
 """
 
+class DatasetDict(TypedDict):
```
ntalluri, line 15: Is this supposed to be a class within a class situation?

ntalluri, line 15: Could you also add a docstring comment about this

test/dataset/fixtures/empty/network.txt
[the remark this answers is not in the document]
ntalluri: I think it might be fine adding only the empty network test case, and then make a separate pr for fixing the empty other files (sources/targets/node-prizes) (Only because we should test to see what caused sumedha's error).

spras/dataset.py
```diff
@@ -89,6 +97,13 @@ def load_files_from_dict(self, dataset_dict):
                 "Direction",
             ]
 
+            weight_series = pd.Series(self.interactome.Weight)
+            weight_series_out_range = ~weight_series.between(0, 1)
```
ntalluri, line 101: @agitter @annaritz Since a weight could be negative, should we allow that as well?

test/dataset/fixtures/not-in-range/node-prizes.txt
ntalluri, line 1: This is more of a reminder, but when the indicator files get depreciated, these node-prizes files should be updated to have sources and target columns.

test/dataset/fixtures/empty/nodes.txt
ntalluri, line 1: should there also be a test case for just column headers (no body).

ntalluri (commented):
Here is a first pass of reviews

spras/dataset.py
tristan-f-r: Forgot to respond to this - wasn't there an issue open that established they can't be negative? #237

ntalluri: In general, some graph algorithms do allow for negative weights. I forgot about this in #237.

ntalluri: Yeah, let's ignore negative weights for now. If needed, we can add support for them in the #302 PR or address it in a separate PR after #302 merges, since that one already introduces a lot of new code. I suspect we won’t need to handle negative weights for the benchmarking datasets (and I'm assuming the people that use SPRAS right probably don't deal with negative weights), so in my ming it’s low priority. However, to make sure we don’t lose track of it, we should create a separate issue to revisit this later in case it becomes necessary.

tristan-f-r:
Closing in favor of a dataset test PR. See #322 for more info. (Actual weight assumptions can be moved after #302 is merged)

PR #260, opened 2025-06-09
Profiling

jhiemstrawisc:


spras/containers.py
```diff
@@ -240,14 +242,98 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
+    except Exception as e:
+        print(f"Failed to read cpu.stat from cgroup: {e}")
+
+    # Set up the header for the TSV file
+    header = ["peak_memory_bytes", "cpu_usage_usec", "cpu_user_usec", "cpu_system_usec"]
+    row = [peak_mem, cpu_usage, cpu_user, cpu_system]
+
+    # Write the contents of the file
+    write_header = not os.path.exists(profile_path) or os.path.getsize(profile_path) == 0
+    with open(profile_path, "a", newline="") as out_f:
```
tristan-f-r, line 319: A test on apptainer for testing for the existence of `profile_path` would be good.

spras/containers.py
```diff
@@ -240,14 +242,98 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     return out
 
 
-def run_container_singularity(container: str, command: List[str], volumes: List[Tuple[PurePath, PurePath]], working_dir: str, environment: str = 'SPRAS=True'):
+def create_cgroup() -> str:
```
tristan-f-r, line 245: [I'm generally made uncomfortable with including such a specific function in this ever-growing file, and I would like to see profiling in general moved to a separate file (especially if we decided to add profiling for other platforms/container frameworks). Understandably, though, that is a little much.]

spras/containers.py
```diff
@@ -271,38 +357,54 @@ def run_container_singularity(container: str, command: List[str], volumes: List[
     else:
         singularity_options.extend(['--env', environment])
 
-    # Handle unpacking singularity image if needed. Potentially needed for running nested unprivileged containers
+    expanded_image = None
     if config.config.unpack_singularity:
-        # Split the string by "/"
         path_elements = container.split("/")
-
-        # Get the last element, which will indicate the base container name
```
tristan-f-r, line 279: I do quite miss this comment, as `base_cont` wasn't descriptive. [It could be worthwhile to specify why the last element is the base container name]

spras/containers.py
```diff
@@ -271,38 +357,54 @@ def run_container_singularity(container: str, command: List[str], volumes: List[
...
         base_cont = path_elements[-1]
         base_cont = base_cont.replace(":", "_").split(":")[0]
         sif_file = base_cont + ".sif"
-
-        # Adding 'docker://' to the container indicates this is a Docker image Singularity must convert
         image_path = Client.pull('docker://' + container, name=sif_file)
-
-        # Check if the directory for base_cont already exists. When running concurrent jobs, it's possible
-        # something else has already pulled/unpacked the container.
-        # Here, we expand the sif image from `image_path` to a directory indicated by `base_cont`
```
tristan-f-r, line 289: (and these comments)

tristan-f-r (commented):
Looks good! Almost every comment is a nitpick, except for the test comment - something to verify (not for correctness) that profiling doesn't suddenly error would be reassuring.

tristan-f-r:
[Apologies for the force push - I messed up and forgot to type the `Path` usage, so that most recent commit uses the Path relativalized `out_dir` and `working_dir` to absolute to resolve the merge conflict. (While the libraries still worked without this change before Profiling, Profiling forced serialization of Path, which can't happen). The force push was because my commit history was getting too long.]

jhiemstrawisc:
One additional note when it comes time to verify this unit of work in the CHTC pool -- the tester should add the following constraint to their submit file (if running everything on one EP):
```
requirements = versionGE(split(Target.CondorVersion)[1], "24.8.0") && (isenforcingdiskusage =!= true)
```

or this line under the `default-resources` section of their snakemake profile (if splitting work across EPs):
```
requirements: "'versionGE(split(Target.CondorVersion)[1], \"24.8.0\") && (isenforcingdiskusage =!= true)'"
```

This pins jobs to execution points that a) run the minimal required version of condor and b) don't enable another feature (disk enforcement) that doesn't play well with what profiling does to cgroups.

tristan-f-r:
A question out of curiosity: what exactly happens if `isenforcingdiskusage` is set to true? [All I get from the readthedocs is](https://htcondor.readthedocs.io/en/latest/classad-attributes/machine-classad-attributes.html):

> A boolean value that when True identifies that the machine is setup to enforce disk usage limits for each job the machine executes.

How does the created cgroup sibling mess with the way that this `isenforcingdiskusage` flag works? My initial assumption was that `isenforcingdiskusage` is somehow using a parent cgroup to monitor disk usage, but I don't remember `cgroup` ever being capable of that.

jhiemstrawisc:
> A question out of curiosity: what exactly happens if `isenforcingdiskusage` is set to true? [All I get from the readthedocs is](https://htcondor.readthedocs.io/en/latest/classad-attributes/machine-classad-attributes.html):
> 
> > A boolean value that when True identifies that the machine is setup to enforce disk usage limits for each job the machine executes.
> 
> How does the created cgroup sibling mess with the way that this `isenforcingdiskusage` flag works? My initial assumption was that `isenforcingdiskusage` is somehow using a parent cgroup to monitor disk usage, but I don't remember cgroups ever being capable of that.

I don't fully recall -- this is something the HTCondor developer who most heavily works on their cgroup management said would be needed. I suspect it has something to do with the logical volume mount in the outer container hiding the cgroup tree or making it unwritable.

spras/profiling.py
```diff
@@ -0,0 +1,86 @@
...
+                    cpu_usage = value
+                elif key == "user_usec":
+                    cpu_user = value
+                elif key == "system_usec":
+                    cpu_system = value
+    except Exception as e:
+        print(f"Failed to read cpu.stat from cgroup: {e}")
+
+    # Set up the header for the TSV file
+    header = ["peak_memory_bytes", "cpu_usage_usec", "cpu_user_usec", "cpu_system_usec"]
```
ntalluri, line 74: Is there a way to report runtime of an algorithm as well? I didn't see anything when I was looking at the cgroup documentation.

tristan-f-r: [This PR focuses on profiling pertaining to HTCondor - something like runtime is very platform agonistic (two calls to `time.time` would do), so we might want to bundle that in a different PR which can address #316 as well.]

tristan-f-r:
This created a bad merge conflict with https://github.com/Reed-CompBio/spras/pull/283. I would like to resolve it myself, but I don't know how to test the profiling changes here well enough to guarantee that I didn't miss anything.

spras/profiling.py
ntalluri: We should keep that separate from 316 as this has to do with the performance evaluation and capturing runtime.

spras/cgroup_wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# This script gets invoked by run_singularity_container when enable_profiling is set to true.
+# Its purpose is to move its own PID into the specified cgroup and then execute the container commands
+# passed to it. This has the effect of sticking all processes started by the container into the specified
+# cgroup, which lets us monitor them all in aggregate for resource usage.
+
+CGROUP_PATH="$1"
+echo "My cgroup path is: $CGROUP_PATH"
+shift
```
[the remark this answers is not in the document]
tristan-f-r: (I think this is clear? I'm unaware of a nice way to document this that isn't just code restating) `shift` removes the `CGROUP_PATH` before executing the rest of the passed in arguments with `exec "$@"`.

tristan-f-r (commented):
RWR and ST_RWR are missing the `out_dir` param in `run_container_and_log`, which is causing CI to fail. (I can fix and commit that change to this branch if that's okay.)

config/config.yaml
```diff
@@ -5,15 +5,21 @@ hash_length: 7
...
+# - singularity OR apptainer -- Apptainer (formerly Singularity) is useful in HPC/HTC environments where docker isn't allowed
+# - dsub -- experimental with limited support, used for running on Google Cloud
 container_framework: docker
 
-# Only used if container_framework is set to singularity, this will unpack the singularity containers
+# Enabling profiling adds a file called 'usage-profile.tsv' to the output directory of each algorithm.
+# The contents of this file describe the CPU utilization and peak memory consumption of the algorithm
+# as seen by its runtime container.
+# NOTE: Profiling is currently supported only when the container framework is set to apptainer/singularity
+#       and when the host system supports the 'cgroup' filesystem.
```
agitter, line 16: @jhiemstrawisc would this be a good place to document the requirement for HTCondor >= 24.8.0?

We should move documentation of the config file to readthedocs. In the meantime `config/config.yaml` is our placeholder to gather that information.

agitter:
ResponseNet is also missing `out_dir`

@tristan-f-r did you already re-review after Justin resolved merge conflicts or should I?

tristan-f-r (changes requested):
This still works. The merge conflict resolution introduced a double print of the container stdout (see review comment below), but this otherwise still seems fine 👍

tristan-f-r (approved):
(I would still like to see pathlib usage in `profiling.py`, but this looks good! :+1:)

spras/profiling.py
```diff
@@ -0,0 +1,90 @@
...
+    with open("/proc/self/cgroup") as f:
+        first_line = next(f).strip()
+        cgroup_rel = first_line.split(":")[-1].strip()
+
+    mycgroup = os.path.join("/sys/fs/cgroup", cgroup_rel.lstrip("/"))
+    peer_cgroup = os.path.join(os.path.dirname(mycgroup), f"spras-peer-{os.getpid()}")
+
+    # Create the peer cgroup directory
+    try:
+        os.makedirs(peer_cgroup, exist_ok=True)
```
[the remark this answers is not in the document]
agitter: I see this unresolved comment but am likely to approve anyway to get this merged. We may not be consistent with pathlib throughout the code base even though I agree it would be better to be.

PR #263, opened 2025-06-10
refactor(dataset): drop unused edge_table parameter

tristan-f-r:
This is the same idea as `self.interactome` and is both unused and never set to anything past `None.


agitter (commented):
This makes sense to me. Let's also get rid of it here:
https://github.com/Reed-CompBio/spras/blob/9b0e422be5445a9b816ce1bcf52c2434da91b3c7/spras/evaluation.py#L17

agitter (approved):

PR #270, opened 2025-06-12
refactor: cleanup prm

tristan-f-r:
Adds global type-hints related to PRM, and corrects the NotImplementedError error handling inside PRM.

This also takes the chance to update "NODEID" calls to Dataset.NODEID.

### DIFF
--- a/spras/prm.py
+++ b/spras/prm.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 
 from spras.dataset import Dataset
 
+
 class PRM(ABC):
     """
     The PRM (Pathway Reconstruction Module) class,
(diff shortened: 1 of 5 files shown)


spras/prm.py
```diff
@@ -1,32 +1,34 @@
...
+
 
 class PRM(ABC):
-    def __init__(self, params):
-        self.name = params['name']
-        self.inputdir = params['inputdir']
-        self.outputdir = params['outputdir']
-        self.params = params['params']
-        # debugging
-        #print('Params:',self.params)
```
agitter, line 15: Are you removing these because they weren't used anywhere? Or because they shouldn't be in an abstract class?

tristan-f-r: Both - I also remember that this was actively discouraged from Thursday last week as an OOP-paradigm.

agitter (approved):

PR #275, opened 2025-06-13
feat: container error handling

### LINKED ISSUE #21: Algorithms don't fail gracefully
Currently, we crash when an algorithm goes wrong in almost any step, including when a dataset is used which doesn't contain all the items an algorithm needs. 

This means that currently within a run all datasets must be compatible with all algorithms. This is also a problem if we want to be able to run large workflows with thousands of jobs, as the chances of a single job failing are pretty high. 

We need to be able to let some runs fail gracefully and continue with the overall workflow.

### LINKED ISSUE #119: Capturing errors thrown by algorithms run inside Docker containers
Currently our wrapper functions that run commands in containers catch exceptions aggressively. They print error messages but do not raise exceptions. This can be misleading because it may appear that an algorithm ran successfully even if it did not.

Specific examples are described in https://github.com/Reed-CompBio/spras/issues/21#issuecomment-1152571284 with Omics Integrator 1 and https://github.com/Reed-CompBio/spras/pull/103#issuecomment-1681526958 with DOMINO. DOMINO currently raises an error for all small inputs, so it is desirable to catch that specific error but not mask all errors.

### LINKED ISSUE #218: OmicsIntegrator2: Empty File Issue handling
Image

Using empty node files with Oi2 results in a error message from Oi2 that is hard to understand - can add in better error handling

tristan-f-r:
closes #21, closes #119, closes #218. This PR introduces the new general ContainerError to wrap all potential container handlers from now on.

DOMINO and OI2 are demonstrated using this error by searching for a specific error, or else bubbling the error up if no predicted error happened.


spras/omicsintegrator2.py
```diff
@@ -119,12 +119,27 @@ def run(edges=None, prizes=None, output_file=None, w=None, b=None, g=None, noise
...
+
+        try:
+            run_container_and_log('Omics Integrator 2',
+                                container_framework,
+                                container_suffix,
+                                command,
+                                volumes,
+                                work_dir)
+        except ContainerError as err:
+            needle = "all the input arrays must have same number of dimensions, but the array at index 0 has 2 dimension(s) and the array at index 1 has 1 dimension(s)"
```
tristan-f-r, line 135: @ntalluri I remember you having experience with OI2 errors - I wanted to check if this error only occurs when OI2 gets passed an empty DF, or if I should do a quick-check which makes sure that the DF is empty before ignoring this error.

tristan-f-r:
I still need to update the container handling, and I'm experiencing a weird error when checking for an empty dataframe for DOMINO.

ntalluri:
For the empty df, is the error because of empty sources/targets and prizes? I was checking a piece of code in dataset.py and  lines [99-101] have a check for an empty network. 

```
else:
            raise ValueError(
                f"Edge file {interactome_loc} must have three or four columns but found {num_cols}"
            )
```

I would assume we want a similar thing if there are 0 sources/targets or prizes.

spras/omicsintegrator2.py
ntalluri: For the empty df, is the error because of empty sources/targets and prizes? I was checking a piece of code in dataset.py and  lines [99-101] have a check for an empty network. 

```
else:
            raise ValueError(
                f"Edge file {interactome_loc} must have three or four columns but found {num_cols}"
            )
```

I would assume we want a similar thing if there are 0 sources/targets or prizes.

I think this PR is not the right place to update this issue though.

test/DOMINO/test_domino.py
```diff
@@ -25,57 +25,75 @@ class TestDOMINO:
...
     def test_domino_missing_network(self):
         # Test the expected error is raised when network argument is missing
         with pytest.raises(ValueError):
             # No network
             DOMINO.run(
-                active_genes=TEST_DIR+'input/domino-active-genes.txt',
+                active_genes=TEST_DIR / 'input' / 'simple' / 'domino-active-genes.txt',
                 output_file=OUT_FILE_DEFAULT)
 
+    def test_domino_small(self):
```
agitter, line 64: Does the existing simple test already cover the same condition? It has a comment "output_file should be empty". We may instead need the "required" test to use a larger input so this one can be the small input, empty output test. That could wait for a follow up PR because it is out of scope.

agitter (commented):
I'm having a lot of local test failures on this branch, but I haven't confirmed whether it is related to these changes or some other local problem.

spras/omicsintegrator2.py
```diff
@@ -129,12 +129,18 @@ def run(edges=None, prizes=None, output_file=None, w=None, b=None, g=None, noise
             command.extend(['--seed', str(seed)])
 
         container_suffix = "omics-integrator-2:v2"
+
+        # We use this later either if we encounter unrecoverable OI2 errors
+        # or as the main output file we want to post-process in parse_output.
+        output_tsv = Path(out_dir, 'oi2.tsv')
```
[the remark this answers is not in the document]
tristan-f-r: Accidentally left over from https://github.com/tristan-f-r/spras/blob/62009eb3e009757be2c04d5fe390016e759a90d6/spras/omicsintegrator2.py#L123-L141, before @ntalluri corrected me on how that old error should not be handled.

spras/containers.py
```diff
@@ -179,22 +219,30 @@ def run_container_and_log(name: str, framework: str, container_suffix: str, comm
...
                 out = str(out, "utf-8")
             print(indent(out))
     except docker.errors.ContainerError as err:
-        print(f"(Command formatted as list: `{err.command}`)")
-        print(f"An unexpected non-zero exit status ({err.exit_status}) inside the docker image {err.image} occurred:")
-        err = str(err.stderr if err.stderr is not None else "", "utf-8")
-        print(indent(err))
-    except Exception as err:
-        raise err
+        # TODO: does this lose us any information provided by stdout while the container was running?
```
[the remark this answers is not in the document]
tristan-f-r: This question is unfortunately not answerable in documentation, so we'll only know the answer on a version-by-version basis, but I'll do some digging around in the `dockerpy` code for any answer :+1:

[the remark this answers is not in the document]
tristan-f-r: https://github.com/Reed-CompBio/spras/issues/388.

test/DOMINO/test_domino.py
tristan-f-r: I think we have a larger network that seems to work for DIAMOnD that we may be able to reuse here - though I agree that that should be a separate PR.

agitter:
My last comment referred to "two unresolved comments pertain". One of those has an issue (#388). Is the other the need for a larger test case? If so, please create an issue. Then this should be ready to merge.

PR #283, opened 2025-06-20
fix(singularity): mv unpacked files

tristan-f-r:
This is a follow-up to https://github.com/Reed-CompBio/spras/pull/145#issuecomment-1970044310, taking the neutral approach where GC is left up to the user, but in such a way where it is made convenient. This addresses the `allpairs` creating git-commitable artifacts on the cwd.

This path could be made a parameter in the encompassing `run_container_singularity` function, though that seems out of scope for SPRAS itself.

[The commit name is misleading - it should be `mv`, not `gc`.]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,5 +142,5 @@ TempMat.mat
 # OSX-specific stuff
 **/.DS_Store
 
-# SPRAS singularity container
-spras.sif
+# Singularity cache
+unpacked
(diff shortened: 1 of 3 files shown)


agitter:
I looked at the code changes and don't have a strong opinion about them. If they improve quality of life for users running Singularity/Apptainer locally, it makes sense to me.

This is in Justin's domain, so he should weigh in before we merge.

PR #287, opened 2025-06-24
refactor: don't use globals in runner

tristan-f-r:
This makes 'unused variable' warnings in runner.py meaningful - this bit @AMINOexe and @gli527 when running into an 'algorithm is not supported' error despite having the class prepared. This also has the convenient bonus of adding types to `get_algorithm`.


agitter (changes requested):
This is a good implementation change.

Step 4 in the contributing guide will need slight rephrasing
> Import the new class LocalNeighborhood in spras/runner.py so the wrapper functions can be accessed.

now the algorithm is also added to the `algorithms` dict. Same for the "General steps for contributing a new pathway reconstruction algorithm" at the end of the contributing file.

agitter (approved):

PR #292, opened 2025-06-25
refactor!: config parsing

### LINKED ISSUE #299: Make the config file parsing robust to missing required fields
[I do not like GitHub subissues right now. I didn't know it just does this. I meant to just mention this in #292...]

tristan-f-r:
Closes #299. I did want to keep this separate from https://github.com/Reed-CompBio/spras/pull/228 so that way those details didn't get lost, but this does the proper work of using structural parsing on the config, throwing useful errors on missing keys, wrongly-written keys, default paramaters, and with proper input validation and casting.

This doesn't clean up the parsing done in #228, but it lays the ground for cleaner configuration parsing.

I've also made some issues for suggestions for the configuration format #297 #298 #299 #309.
For users, this affects error messages: Now, we get good error messages for missing keys and too many keys, or improperly formatted values.

> [!NOTE]
> The breaking change indicator in the PR header (`!`) is present because this will cause the config to raise an error if there are too many keys present inside any one of the objects (except for the `runs`... object). This also removes `reconstruction_settings -> run`.


annaritz (commented):
Suggestion: rename `raw_config` to `config_specs`. Keep `config.py` as named for now, but a later refactor might be in order. 

@agitter if you can take a look at this and give some comments, that would be great!

test/PathLinker/test_pathlinker.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-import spras.config as config
+import spras.config.config as config
```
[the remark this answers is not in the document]
agitter: I support a renaming. All the "config"s here is unaesthetic.

spras/config/config.py
```diff
@@ -42,39 +43,41 @@ def init_from_file(filepath):
     try:
         with open(filepath, 'r') as yaml_file:
             config_dict = yaml.safe_load(yaml_file)
-    except FileNotFoundError:
-        print(f"Error: The specified config '{filepath}' could not be found.")
-        return False
```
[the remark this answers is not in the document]
agitter: I don't remember why. @jhiemstrawisc added that initially. Was that a Python convention or a Snakemake convention?

config/config.yaml
```diff
@@ -164,8 +164,6 @@ reconstruction_settings:
     # TODO move to global
     reconstruction_dir: "output"
 
-  run: true
```
tristan-f-r, line 167: I noticed in a separate thread on #72 that we might want to try incorporating a 'formal' deprecation system instead. Should we try this here instead of outright removing it?

agitter: We still have a small enough user base that I am fine removing syntax or features. Deprecating them is a better practice but adds overhead for us.

spras/config/config.py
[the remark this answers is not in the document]
tristan-f-r: image

Looking at this again, it also seems that `init_from_file` is only used in test cases.

spras/config/analysis_schema.py
```diff
@@ -0,0 +1,49 @@
+from pydantic import BaseModel, ConfigDict
+
+from spras.config.util_enum import CaseInsensitiveEnum
+
+
+class SummaryAnalysis(BaseModel):
+    include: bool
```
agitter, line 7: Why are these sometimes `bool` and sometimes `bool = True`?

spras/config/analysis_schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+    model_config = ConfigDict(extra='forbid')
```
agitter, line 9: As a newcomer, it isn't clear why this `CongfigDict` is included at the end of some classes and not others. Is it skipped when a class is a subclass of another?

spras/config/analysis_schema.py
agitter, line 1: Why is the analysis split out into a separate file?

spras/config/config.py
agitter, line 1: There may be some unused imports at the top of this file now. I'm not sure if that is new or old.

spras/config/config.py
```diff
@@ -42,39 +43,41 @@ def init_from_file(filepath):
...
-        # Since process_config winds up modifying the raw_config passed to it as a side effect,
-        # we'll make a deep copy here to guarantee we don't break anything. This preserves the
-        # config as it's given to the Snakefile by Snakemake
+    def __init__(self, raw_config: dict[str, Any]):
+        # Since snakemake provides an empty config, we provide this
+        # wrapper error first before passing validation to pydantic.
+        if raw_config == {}:
+            raise ValueError("Config file cannot be empty. Use --configfile <filename> to set a config file.")
+
+        parsed_raw_config = RawConfig.model_validate(raw_config)
```
agitter, line 62: I assume this no longer modifies `raw_config` which is why we no longer make a deep copy?

spras/config/config.py
```diff
@@ -105,69 +110,29 @@ def __init__(self, raw_config):
...
-        self.datasets = {dataset["label"]: dict(dataset) for dataset in raw_config["datasets"]}
-
-        for key in self.datasets:
-            pattern = r'^\w+$'
-            if not bool(re.match(pattern, key)):
-                raise ValueError(f"Dataset label \'{key}\' contains invalid values. Dataset labels can only contain letters, numbers, or underscores.")
+        self.datasets = {}
+        for dataset in raw_config.datasets:
+            label = dataset.label
+            if label in self.datasets:
```
agitter, line 130: Do we want case-insensitive comparisons? On some operating systems, that could be important when creating paths later.

spras/config/config.py
```diff
@@ -181,35 +146,36 @@ def process_config(self, raw_config):
...
 
-            if "directed" in cur_params:
-                print("UPDATE: we no longer use the directed key in the config file")
-                cur_params.pop("directed")
+            if cur_params.directed is not None:
+                warnings.warn("UPDATE: we no longer use the directed key in the config file", stacklevel=2)
+
+            cur_params = cur_params.__pydantic_extra__
+            if cur_params is None:
+                raise RuntimeError("An internal error occured: ConfigDict extra should be set on AlgorithmParams.")
```
agitter, line 175:
```suggestion
                raise RuntimeError("An internal error occurred: ConfigDict extra should be set on AlgorithmParams.")
```

This message reads as if it is aimed at developers. Will a user know what to do?

test/test_config.py
```diff
@@ -22,10 +29,26 @@ def get_test_config():
...
+        }, {
+            "label": "alg2",
+            "data_dir": "faux",
+            "edge_files": [],
+            "other_files": [],
+            "node_files": []
+        }],
+        "gold_standards": [{
+            "label": "gs1",
+            "dataset_labels": [],
```
agitter, line 47: Does this test config work when the gold standard is not matched with any dataset label?

agitter (commented):
The overall pull request is a good idea. I follow the changes at a high level and can read the new schemas fine. I haven't spot checked every line of code in full detail.

It may be hard for the rest of the team and new contributors who haven't worked with pydantic to understand the schema concept well enough to extend the config file and associated schemas. Do we want more comments or light docs on how these pieces work to help future maintainers? Or is it better to send them straight to a pydantic tutorial?

spras/config/analysis_schema.py
tristan-f-r: To make it easier to read. I could have also had one of those header comments (e.g. '# For Analysis), but files just seemed like a nicer way to do this.

spras/config/config.py
tristan-f-r: We should lint for this.

spras/config/config.py
tristan-f-r: A user will not know what to do. However, I can not see a scenario where this would not be caught in testing, since ConfigDict is set in AlgorithmParams.

test/test_config.py
tristan-f-r: We can add this test case :+1:

spras/config/config.py
agitter: Yes, I initially left that off because the `runner.py` imports show up as unused, but you addressed that in a separate pull request.

spras/config/analysis_schema.py
agitter: The `bool = False` made sense, but I also see `bool = True` and `bool` without a default. Do we intend to have all three?

spras/config/analysis_schema.py
agitter: I was looking at `MlLinkage` and `MlMetric` and inferred they are special for some reason because they are used in `MlAnalysis`. But I don't know pydantic, so this was mysterious to me.

tristan-f-r: `MlLinkage` and `MlMetric` are non-pydantic enums: pydantic has [special handling for enums](https://docs.pydantic.dev/2.0/usage/types/enums/) which auto-handles non-conforming values. (We do have a specially defined `CaseInsensitiveEnum` which is why this isn't just over `Enum` instead)

test/test_config.py
tristan-f-r: ([Oh, there already was one](https://github.com/Reed-CompBio/spras/blob/ea59e4cec7d5197fd394f3704a5406facaa6f1a9/test/test_config.py#L234-L235))

tristan-f-r:
> It may be hard for the rest of the team and new contributors who haven't worked with pydantic to understand the schema concept well enough to extend the config file and associated schemas. Do we want more comments or light docs on how these pieces work to help future maintainers? Or is it better to send them straight to a pydantic tutorial?

I wanted to use pydantic to do some algorithm parameter validation with #321 (accompanied by #296). I believe the pydantic use for configuration is decently straightforward, but, if pydantic becomes part of algorithm wrapping, adding a section that links to [pydantic's concepts: model page](https://docs.pydantic.dev/latest/concepts/models/) would be good.

test/test_config.py
agitter: Good catch

agitter:
The questions from my initial review have been addressed. I would still find it helpful for myself and future maintainers who look at this code to have a couple more comments about the parts that deviate from patterns, like the `CaseInsensitiveEnum`s, and reference the relevant concepts in the pydantic concepts model page. As we add more post-processing and evaluation, several of us will need to be comfortable modifying this.

tristan-f-r:
I'll merge `analysis_schema` with the regular `schema` file since I don't want to duplicate a top-level comment.

spras/config/analysis_schema.py
tristan-f-r: When appropriate, yes :+1:

tristan-f-r:
> Sorry, I thought we had to define each of these rules in schema for each algorithm (similar for what is done for ML and evaluation). How does the schema check for the validity for each of the parameters for the algorithms?

Augh, I got my PRs mixed up. Nothing needs to be done for this PR, since there is no argument validation yet. That's all done in #329.

ntalluri:
I agree with Tony, it would be really helpful if the file itself included more context about how everything connects and is being used. While I know the documentation is available, having clear explanations directly in the file about what each class does and why certain patterns are used over others would make it much easier to follow and understand the design choices.  Right now, I can mostly infer what’s happening, but I won’t be fully confident in the reasoning until I try adding something myself aat which point having those comments and explanations would be especially helpful.

tristan-f-r:
I think I'm primarily confused because I'm struggling to understand when this behavior wasn't implicit already - these were exactly the same design decisions we were making, except for `ConfigDict(extra='forbid')` which already has documentation attached if you hover over it.

tristan-f-r:
I could add these comments but they either:
- Feel self-explanitory
- I'm not sure where the design decision came from (for example, from @agitter's original comment about the `include = False/True/Optional` were strictly translated from the original `config.py`. I honestly don't know why these decisions were made)

tristan-f-r:
I want to avoid duplicating a lot of the comments from `config.yaml` till we have a way to actually automatically generate it in #329, but these seem like enough comments to motivate some of the decisions in this file.

spras/config/analysis_schema.py
ntalluri: Similar to this question, I was also confused until I figured it out myself. I think information like this could be useful to put as a docstring for each of these methods.

spras/config/schema.py
```diff
@@ -0,0 +1,166 @@
...
+
+    model_config = ConfigDict(extra='forbid')
+
+class EvaluationAnalysis(BaseModel):
+    include: bool
+    aggregate_per_algorithm: bool = False
+
+    model_config = ConfigDict(extra='forbid')
+
+class Analysis(BaseModel):
```
ntalluri, line 66: There is a couple things that get coupled with Evaluation and ML post processing (the logic/coupling is at the bottom of config.py). Is there a way to have a user set the values to true or false based on the coupling if they have evaluation turned to true?

tristan-f-r: Since Pydantic doesn't have a nice solution for dependent fields, getting that system working is a little weird. I've avoided it both here and in #329, but it might be easier just to redesign the config around using nested properties over interdependent fields to get real JSON schema support at that level.

spras/config/config.py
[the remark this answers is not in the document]
agitter: When I talked to Justin about this, he did not express a strong preference or recollection about the existing pattern of printing a message and returning. It's safe to change.

ntalluri:
I’ll do another review before merging.

spras/config/util.py
```diff
@@ -0,0 +1,19 @@
+from enum import Enum
+from typing import Any
+
+
+# https://stackoverflow.com/a/76883868/7589775
+class CaseInsensitiveEnum(str, Enum):
+    """
+    We prefer this over Enum to make sure the config parsing
```
ntalluri, line 8: I'm a little confused by this comment. I thought you explicitly have to define an Enum class when you want to use Enums? Is there a way to use Enum by itself?

spras/config/schema.py
```diff
@@ -0,0 +1,168 @@
...
+    model_config = ConfigDict(extra='forbid')
+
+class RawConfig(BaseModel):
+    # TODO: move these container values to a nested container key
+    container_framework: ContainerFramework = ContainerFramework.docker
+    unpack_singularity: bool = False
+    container_registry: ContainerRegistry
+
+    hash_length: int = DEFAULT_HASH_LENGTH
+    "The length of the hash used to identify a parameter combination"
```
ntalluri, line 157: Is this supposed to be a comment? Or a Field(description="")?

spras/config/schema.py
```diff
@@ -0,0 +1,168 @@
...
+    model_config = ConfigDict(extra='forbid')
+
+
+# The default length of the truncated hash used to identify parameter combinations
+DEFAULT_HASH_LENGTH = 7
+
+def label_validator(name: str):
+    """
+    A validator takes in a label
+    and ensures that it's alphanumeric (with underscores).
```
ntalluri, line 83:
```suggestion
    and ensures that it contains only letters, numbers, or underscores.
```

spras/config/schema.py
```diff
@@ -0,0 +1,168 @@
...
+
+class ContainerRegistry(BaseModel):
+    base_url: str
+    owner: str = Field(description="The owner or project of the registry")
+
+    model_config = ConfigDict(extra='forbid')
+
+class AlgorithmParams(BaseModel):
+    include: bool
+    directed: Optional[bool] = None
```
ntalluri, line 106: what is this parameter?

spras/config/schema.py
```diff
@@ -0,0 +1,168 @@
...
+
+class GoldStandard(BaseModel):
+    label: Annotated[str, AfterValidator(label_validator("Gold Standard"))]
+    node_files: list[str]
+    data_dir: str
+    dataset_labels: list[str]
+
+    model_config = ConfigDict(extra='forbid')
+
+class Locations(BaseModel):
```
ntalluri, line 139: Why is Locations separated from ReconstructionSettings? Can we not just use reconstruction_dir: str in ReconstructionSettings?

config/config.yaml
```diff
@@ -169,8 +169,6 @@ reconstruction_settings:
...
```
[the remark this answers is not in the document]
tristan-f-r: This PR made one breaking change to the config, which is to remove the `run: bool` param entirely from reconstruction settings, namely because it was not used and seemingly served no purpose.

test/test_config.py
```diff
@@ -3,8 +3,15 @@
 import numpy as np
 import pytest
 
-import spras.config as config
+import spras.config.config as config
+from spras.config.schema import DEFAULT_HASH_LENGTH
```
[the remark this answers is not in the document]
tristan-f-r: It's a test we can delete, but we were checking that the hash length was `DEFAULT_HASH_LENGTH` if it wasn't specified:

https://github.com/Reed-CompBio/spras/blob/18a173f9442ef9acc0b325e8417727e73d1c69b6/test/test_config.py#L149-L151

spras/config/schema.py
tristan-f-r: It's a holdout deprecated parameter ([see main branch code](https://github.com/Reed-CompBio/spras/blob/32a0424546a94b3d8f17362bcb4b6c4e6e6398fa/spras/config.py#L207-L209))

spras/config/util.py
tristan-f-r: I don't think I'm properly understanding this review comment, but you can extend Enums before using them.

spras/config/schema.py
ntalluri: Oh interesting. Do you know why we can just leave it as a string like this?

spras/config/schema.py
ntalluri: Oh I see. I'm not sure if we should delete it. We want to make sure people who use old configs are still able to run the code which is why that part in the main branch code exists.

tristan-f-r: I think we mentioned in an earlier conversation that we don't need deprecation logic at the moment. We can keep it here, though, since this PR should probably avoid as much breaking changes as possible.

spras/config/schema.py
ntalluri: Ok this makes more sense now

ntalluri:
@agitter Tristan addressed all of my remaining questions. This is ready to merge, but we should hold off until we look over the other PRs today in case we decide to wait on merging.

ntalluri:
I have a feeling we’ll end up merging it later today, but it would be helpful to first have a conversation about all the PRs that depend on this one. There may be a reason not to merge it yet, though nothing comes to mind right now.

tristan-f-r:
Actually, I'll merge upstream first.

tristan-f-r:
> I want to split off an independent chunk of https://github.com/Reed-CompBio/spras/pull/329 which depends on this PR that should lighten the diff on https://github.com/Reed-CompBio/spras/pull/329 a little (about 13 lines of actual code with a 1k line artifact),

#358.

agitter (approved):
>  it would be helpful to first have a conversation about all the PRs that depend on this one

That was to understand the dependent changes and how much further restructuring was coming and what new maintenance burden, if any, would be introduced.

PR #293, opened 2025-06-25
fix: multithreading error

### LINKED ISSUE #268: GitHub Actions Run Snakemake Workflow Failing at Random
For the past couple of days, my Github Actions job for running the SnakeMake workflow entirely will fail randomly on one of the algorithms. The algorithm it fails on also seems to change every time, but every time it is an error similar to below:

```
RuleException:
OSError in file /home/runner/work/spras/spras/Snakefile, line 194:
Cannot save file into a non-existent directory: 'output/prepared/data0-meo-inputs'
  File "/home/runner/work/spras/spras/Snakefile", line 194, in __rule_prepare_input
  File "/home/runner/work/spras/spras/spras/runner.py", line 59, in prepare_inputs
  File "/home/runner/work/spras/spras/spras/meo.py", line 112, in generate_inputs
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/util/_decorators.py", line 211, in wrapper
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/core/generic.py", line 3721, in to_csv
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/util/_decorators.py", line 211, in wrapper
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/io/formats/format.py", line 1189, in to_csv
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/io/formats/csvs.py", line 241, in save
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/io/common.py", line 735, in get_handle
  File "/usr/share/miniconda3/envs/spras/lib/python3.11/site-packages/pandas/io/common.py", line 598, in check_parent_directory
```

If I rerun the job once (it might fail again) or a couple times it will start to pass again.

### LINKED ISSUE #279: Multi-threaded Race Condition?
When testing the spras-benchmarking CI pipeline, I ran into this issue when using `--cores 2`:

```
RuleException:
FileNotFoundError in file /home/runner/work/spras-benchmarking/spras-benchmarking/spras/Snakefile, line 193:
[Errno 2] No such file or directory: 'output/prepared/hiv05-omicsintegrator1-inputs/dummy_nodes.txt'
  File "/home/runner/work/spras-benchmarking/spras-benchmarking/spras/Snakefile", line 193, in __rule_prepare_input
  File "/home/runner/work/spras-benchmarking/spras-benchmarking/spras/spras/runner.py", line 59, in prepare_inputs
  File "/home/runner/work/spras-benchmarking/spras-benchmarking/spras/spras/omicsintegrator1.py", line 95, in generate_inputs
[Tue Jun 17 16:16:44 2025]
Error in rule prepare_input:
    jobid: 0
    input: output/hiv05-merged.pickle
    output: output/prepared/hiv05-omicsintegrator1-inputs
```

https://github.com/tristan-f-r/spras-benchmarking/actions/runs/15713733225/job/44278169174

[logs_40272997423.zip](https://github.com/user-attachments/files/20780561/logs_40272997423.zip)

tristan-f-r:
Anyone is welcome to try to create a test case for this, but I've tested it by running this patch on spras-benchmarking several times over 16 cores: without the patch, Snakemake fails almost every single time. With this patch, it has never failed despite running this several times.

Closes #279. Closes #268.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -226,7 +226,7 @@ def collect_prepared_input(wildcards):
             for input in prepared_inputs:
                 if not os.path.isfile(input):
                     missing_inputs.append(input)
-            # If any expected files were missing, delete the entire directory so the call below triggers running prepare_input
+            # If any expected files were missing, ask to delete the entire directory so the call below triggers running prepare_input
             if len(missing_inputs) != 0:
                 raise RuntimeError(f"Not all input files were provided. (Missing {missing_inputs})\n" +
                     "fTo prevent multithreading errors, please remove the {prepared_inputs} directory.")
--- a/Snakefile
+++ b/Snakefile
@@ -214,14 +214,22 @@ def collect_prepared_input(wildcards):
     prepared_inputs = expand(f'{prepared_dir}{SEP}{{type}}.txt',type=runner.get_required_inputs(algorithm=wildcards.algorithm))
     # If the directory is missing, do nothing because the missing output triggers running prepare_input
     if os.path.isdir(prepared_dir):
-        # If the directory exists, confirm all prepared input files exist as well (as opposed to some or none)
-        missing_inputs = False
-        for input in prepared_inputs:
-            if not os.path.isfile(input):
-                missing_inputs = True
-        # If any expected files were missing, delete the entire directory so the call below triggers running prepare_input
-        if missing_inputs:
-            shutil.rmtree(prepared_dir)
+        # First, check if .snakemake_timestamp, the last written file in a directory rule,
+        # exists. This prevents multithreading errors if we accidentally read a directory
+        # before it finishes. A proper API for querying this is opened as an issue at
+        # https://github.com/snakemake/snakemake/issues/439.
+        if not os.path.isfile(os.path.join(prepared_dir, '.snakemake_timestamp')):
+            checkpoints.prepare_input.get(**wildcards)
+        else:
+            # If the directory exists, confirm all prepared input files exist as well (as opposed to some or none)
+            missing_inputs = []
+            for input in prepared_inputs:
+                if not os.path.isfile(input):
+                    missing_inputs.append(input)
+            # If any expected files were missing, delete the entire directory so the call below triggers running prepare_input
+            if len(missing_inputs) != 0:
+                raise RuntimeError(f"Not all input files were provided. (Missing {missing_inputs})\n" +
+                    "fTo prevent multithreading errors, please remove the {prepared_inputs} directory.")
 
     # Check whether prepare_input has been run for these wildcards (dataset-algorithm pair) and run if needed
     # The check is executed by checking whether the prepare_input output exists, which is a directory
--- a/Snakefile
+++ b/Snakefile
@@ -219,6 +219,13 @@ def collect_prepared_input(wildcards):
         # before it finishes. A proper API for querying this is opened as an issue at
         # https://github.com/snakemake/snakemake/issues/439.
         if not os.path.isfile(os.path.join(prepared_dir, '.snakemake_timestamp')):
+            # Running this has two goals:
+            # - If there is another thread running this, in correspondence with
+            # https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution,
+            # this will raise a IncompleteCheckpointException and poll again until that checkpoint is ready.
+            # - If the prior Snakemake execution was forcefully terminated (and prepared_inputs didn't finish,
+            # which must be the case since .snakemake_timestamp is the last file to be added),
+            # we can reproduce the prepared_inputs file, allowing resilliency against unexpected crashes.
             checkpoints.prepare_input.get(**wildcards)
         else:
             # If the directory exists, confirm all prepared input files exist as well (as opposed to some or none)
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -214,14 +214,22 @@ def collect_prepared_input(wildcards):
...
-        for input in prepared_inputs:
-            if not os.path.isfile(input):
-                missing_inputs = True
-        # If any expected files were missing, delete the entire directory so the call below triggers running prepare_input
-        if missing_inputs:
-            shutil.rmtree(prepared_dir)
+        # First, check if .snakemake_timestamp, the last written file in a directory rule,
+        # exists. This prevents multithreading errors if we accidentally read a directory
+        # before it finishes. A proper API for querying this is opened as an issue at
+        # https://github.com/snakemake/snakemake/issues/439.
```
tristan-f-r, line 220: The worst-case scenario here is that someone runs two Snakemake files with differing `generate_input` functions at the same time. I also can not imagine a realistic scenario where this happens other than on some strangely shared-fs dev machine.

tristan-f-r: Note: This happens because this entire rule isn't sequential. That's the heart of this problem: the wildcards are being parallelized across threads, despite multiple parameters locking in to a single folder. The other solution to fix this is to attach the parameters hash to the prepare-inputs folder, but that is a lot of unnecessary copying.

Snakefile
```diff
@@ -214,14 +214,22 @@ def collect_prepared_input(wildcards):
...
+        else:
+            # If the directory exists, confirm all prepared input files exist as well (as opposed to some or none)
+            missing_inputs = []
+            for input in prepared_inputs:
+                if not os.path.isfile(input):
+                    missing_inputs.append(input)
+            # If any expected files were missing, ask to delete the entire directory so the call below triggers running prepare_input
+            if len(missing_inputs) != 0:
+                raise RuntimeError(f"Not all input files were provided. (Missing {missing_inputs})\n" +
+                    f"To prevent multithreading errors, please remove the {prepared_inputs} directory.")
```
annaritz, line 232: Maybe some silly questions:
1. what happens when snakemake throws a runtime error for one rule in a multithreaded situation? 
2. In a single-threaded case, it makes sense to delete the directory (the original behavior), right? If so, is there a way to only throw this error when using snakemake with more than 1 core? 
3. Is there ever a case where the `.snakemake_timestamp` is written but there's a missing input? I can't think of this scenario off-hand - I suppose that's the reason for the `.snakemake_timestamp` file in the first place.

tristan-f-r: 1. Snakemake will drop the entire execution and clean up output files that were being generated at the time of the error ([this does not happen with `log`](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files)).
2. [Despite some searching, I'm unsure of a way to get the number of cores passed into Snakemake] I'm not sure if I would necessarily trust the behavior of deleting the original directory: those directory changes would happen if (this list might be non-exhaustive):
  - The snakemake input is tampered with between executions
  - The algorithm is updated
  - `generate_inputs` problematically generates all of the required inputs every time
  For the first and third options, I would not want Snakemake to repair itself silently (the first option really shouldn't happen at all). I'm unsure what [the intent of the commit that introduced was](https://github.com/Reed-CompBio/spras/commit/93f110b92dedb86a53e4af7153b2d55383c2c5a2), but my best guess based on the commit message is that it was added in case someone tampers with the input folder.
3. That case would happen if `generate_inputs` was probabilistic, but `PRA#run` catches this anyway (PR #290 contains all of the instances where this is caught).

Snakefile
ntalluri: Was the same work happening on multiple threads or just the same wildcards were being sent to multiple threads?

Snakefile
agitter: >  my best guess based on the commit message is that it was added in case someone tampers with the input folder

It's taxing my memory a bit to remember the intention of that 4 year old commit, but looking back your assumption seems correct. The code was added to recover in case the input directory exists but a user had manually deleted some of the expected input files from the directory.

agitter (commented):
This generally looks good. I'd like to confirm understand the fix in detail before merging.

Snakefile
```diff
@@ -214,14 +214,22 @@ def collect_prepared_input(wildcards):
...
-                missing_inputs = True
-        # If any expected files were missing, delete the entire directory so the call below triggers running prepare_input
-        if missing_inputs:
-            shutil.rmtree(prepared_dir)
+        # First, check if .snakemake_timestamp, the last written file in a directory rule,
+        # exists. This prevents multithreading errors if we accidentally read a directory
+        # before it finishes. A proper API for querying this is opened as an issue at
+        # https://github.com/snakemake/snakemake/issues/439.
+        if not os.path.isfile(os.path.join(prepared_dir, '.snakemake_timestamp')):
+            checkpoints.prepare_input.get(**wildcards)
```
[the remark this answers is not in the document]
agitter: Perfect. That clarifies everything for me.

agitter (approved):

PR #300, opened 2025-06-26
chore: use pixi

tristan-f-r:
Pixi is a conda alternative [used by snakemake ](https://github.com/snakemake/snakemake/blob/9030f352b6a0e4b2def508cc62260b9cba613693/pyproject.toml#L88-L90) which installs faster and removes most of the conda headaches we get. It has classic environment support, but it's generally intended to be ran with `pixi run <command>`.

This PR is a draft as this needs to update the documentation to encourage using `pixi`.

### DIFF
--- a/.github/workflows/test-spras.yml
+++ b/.github/workflows/test-spras.yml
@@ -14,9 +14,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Install pixi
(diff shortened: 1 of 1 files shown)


tristan-f-r:
Okay, this CI warning is too much to consider using `pixi`. I don't know how snakemake got this working well, but the time investment has been ridiculous for an easy to use tool.

Issue #155, opened 2024-05-07
Support hypergraphs and hypergraph algorithms


agitter:
This issue is for brainstorming what would be involved in generalizing SPRAS to support hypergraphs and hypergraph algorithms. One possible goal could be showing how converting hypergraphs to graphs and running the graph algorithms in SPRAS is ineffective.

There are no short term plans to work on this.

gkrieg:
Are any of the datasets naturally a hypergraph that have instead been converted into a graph for graph algorithms? I know that PPIs are typically graphs, and may be difficult to convert into hypergraphs, but any reaction network can be naturally converted to a hypergraph. 

I suspect that many evaluation metrics are likely directly applicable to hyperpaths, as long as they refer to the original nodes and reactions within the reaction network.

agitter:
We don't have many datasets in SPRAS yet and are making a push to add them this summer. We will likely add pathway databases, so those (e.g. Reactome) would support hypergraph representations.

gkrieg:
That would be great. Do you already have a set of metrics to compare different pathway reconstruction methods? If so, could you point me to them?

agitter:
We don't have anything written up yet. @ntalluri is working on that. We will likely use fairly standard machine learning classification metrics. The question is what to use for ground truth and when we will have edges versus only nodes for ground truth, which is likely when working with real data.

gkrieg:
Sounds good. I would love to test our hyperpath and factory methods on pathway reconstruction benchmarks once they are available.

tristan-f-r:
We can support this by using comma-separated gene names on the current interactome file format, giving the idea of a mixed hypergraph:

```
A	D	U
B,C,D	B,C,D	U
C	E	D
B,D	G,F,E	D
A	A	U
```

The main problem is representing those undirected hyperedges well: we should be tolerant of empty columns in those cases. For example, all of these should produce the same graph:

```
A	D	U
B,C,D	B,C,D	U
```


```
A	D	U
C,D	B,C	U
```


```
A	D	U
	B,C,D	U
```


```
A	D	U
B,C,D		U
```

tristan-f-r:
I should mention this here: I'm currently working on a PR that adds mixed hypergraph support to halp: https://github.com/Murali-group/halp/pull/34.

PR #310, opened 2025-07-01
chore: bump dependencies

### LINKED ISSUE #134: Clustering warning in sklearn version 1.2
After we upgraded sklearn to version 1.2, there is a warning shown when I run the tests:
```
site-packages\sklearn\cluster\_agglomerative.py:983: FutureWarning: Attribute `affinity` was deprecated in version 1.2 and will be removed in 1.4. Use `metric` instead
```

### LINKED ISSUE #210: Different dependency version matching between conda and python
We may have slightly different dependency matching affecting the package versions that are installed from conda and pip. That means that when following the install practices recommended in our readme and test case after #202, some packages are installed by conda, then uninstalled and reinstalled by pip.

See https://github.com/Reed-CompBio/spras/pull/202#issuecomment-2744386309

tristan-f-r:
Closes #210. Closes #134.

- [x] Depends on #340.


ntalluri:
Do you have an example of this (the image) that you could add as a comment?

tristan-f-r:
NixOS 25.05.20250108.bffc22e (Warbler) on x86_64, though I do think this issue is reproducing on CI, so I doubt this is platform dependent. I'll reproduce it once I switch from working on https://github.com/Murali-group/halp/pull/34 for #155.

tristan-f-r:
@ntalluri (for https://github.com/tristan-f-r/spras/blob/54bd449d376756cb5d49785c9cf85e2d7c17aa21/test/ml/test_ml.py#L83-L95):

```
Got:
      algorithm      PC1      PC2
0  test-data-s1 -2.00665 -0.98659
1  test-data-s2 -1.52765  1.07995
2  test-data-s3  3.53430 -0.09336       algorithm      PC1      PC2
0  test-data-s1 -2.00665 -0.98659
1  test-data-s2 -1.52765  1.07995
2  test-data-s3  3.53430 -0.09336

Expected:
      algorithm      PC1      PC2
0  test-data-s1  2.00665 -0.98659
1  test-data-s2  1.52765  1.07995
2  test-data-s3 -3.53430 -0.09336       algorithm      PC1      PC2
0  test-data-s1 -2.00665 -0.98659
1  test-data-s2 -1.52765  1.07995
2  test-data-s3  3.53430 -0.09336
```

ntalluri:
Based on what I’m reading, bumping the version is a good idea because it improves PCA in a meaningful way. Before, PCA signs were chosen by looking at how the data looked after projection (transformed data), but now they are chosen by inspecting each component vector directly. Previously, each solver decided the signs in its own way, which meant that changing the solver could cause components to unexpectedly flip signs. While the mathematical results seem to always still be correct, this causes confusion.  The new approach in version 1.7 makes the sign choice independent of the solver because it relies only on the component vector’s own values, ensuring we always get consistent component signs no matter which solver we use.

We don't allow a user to pick a solver, we have it set to auto at the moment.
https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA:~:text=svd_solver%7B%E2%80%98auto%E2%80%99%2C%20%E2%80%98full,optionally%20truncated%20afterwards.

tristan-f-r:
[Let me know if this commit is the right way to approach that.](https://github.com/Reed-CompBio/spras/pull/310/commits/ae02c993b8fe7fff590d3cf17784a46f4135dd13) To my understanding, this does mess with our sampling test because our component vectors differ per run; hopefully, the aforementioned commit addresses that.

test/ml/test_ml.py
```diff
@@ -102,7 +106,7 @@ def test_pca_robustness(self):
             coord = coord.round(5)  # round values to 5 digits to account for numeric differences across machines
             coord.sort_values(by='algorithm', ignore_index=True, inplace=True)
 
-            assert coord.equals(expected)
+            assert coord.equals(expected) or coord.equals(expected_other)
```
ntalluri, line 109: Is there 2 possible outputs?

ntalluri:
> [Let me know if this commit is the right way to approach that.](https://github.com/Reed-CompBio/spras/pull/310/commits/ae02c993b8fe7fff590d3cf17784a46f4135dd13) To my understanding, this does mess with our sampling test because our component vectors differ per run; hopefully, the aforementioned commit addresses that.

For that specific test, we want the output to be the same despite shuffling the rows and columns. So having 2 different outputs is not what we would want.

test/ml/test_ml.py
agitter: I believe the answer is yes? It should be valid to flip the solution.

tristan-f-r:
> Can you help me get caught up on the core question? It seems valid to have a transformation of the PCA solution where the signs are flipped.

The essence of the test that was broken here was row shuffling preserving PCA output:

```py
dataframe_shuffled = dataframe.sample(frac=1, axis=1)  # permute the columns
ml.pca(dataframe_shuffled, OUT_DIR + 'pca-shuffled-columns.png', OUT_DIR + 'pca-shuffled-columns-variance.txt',
    OUT_DIR + 'pca-shuffled-columns-coordinates.tsv')
coord = pd.read_table(OUT_DIR + 'pca-shuffled-columns-coordinates.tsv')
coord = coord.round(5)  # round values to 5 digits to account for numeric differences across machines
coord.sort_values(by='algorithm', ignore_index=True, inplace=True)
```

This now needs an assertion check for two possible signings:

```
assert coord.equals(expected) or coord.equals(expected_other)
```

test/ml/expected/expected-pca-coordinates-2.tsv
```diff
@@ -0,0 +1,4 @@
+algorithm	PC1	PC2
```
agitter, line 1: If these two files are the same except for the PC1 sign, do we want them to be stored separately? Or is it better to track a single file and have the test make a copy and negate PC1?

agitter: Fair point. I like `negated` better.

agitter (approved):
I haven't run locally in the new environment, but all the changes look okay.

ntalluri:
If it is the ensembling one, there was a data.pickle I added that has a toy dataset network that is needed for ensembling in general that we use to ensure we calculate recall correct. I'm not sure why that would need to be updated (if it is this pickled dataset) because it has nothing to do with the pca.

agitter:
> Regardless, we probably should stop committing artifacts to main anyway - this causes problems; see https://github.com/Reed-CompBio/spras/pull/320 for an example of a large test refactor because of a related issue

This seems worthwhile to discuss in a new issue as a testing strategy change if it affects multiple tests.

tristan-f-r:
https://github.com/Reed-CompBio/spras/issues/339

tristan-f-r:
> @tristan-f-r Which PR was it? The ensembling one?

It was #212 ([origin commit](https://github.com/Reed-CompBio/spras/commit/c3fe362d16371421a78d25e4480d92b6c6c77f9b)).

tristan-f-r:
> If it is the ensembling one, there was a data.pickle I added that has a toy dataset network that is needed for ensembling in general that we use to ensure we calculate recall correct. I'm not sure why that would need to be updated (if it is this pickled dataset) because it has nothing to do with the pca.

Pickled objects contain internal representations of the object, and are subject to change throughout package versions. In this case, `pandas` itself changed.

tristan-f-r:
https://github.com/Reed-CompBio/spras/pull/340

ntalluri (dismissed):

test/ml/expected/expected-pca-coordinates-negated.tsv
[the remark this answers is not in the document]
tristan-f-r: It only negated one column, like before.

test/ml/test_ml.py
```diff
@@ -88,6 +88,7 @@ def test_pca_remove_empty_pathways(self):
         coord = pd.read_table(OUT_DIR + 'pca-coordinates.tsv')
         coord = coord.round(5)  # round values to 5 digits to account for numeric differences across machines
         expected = pd.read_table(EXPECT_DIR + 'expected-pca-coordinates.tsv')
+        pd.read_table(EXPECT_DIR + 'expected-pca-coordinates-negated.tsv')
```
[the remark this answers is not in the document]
tristan-f-r: It isn't 👍

ntalluri:
The code itself looks good to me. I still need to test the new environment updates locally.

ntalluri (commented):
Code looks good to me. However, I am going to test the environment updates locally before approving.

tristan-f-r:
Okay - sorry about this terribly long hill of a PR for what it was supposed to be. That should be the last commit which fixes the tests from the latest merge :+1:

agitter (approved):
I ran the `TestML` tests locally and they passed

PR #312, opened 2025-07-02
chore: add devcontainer

tristan-f-r:
[I lost track of this devcontainer after messing up my branch switch, so I would just like this to be in SPRAS. This is different from the usual editor config as it supports multiple editors and is not personalized.]

I've been using this to make my PRs to SPRAS and to test. This runs singularity and docker perfectly fine within the devcontainer in VSCode.

 For some extra motivation, [devcontainers](https://containers.dev/) are similar to [nix shells](https://nixos.wiki/wiki/Development_environment_with_nix-shell) [if they worked consistently] and with better first-class editor integration and cloud support. For example, if any potential SPRAS user has trouble setting up SPRAS in their local environment, they could open the repository through any one of the many places that support devcontainers (e.g. VSCode, GitHub Codespaces, or GitPod)


tristan-f-r:
@agitter at the top, you can:

image

open this repository in Codespaces, which will, by default, use the provided devcontainer. Or, if you use `vscode`, you should be able to install the devcontainer extension and open `spras` inside the devcontainer. Whichever one is easier.

PR #313, opened 2025-07-02
test: make sure all inputs exist

tristan-f-r:
This catches the error that occurred in #223 earlier than after the docker container is ran.

### DIFF
--- a/test/generate-inputs/test_generate_inputs.py
+++ b/test/generate-inputs/test_generate_inputs.py
@@ -49,6 +49,6 @@ class TestGenerateInputs:
             exp_file_name = algo_exp_file[algo]
             assert filecmp.cmp(OUTDIR + f"{algo}-{exp_file_name}.txt", EXPDIR + f"{algo}-{exp_file_name}-expected.txt",
                                shallow=False)
-            
+
             for file in filename_map.values():
                 assert Path(file).exists()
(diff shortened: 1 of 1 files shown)


tristan-f-r:
All of the required inputs weren't being generated in the initial commits for #223. This was caught by Snakemake, but not by our tests.

test/generate-inputs/test_generate_inputs.py
```diff
@@ -49,3 +49,6 @@ def test_prepare_inputs_networks(self):
             exp_file_name = algo_exp_file[algo]
             assert filecmp.cmp(OUTDIR + f"{algo}-{exp_file_name}.txt", EXPDIR + f"{algo}-{exp_file_name}-expected.txt",
```
ntalluri, line 50: Would this fail along with the new assert in 53-54 adds? Or was this blindly passing?

tristan-f-r: This would not fail, since we only check against one of the required input files.

tristan-f-r: (Though, generally, we do have a deep weakness against all of these tests, which is that we do no full workflow runs other than the Snakemake run in CI. That run carries a lot of testing weight with it that it really should not have.)

Issue #272, opened 2025-06-12
Push multiple architectures to Docker


tristan-f-r:
When making the RWR PR, we had the following error [in the CI](https://github.com/Reed-CompBio/spras/actions/runs/15616692753/job/43991210548?pr=261):

```
2025-06-12T17:11:19.8730341Z PASSED
2025-06-12T17:11:20.3701315Z test/RWR/test_RWR.py::TestRWR::test_rwr 404 Client Error for http+docker://localhost/v1.48/images/docker.io/reedcompbio/rwr:latest/json: Not Found ("No such image: reedcompbio/rwr:latest")
2025-06-12T17:11:20.3703416Z None
```

However, at the time of writing, there is an image already present:

![Image](https://github.com/user-attachments/assets/34d5cddb-899c-41fe-a9b7-dbe8bab976d3)

It seems that the CI is looking for the `amd64` architecture, but only the `arm` architecture is pushed - it would be useful to have more than just the `amd64` architecture pushed, and to document that `arm`-based machines need to also push an `amd64` architecture.

annaritz:
Having issues with `buildx` - it seems to hang when asking for authenticating. Here are the commands I tried:

1. Make a `docker-container` builder (but I don't set it as default).

```
docker buildx create \
  --name container \
  --driver=docker-container
```

2. Push `rwr` for both `arm64` and `amd64` architectures:

```
docker buildx build \
 --tag reedcompbio/rwr:v1 \
 --file Dockerfile \
 --platform linux/arm64,linux/amd64 \
 --builder container \
 --push .
```

The output where it hangs is below (I've let it run for ~3 min - it should definitely get past credential authorization by then). @tristan-f-r can you try these commands and see if you can push multiple architectures? It could be something with `reedcompbio` (though I can push and pull without any issue). 

```
% docker buildx build \
 --tag reedcompbio/rwr:v1 \
 --file Dockerfile \
 --platform linux/arm64,linux/amd64 \
 --builder container \
 --push .
[+] Building 41.3s (13/14)                                                                                                                 docker-container:container
 => CACHED [linux/arm64 2/4] RUN pip install networkx==2.8.8 numpy==1.26.4 scipy==1.15.2                                                                         0.0s
 => CACHED [linux/arm64 3/4] WORKDIR /RWR                                                                                                                        0.0s
[...]

PR #314, opened 2025-07-03
ci: docker push (-> ghcr)

### LINKED ISSUE #77: Consider migrating containers to GitHub Container registry
Docker killed free team organizations and then reversed its decision, as summarized [here](https://www.docker.com/developers/free-team-faq/). We should consider migrating our containers to the [GitHub Container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) as an alternative to DockerHub in case Docker goes ahead with this change in the future.

tristan-f-r:
This PR makes it easy to push docker images. This also gives us a static list of what architectures we support, closing #272.

Since this is a little more inconvenient to do with the Docker Registry, we continue with the GHCR (closes #77), and create a new workflow that allows maintainers to deploy containers on certain versions.

To make this convenient, a `metadata.json` file is added to every single `docker-wrapper` (should we lint this?) which specifies what the actual desired docker name is.

---

The versioning comments have been addressed. Original comment:

> _Note_: Until SPRAS is out of zerover, should we follow the versioning scheme `v1.0.0-x` for breaking changes to these containers? Or, should we just do `v0.4.1-x` (current version of SPRAS at the time of writing), where `x` is incremented for every breaking change?
> 
> I like the idea of using `v1.0.0-x`, as it indicates that SPRAS should be used once it's stable. I've currently pushed the containers as `v1` (in current convention), but we can change it.


tristan-f-r:
DOMINO and OI2 aren't building on arm64. Can't find the source for DOMINO, but an issue is open related to OI2: #315.

tristan-f-r:
I've decided to allow images to specify architectures, and make some architecture support a burden of an issue instead.

docker-wrappers/AllPairs/metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "dockerName": "allpairs",
```
agitter, line 2: `dockerName` feels slightly off but I'm not sure what else to call this. `image`? `imageName`?

docker-wrappers/DOMINO/metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "dockerName": "domino",
+    "architectures": ["linux/amd64"]
```
agitter, line 3: What dictates whether an image can be built for an architecture? Should that go in the DOMINO readme or in some other documentation?

docker-wrappers/OmicsIntegrator1/Dockerfile
```diff
@@ -24,7 +22,5 @@ RUN commit=0a57ede6beeef6e63b86d19898e560d62015e85d && \
 ENV MSGSTEINER_PATH=/OmicsIntegrator/msgsteiner-1.3/msgsteiner
 WORKDIR /OmicsIntegrator
 
-COPY environment.yml .
```
agitter, line 27: We could delete `environment.yml` if we are no longer using it to build the image in either Dockerfile.

docker-wrappers/OmicsIntegrator1/metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "dockerName": "omics-integrator-1",
```
agitter, line 2: What is the convention for adding hyphens or not?

docker-wrappers/push.py
```diff
@@ -0,0 +1,73 @@
...
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(
+        description="Docker Wrapper builder python wrapper"
+    )
+    parser.add_argument("--dir", type=str, required=True, help="The directory of the docker wrapper to use")
+    parser.add_argument("--version", type=str, required=True, help="The version (tag name) to use for the container")
+    parser.add_argument("--org-name", type=str, help="The organization to push to", default="ghcr.io/reed-compbio/")
+    parser.add_argument("--yes", type=bool, help="Whether to automatically agree to pushing a container", action=argparse.BooleanOptionalAction)
+    parser.add_argument("--relax", type=bool, help="Whether to not be strict on tag naming", action=argparse.BooleanOptionalAction)
```
agitter, line 31: From the message alone I couldn't tell what this argument does.

docker-wrappers/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+This folder contains the list of all docker wrappers.
+
+We build these docker images with `buildx` to support multiple docker architectures.
+
+For convenience, we provide a `push.py` python file which pushes a docker image with the architectures
+that we want to support.
+
+```
+usage: push.py [-h] --tag TAG --dir DIR
```
agitter, line 11: tag should be version

agitter (commented):
I took an initial pass and like @jhiemstrawisc's opinions as well.

This change would mean we are no longer pushing new images to our DockerHub organization. Are there consequences to that? SPRAS doesn't much of an external userbase, so perhaps not yet. If it did, we may be abandoning users who couldn't get new images that newer versions of the SPRAS package required at some point in the future.

In terms of versioning, Justin and I originally discussed how the Docker images aren't tied to a specific SPRAS release. That's why we started an integer versioning convention. There is some implicit relationship where pathway reconstruction algorithm vX is only compatible with SPRAS >= vY. Ideally that would be explicit.

That versioning still makes sense to me because users need to know what version of a pathway algorithm image they are using and may need to change that version independently of the SPRAS package version. Even if we switch registries, I don't think we should reset versions at v1.

I would also like some way to document release notes for each Docker image as we iterate on them. That is currently done in the readmes. Should we keep it there?

Making this change would add another step to the contributing guide.

.github/workflows/push-docker.yml
```diff
@@ -0,0 +1,27 @@
...
+  push-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
tristan-f-r: I believe I tested it on my fork, but I'll make sure :+1:

docker-wrappers/OmicsIntegrator1/metadata.json
tristan-f-r: I believe this was an arbitrary choice, but I preserved it from `build-containers.yml`.

docker-wrappers/AllPairs/metadata.json
tristan-f-r: `imageName` sounds good. 👍

tristan-f-r:
> This change would mean we are no longer pushing new images to our DockerHub organization. Are there consequences to that?

As you've mentioned, this would only affect current users that have not changed their configured container registry to be `ghcr.io` but update SPRAS.

> That versioning still makes sense to me because users need to know what version of a pathway algorithm image they are using and may need to change that version independently of the SPRAS package version.

I don't believe this case would happen, but I do see the rest of the motivation for decoupling the docker image versions from the SPRAS versions - also possibly because seeing a different SPRAS version appear in errors for docker containers can be misleading.

> I would also like some way to document release notes for each Docker image as we iterate on them. That is currently done in the READMEs. Should we keep it there?

This is fine, we just need to be consistent about it. It might be a good idea to begin making all of the container READMEs consistent in the first place.

agitter:
This is making nice progress. I'd like to talk to Justin in our next in person meeting before I give my next round of comments.

tristan-f-r:
#386 revealed that `Docker_no_conda` from main hasn't been building, since Buster, the latest image version for Python 2.7 required by OI1, is deprecated. This has been fixed in the latest commit, and pushed to GHCR as OI1 v2.

tristan-f-r:
Did SPRAS ever work on ARM core machines? The currently uploaded docker hub containers [with only a few exceptions thanks to #272] are all `amd64` arch.

docs/contributing/index.rst
```diff
@@ -421,7 +421,7 @@ General steps for contributing a new pathway reconstruction algorithm
     ``<algorithm>``, and include any other files required to build that
     image in the subdirectory
 3.  Build and push the Docker image to the
-    `reedcompbio <https://hub.docker.com/orgs/reedcompbio>`__ Docker
+    `reed-compbio <https://hub.docker.com/orgs/reed-compbio>`__ Docker
```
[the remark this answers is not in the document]
tristan-f-r: This was. I believe I never decided what to do about the 'push to org' section in the contributing guide, but it seems best to only quickly mention `--push`.

docker-wrappers/OmicsIntegrator2/metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "imageName": "omics-integrator-2",
+    "architectures": ["linux/amd64"]
```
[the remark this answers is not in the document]
tristan-f-r: (I was wary of this since python doesn't seem to support jsonc in their stdlib - I do absolutely agree that there should be info about this in the README, which I believe I've already added.)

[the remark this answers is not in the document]
tristan-f-r: I do: I consider them bugs (#315) and not an innate failure of the wrapped code. [They are in the TODO section of the container READMEs](https://github.com/Reed-CompBio/spras/pull/314/files#diff-02777589c892dd87783d0b60ba42f61c4c4ad5a9462764809130c58a6d6f57b8R33).

.github/workflows/push-docker.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Run push.py
+        run: python3 docker-wrappers/push.py --dir ${{ inputs.folder }} --version ${{ inputs.version }} --yes
```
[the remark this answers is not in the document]
tristan-f-r: With the `metadata.json` file, I'm more in favor of maintaining some local code instead of having to maintain code only testable via GitHub actions. Though, it's still not optimal.

PR #326, opened 2025-07-11
test: dataset

tristan-f-r:
Adds dataset tests borrowed from #255. Also refactors `Dataset`.

### DIFF
--- a/spras/config/dataset.py
+++ b/spras/config/dataset.py
@@ -1,4 +1,3 @@
-import os
 from typing import Annotated
 
 from pydantic import AfterValidator, BaseModel, ConfigDict


agitter:
Are we sure all algorithms must have edge weights between 0 and 1? All pairs shortest paths uses networkx's shortest_path function, which runs Dijkstra's algorithm by default. That should work for non-negative edge weights.

agitter:
Okay. I'm totally sure what the scope of this PR is without figuring it out from the code. I first tried inferring it from the commit messages, but that led me to the wrong conclusion.

agitter:
@tristan-f-r before reviewing this, I would like to have the overall scope and goal of this pull request written out to guide me.

tristan-f-r:
Right - the description has changed since I first wrote this. This PR is doing a number of things now:

- Dataset parsing tests
- Makes Dataset take in a `DatasetSchema` instead of a duplicated list of keys from `DatasetSchema`. This makes the API more annoying.
- Adds new path types for file paths, file descriptors, or open file links (this change isn't motivated here, but later, I would like to be able to pass arbitrary file handles in certain functions instead of having to pass a path). Originally, I wanted to use Python's `typeshed`, but you have to do a hacky import trick that seemed worse then just explicitly declaring the path type aliases.

spras/dataset.py
```diff
@@ -1,57 +1,49 @@
...
+from typing import Union
 
 import pandas as pd
 
+from spras.config.dataset import DatasetSchema
+from spras.util import FileLike, open_weak
+
 """
 Author: Chris Magnano
 02/15/21
```
agitter: We could strip this. It isn't in any other files.

spras/config/dataset.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+
+    # We prefer AfterValidator here to allow pydantic to run its own
+    # validation & coercion logic before we check it against our own
+    # requirements
+    label: Annotated[str, AfterValidator(label_validator("Dataset"))]
+    node_files: list[str | os.PathLike]
+    edge_files: list[str | os.PathLike]
+    other_files: list[str | os.PathLike]
+    data_dir: str | os.PathLike
```
agitter: Can these all be `LoosePathLike`? My understanding is that its purpose is to remove syntax link this.

spras/dataset.py
```diff
@@ -1,57 +1,46 @@
...
-        """
-        with open(file_name, "wb") as f:
+    def to_file(self, file: LoosePathLike):
+        """Saves dataset object to pickle file"""
+        with open(file, "wb") as f:
             pkl.dump(self, f)
 
     @classmethod
-    def from_file(cls, file_name: str):
+    def from_file(cls, file: Union[LoosePathLike, "Dataset"]):
```
agitter: Why is Dataset a string here?

spras/runner.py
```diff
@@ -78,12 +79,12 @@ def prepare_inputs(algorithm: str, data_file: str, filename_map: dict[str, str])
...
+def parse_output(algorithm: str, raw_pathway_file: LoosePathLike, standardized_pathway_file: LoosePathLike, params: Mapping[str, Any]):
     """
     Convert a predicted pathway into the universal format
     @param algorithm: algorithm name
     @param raw_pathway_file: pathway file produced by an algorithm's run function
     @param standardized_pathway_file: the same pathway written in the universal format
     """
     algorithm_runner = get_algorithm(algorithm)
-    return algorithm_runner.parse_output(raw_pathway_file, standardized_pathway_file, params)
+    return algorithm_runner.parse_output(str(raw_pathway_file), str(standardized_pathway_file), params)
```
agitter: Do these need to be `str` instead of `LoosePathLike`? If that is a reasonable change but outside of the scope of this dataset-centric pull request, let's keep this fixed scope and note the possible follow up somewhere.

PR #286, opened 2025-06-23
docs(rtd): prms

tristan-f-r:
My main motivation for this PR was to be able to link the algorithm-specific pages in spras-benchmarking to extra context pages.

This adds some light documentation to each PRM - since OI1 and OI2 have no parameter documentation, I've avoided talking about their parameters.

PR #285, opened 2025-06-23
refactor: redo PRM properties

tristan-f-r:
After understanding the consequences of `@property` on ABC, moving the property handling over to `__init_subclass__` helps correctly address #222. To demonstrate this, we add a DOI property, which can be used in pursuit of #5.

Generally, this patch gives us actual property errors and type-hints on properties,

PR #329, opened 2025-07-14
feat!: typed PRA#run

### LINKED ISSUE #296: [config] nested `runs`
Currently, runs are specified per-algorithm as so:

```yaml
run1:
  a: 5
  b: 6
run2:
  a: 4
  b: 5
```

Instead, we should organize this to be nested inside `runs` for better compatibility with keys we may add to algorithm (e.g. `include`) in the future.

```yaml
runs:
  run1:
    a: 5
    b: 6
  run2:
    a: 4
    b: 5
```

This has parsing (no extraneous key support) and readability benefits.

### LINKED ISSUE #297: `containers`
Config file options should be under a general `containers` rule, so we can not only specify it per config, but also per algorithm and per analysis.

Essentially, moving:

```yaml
container_framework: docker
unpack_singularity: false
container_registry:
  base_url: docker.io
  owner: reedcompbio
```

to:

```yaml
containers:
  framework: docker
  unpack_singularity: false
  registry:
    base_url: docker.io
    owner: reedcompbio
```

### LINKED ISSUE #321: Typed PRA#run
Depending on #292 (and #296 would be nice for this, but not required), this handles nondeterministic algorithm collisions, type issues with the PRA class related to PRA#run, and better documents defaults in the generated `parameters.yml`.

In every single PRA method implementation, we should have a [pydantic BaseModel](https://docs.pydantic.dev/latest/api/base_model/) which documents the algorithm parameters separately from the associated `#run` method. This gets rid of the need to use `**kwargs` for passing in `run` options.

For nondeterministic algorithms, we can have a parameter, `_time`, which defaults to the current time. (pydantic ignores fields beginning with an underscore): this parameter will still make it to before the parameters are hashed.

This has the final benefit of, in combination with #292, allowing for a fully featured configuration schema with all of the options it takes.

### LINKED ISSUE #323: Guaranteed immutable output
To support OSDF, all of our file names must be unique.

Adding this for deterministic algorithms is easier: In #320, we make all SPRAS files change their name whenever the SPRAS version is bumped.

For nondeterministic algorithms, we need more changes:

- #329, which allows SPRAS algorithm arguments to be serialized and deserialized.
- #332, which adds a SPRAS CLI.

This is all used in the #335 seeds PR, which will (when `spras prepare` [subcommand name TBD] from the SPRAS CLI is used) serialize the entire SPRAS configuration, including the seed parameters, to a new file, which can then be run with `spras run`. Since seeds now become a part of the SPRAS configuration when being ran, that guarantees us file immutability specifically for algorithms.

See #335.

tristan-f-r:
> [!NOTE]
> This also contains changes that close #297 to resolve a dependency diamond. I was debating about splitting that to another PR, but that 'containers' PR would depend on #292 (giving this 2 degrees of dependency), and the types in PRA#run actually make that change easier to follow and motivate.

- [x] Depends on #285 
- [x] Depends on #292
- [x] Depends on #287
- [x] Needs #389 for tests to pass
- [x] Depends on #387
- [x] Depends on #390

This also borrows (not a direct dependency) from #286. #286 should be merged after #329 is merged, as there's a good chance that we can just auto-generate documentation from the documentation here instead.

Closes #321, closes #296, and closes #297.

Arguments are now specified as a pydantic BaseModel with attached documentation:

```py
class DominoParams(BaseModel):
    module_threshold: Optional[float]
    "the p-value threshold for considering a slice as relevant (optional)"

    slice_threshold: Optional[float]
    "the p-value threshold for considering a putative module as final module (optional)"

    model_config = ConfigDict(use_attribute_docstrings=True)
```

When constructing a `PRM`, this is passed in as a generic:

```py
class DOMINO(PRM[DominoParams]):
```

For algorithms that don't specify parameters, the `Empty` type is preferred instead (whose signature is the empty `BaseModel` with an attached `model_config` that doesn't allow any other parameters.

This also changes the signature of `PRA#run` (reflecting the PR title) to be:

```py
# Where T is the TypedVar which must be a pydantic.BaseModel
def run(inputs: dict[str, str | os.PathLike], output_file: str | os.PathLike, args: T, container_settings: ProcessedContainerOptions):
```

This has the disadvantage that `inputs` no longer has code completion, but this was probably something to not hide from the developer-user anyway, as we were passing inputs in via kwargs. See:

```py
# This PR is more verbose when passing in arguments, which is a problem for people who
# are directly using PRA#run. However, I don't care about this audience.
PathLinker.run({"nodetypes": TEST_DIR+'input/sample-in-nodetypes.txt',
                "network": TEST_DIR+'input/sample-in-net.txt'},
                output_file=OUT_FILE_100,
                args=PathLinkerParams(k=100))
```

All of this gives us:
- Encouraged, easily parsable PRM argument documentation
- Parameter validation
- a fully-specified JSON schema if #358 
- default factories, which are used to fix nondeterminism (#335).
- Parameter types, to be used for parameter tuning for automatically determining what parameters to select
  - (and AST-based range parsing for us to easily divide the step size for parameter tuning)

### For reviewers

`algorithms.py` is the only "dense code" in this PR. That file extends our previous `eval` system and includes some hard-to-follow workflow code that enables type checking for our config file.

Otherwise, most of this is just refactoring all of the `PRA#run` calls to meet the new signature, specifying the new parameter models, and adding more parameter documentation.


tristan-f-r:
I'm going to split this PR into two PRs - this one will just add the parameter configs, then I'll make a new PR that adds `seed` and `_time`. That way, I can get this PR to fully pass CI checks.

config/schema.json
ntalluri, line 1: Is there a way to auto generate this file?

tristan-f-r: #358

spras/pathlinker.py
```diff
@@ -10,7 +13,13 @@
 from spras.prm import PRM
 from spras.util import duplicate_edges, raw_pathway_df
 
-__all__ = ['PathLinker']
+__all__ = ['PathLinker', 'PathLinkerParams']
+
+class PathLinkerParams(BaseModel):
+    k: int = 100
```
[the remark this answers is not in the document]
ntalluri: > If a user were to specify pathlinker include:true, but not include any k's for the parameters, would this default SPRAS to run pathlinker with k = 100?

I'm more curious if SPRAS would somehow allow this because a default does exist (I'm assuming that the config parser wouldn't let this happen).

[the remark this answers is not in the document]
tristan-f-r: Right! Okay, I forgot that this was detached from the context of this PR: I wanted to preserve defaults for making sure that configurations preserved as much information as possible across runs.

PathLinker's internal algorithm has `k=100` as a default, [but we currently let the internal algorithm handle defaults](https://github.com/Reed-CompBio/spras/blob/c3b02cd55fbc918c5f23959528e06e16c4301194/spras/pathlinker.py#L112-L113).

Because the nondeterminism seeding serializes defaults, it seemed like a nice feature to also know what the parameter defaults were at the time the configuration was ran.

Ignoring nondeterminism, these defaults have no effect.

docs/contributing/index.rst
```diff
@@ -242,7 +244,9 @@ the format ``<vertex1>|<vertex2>``, which also differs from the
 ``omicsintegrator1.py`` example. ``spras/dataset.py`` provides functions
 that provide access to node information and the interactome (edge list).
 
-Implement the ``run`` function, following the PathLinker example. The
+Implement the ``run`` function, following the PathLinker example. Since
+there are no arguments, you do not need to instantiate a new instance of ``Empty``
+if parameters are not specified. The
```
ntalluri:
```suggestion
if parameters are not specified.
```

docs/contributing/index.rst
```diff
@@ -242,7 +244,9 @@ the format ``<vertex1>|<vertex2>``, which also differs from the
 ``omicsintegrator1.py`` example. ``spras/dataset.py`` provides functions
 that provide access to node information and the interactome (edge list).
 
-Implement the ``run`` function, following the PathLinker example. The
+Implement the ``run`` function, following the PathLinker example. Since
```
ntalluri: Should this follow a different example (like AllPairs)?

spras/config/algorithms.py
```diff
@@ -0,0 +1,161 @@
...
+
+    **Safety Note**: This does not prevent availability attacks: this can still exhaust
+    resources if wanted. This only prevents secret leakage.
+    """
+
+    if not isinstance(value, str):
+        return value
+
+    # These strings are in the form of function calls `function.name(param1, param2, ...)`.
+    # Since we want to avoid `eval` (since this might be running in the secret-sensitive HTCondor),
```
ntalluri: Could you remind me why we want to avoid eval?

spras/config/algorithms.py
```diff
@@ -0,0 +1,161 @@
...
+    Whether the passed in type can have any numpy helpers.
+    This is mainly used to provide hints in the JSON schema.
+    """
+    allowed_types = (int, float)
+
+    # check basic types, then check optional types
+    return type in allowed_types or \
+        any([arg for arg in get_args(type) if arg in allowed_types])
+
+def python_evalish_coerce(value: Any) -> Any:
```
ntalluri: Could you add a more detailed comment on what is happening in this function and why we are using this. This way someone doesn't need to read the code fully to understand what is happening.

spras/config/algorithms.py
ntalluri: When is this file used? When we are reading in the config using config.py?

spras/config/container_schema.py
```diff
@@ -33,6 +33,8 @@ class ContainerRegistry(BaseModel):
 class ContainerSettings(BaseModel):
     framework: ContainerFramework = ContainerFramework.docker
     unpack_singularity: bool = False
+
+    model_config = ConfigDict(extra='forbid')
```
ntalluri: Why is model_config included in this file / for container settings in general?

spras/config/config.py
```diff
@@ -152,58 +147,29 @@ def process_algorithms(self, raw_config: RawConfig):
...
-            cur_params = alg.params
-            if cur_params.include:
+            if alg.include:
                 # This dict maps from parameter combinations hashes to parameter combination dictionaries
                 self.algorithm_params[alg.name] = dict()
             else:
                 # Do not parse the rest of the parameters for this algorithm if it is not included
                 continue
 
-            if cur_params.directed is not None:
```
ntalluri: Did we push the PR that makes a nice error when this is key is not provided?

spras/meo.py
```diff
@@ -66,6 +70,21 @@ def write_properties(filename=Path('properties.txt'), edges=None, sources=None,
 
         # Do not need csp.phase, csp.gen.file, or csp.sol.file because MAXCSP is not supported
 
+class MEOParams(BaseModel):
+    max_path_length: Optional[int] = None
+    "the maximal length of a path from sources and targets to orient."
+
+    local_search: Optional[bool] = None
```
ntalluri: Why is this a bool, but above it mentions it is a yes no parameter.

spras/mincostflow.py
```diff
@@ -61,35 +73,24 @@ def generate_inputs(data, filename_map):
                      header=False)
 
     @staticmethod
-    def run(sources=None, targets=None, edges=None, output_file=None, flow=None, capacity=None, container_settings=None):
-        """
-        Run min cost flow with Docker (or singularity)
```
ntalluri: same comment as before

spras/pathlinker.py
```diff
@@ -11,7 +13,13 @@
...
+    "path length"
```
ntalluri: Is this top k number of paths, what is path length in this context?

spras/pathlinker.py
```diff
@@ -65,34 +73,21 @@ def generate_inputs(data, filename_map):
         edges.to_csv(filename_map["network"],sep="\t",index=False,columns=["Interactor1","Interactor2","Weight"],
                      header=["#Interactor1","Interactor2","Weight"])
 
-    # Skips parameter validation step
     @staticmethod
-    def run(nodetypes=None, network=None, output_file=None, k=None, container_settings=None):
-        """
-        Run PathLinker with Docker
```
ntalluri: Same comment as before

spras/prm.py
```diff
@@ -47,3 +83,35 @@ def validate_required_inputs(cls, filename_map: dict[str, str]):
...
+
+    @classmethod
+    def validate_required_run_args(cls, inputs: dict[str, str | os.PathLike], relax: Optional[list[str]] = None):
+        """
+        Validates the `inputs` parameter for `PRM#run`.
+
+        @param inputs: See `PRM#run`.
+        @param relax: List of inputs that aren't required: if they are specified, they should be valid path
+        """
+        if not relax: relax = []
```
ntalluri: What is the relax parameter supposed to do? I'm confused when someone would have a list of inputs that aren't required; are these all optional parameters?

spras/rwr.py
```diff
@@ -38,12 +49,11 @@ def generate_inputs(data, filename_map):
         edges.to_csv(filename_map['network'],sep='|',index=False,columns=['Interactor1','Interactor2'],header=False)
 
     @staticmethod
-    def run(network=None, nodes=None, alpha=None, output_file=None, container_settings=None, threshold=None):
```
ntalluri: same comment as before

ntalluri (commented):
This is my first pass of this PR. 

Could you add an example demonstrating how these files and components fit together for an example algorithm and its configuration file, and show how the new files are used for validation and execution?

Also, for someone looking to integrate a new algorithm into SPRAS, what details or requirements should they be aware of? I’m assuming the new key piece would involve the Pydantic models.

spras/prm.py
```diff
@@ -1,11 +1,16 @@
...
 
+from pydantic import BaseModel
+
+from spras.config.container_schema import ProcessedContainerSettings
 from spras.dataset import Dataset
 
+T = TypeVar('T', bound=BaseModel)
 
-class PRM(ABC):
+class PRM(ABC, Generic[T]):
```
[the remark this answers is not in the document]
tristan-f-r: I'll write a doccomment here, but it's how we pass the typed parameter object over: e.g. in `DOMINO(PRM[DominoParams]):`, `DominoParams` is `T` and `Generic` allows us to specify that as so.

spras/pathlinker.py
tristan-f-r: ([see parent comment](https://github.com/Reed-CompBio/spras/pull/329#discussion_r2550582017))

spras/rwr.py
tristan-f-r: ([see parent comment](https://github.com/Reed-CompBio/spras/pull/329#discussion_r2550582017))

spras/prm.py
tristan-f-r: [We use this in OmicsIntegrator1](https://github.com/Reed-CompBio/spras/pull/329/files#diff-eed6dce68f02179a25278ef3edc5cd193fcbae298acc1f5969cd5e69dd408e47R161) for the optional dummy node file.

tristan-f-r:
> Could you add an example demonstrating how these files and components fit together for an example algorithm and its configuration file, and show how the new files are used for validation and execution?

I'm confused by this question: we do this with the present algorithms.

Snakefile
```diff
@@ -273,16 +272,12 @@ rule reconstruct:
...
-        params['output_file'] = output.pathway_file
-        # Remove the default placeholder parameter added for algorithms that have no parameters
-        if 'spras_placeholder' in params:
-            params.pop('spras_placeholder')
-        params['container_settings'] = container_settings
-        runner.run(wildcards.algorithm, params)
+        # Declare the input files as a dictionary.
+        inputs = dict(zip(runner.get_required_inputs(wildcards.algorithm), *{input}, strict=True))
+        # Remove the _spras_run_name parameter added for keeping track of the run name for parameters.yml
+        if '_spras_run_name' in params:
```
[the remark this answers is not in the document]
tristan-f-r: This is defined in `config.py` for testing the configuration.

docker-wrappers/SPRAS/example_config.yaml
agitter: Why don't we make the full set of formatting changes to this config file as well?

spras/config/config.py
```diff
@@ -152,58 +147,29 @@ def process_algorithms(self, raw_config: RawConfig):
...
-                                    obj = [True]
-                                elif obj.lower() == "false":
-                                    obj = [False]
-                                else:
-                                    # Catch-all for strings
-                                    obj = [obj]
-                            if not isinstance(obj, Iterable):
-                                raise ValueError(f"The object `{obj}` in algorithm {alg.name} at key '{p}' in run '{run_params}' is not iterable!") from None
-                        all_runs.append(obj)
+                run_subscriptable = vars(runs[run_name])
```
agitter: I don't see `vars` used very frequently so a comment about the structure of `run_subscriptable` would help readability if we return to this in the distant future. I understand it today.

spras/config/util.py
```diff
@@ -17,3 +26,18 @@ def _missing_(cls, value: Any):
...
+# https://github.com/yaml/pyyaml/issues/722#issue-1781352490
+yaml.SafeDumper.add_multi_representer(
+    CaseInsensitiveEnum,
+    yaml.representer.SafeRepresenter.represent_str,
+)
+
+class Empty(BaseModel):
+    """
+    The empty base model. Used for specifying that an algorithm takes no parameters,
+    yet are deterministic.
```
agitter:
```suggestion
    yet is deterministic.
```
Is this the correct phrasing?

spras/runner.py
```diff
@@ -1,46 +1,53 @@
...
-    "mincostflow": MinCostFlow,
-    "omicsintegrator1": OmicsIntegrator1,
-    "omicsintegrator2": OmicsIntegrator2,
-    "pathlinker": PathLinker,
-    "responsenet": ResponseNet,
-    "rwr": RWR,
-    "strwr": ST_RWR,
+# Algorithm names to a three-tuple of (PRM, BaseModel, default BaseModel or None if there are no good defaults).
+# This is used for the configuration and to fetch algorithms during reconstruction
+algorithms: dict[str, tuple[type[PRM], type[BaseModel], Optional[BaseModel]]] = {
```
agitter: This gets pretty awkward

agitter (commented):
I can follow the overall design goals. My big picture takeaway is that I can see why we need these changes, but the typing adds indirection and hurts code readability. I don't have a solution for that.

I haven't looked at every pathway reconstruction algorithm and test case update yet, so I'll need to take at least one more pass. I wanted to leave some initial comments.

`algorithms.py` has some sophisticated Python. I am fairly sure I understand it when reviewing it today. I'm not sure about my ability to troubleshoot things if/when they break.

docs/contributing/index.rst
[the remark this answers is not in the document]
tristan-f-r: AllPairs is the new example - I'll drop the mentions to PathLinker.

spras/prm.py
```diff
@@ -30,11 +35,44 @@ def __init_subclass__(cls):
...
+        T_class: type[T] = get_args(cast(Any, cls).__orig_bases__[0])[0]
+
+        # Since we just used reflection, we provide a mountain-dewey error message here
+        # to protect against any developer confusion.
+        if not issubclass(T_class, BaseModel):
+            raise RuntimeError("The generic passed into PRM is not a pydantic.BaseModel.")
+
+        # (and pydantic already provides nice error messages, so we don't need to worry about
+        # catching this.)
+        T_parsed = T_class.model_validate(args)
```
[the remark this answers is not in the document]
tristan-f-r: Mentioned in a comment :+1:

spras/prm.py
```diff
@@ -30,11 +35,44 @@ def __init_subclass__(cls):
...
+    @classmethod
+    def run_typeless(cls, inputs: dict[str, str | os.PathLike], output_file: str | os.PathLike, args: dict[str, Any], container_settings: ProcessedContainerSettings):
+        """
+        This is similar to PRA.run, but it does pydantic logic internally to re-validate argument parameters.
+        """
+        # awful reflection here, unfortunately:
+        # https://stackoverflow.com/a/71720366/7589775
+        # alternatively, one could have a T_class parameter
+        # for PRA here, but this level of implicitness seems alright.
+        T_class: type[T] = get_args(cast(Any, cls).__orig_bases__[0])[0]
```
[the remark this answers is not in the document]
tristan-f-r: I made it more clear in a comment by characterizing what it does rather than its implementation: the result is one of the purpose of `__orig_bases__`.

Since we're on Python 3.11, we have to use the internal `__orig_bases__` instead of the now public `types#get_original_bases` API, which makes this code a little more confusing. (I've also mentioned this in a comment.)

docs/contributing/index.rst
```diff
@@ -117,17 +117,13 @@ Step 2: Create a Local Neighborhood Docker image
 
 Complete the ``Dockerfile`` in the
 ``docker-wrappers/LocalNeighborhood``
-directory to create a Docker image. The PathLinker ``Dockerfile``
+directory to create a Docker image. The AllPairs ``Dockerfile``
```
agitter: The AllPairs Dockerfile does not actually demonstrate `COPY` so we will need to provide a different example or update this. COPY is needed for LocalNeighborhood.

agitter (commented):
My only new comments are small.

In the spirit of helping new contributors who encounter this code base in the future, I'm wondering where to capture some of the information about the overall SPRAS design, especially the part that is changing here. Some of the useful information in the original message of this PR, e.g.
> Arguments are now specified as a pydantic BaseModel with attached documentation:

provides a guide to how SPRAS works and where to find things. Is there a place to retain that knowledge in the repo? Scrolling through individual files to reconstruct it is going to get harder and hard.

agitter (approved):

Issue #324, opened 2025-07-11
Can not install yamlfmt in environment.yml in Windows


tristan-f-r:
> @ntalluri did this work for you after #305? I installed the latest conda env and now get:
> 
> ```
> [INFO] Installing environment for https://github.com/google/yamlfmt.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Restored changes from C:\Users\agitter\.cache\pre-commit\patch1752252703-32208.
> An unexpected error has occurred: CalledProcessError: command: ('C:\\Users\\agitter\\.conda\\envs\\spras\\bin\\go.EXE', 'install', './...')
> return code: 2
> expected return code: 0
> stdout: (none)
> stderr:
>     go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
> ```
> 
> It looks like the conda installation gave me `go.exe` and `gofmt.exe` on Windows but not the full go SDK. I reopening this because Go is broken on Windows for me.
> 
> This appears to be an unresolved issue https://github.com/conda-forge/go-feedstock/issues/203 

 _Originally posted by @agitter in [#295](https://github.com/Reed-CompBio/spras/issues/295#issuecomment-3063067404)_

Extra note from @ntalluri: This was not caught in the windows "Test conda environment" workflow because it already comes with Go.

agitter:
Until the upstream issue is resolved, Windows developers can follow these steps:
- `conda activate spras`
- `conda uninstall go` (confirm)
- Install go following instructions at https://go.dev/doc/install
- Test installation with `go version` and `go env`

ntalluri:
If we rollback #305, we should also rollback #239.

ntalluri:
We might need to do this before the tutorial.

PR #332, opened 2025-07-15
feat: cli

### LINKED ISSUE #410: SPRAS CLI
Instead of putting everything inside Snakemake, having a wrapping CLI allows for separation of the SPRAS workflow with SPRAS-configuration-generating procedures.

This is important for nondeterminism (we want a way to be able to seed configurations, which is nearly impractical to do within Snakemake) and for tuning (more practical to do in Snakemake, which requires less code, but it's relatively untestable.)

Implemented by #332.

tristan-f-r:
This PR adds a very naive SPRAS cli wrapper over Snakemake. Having a wrapping command line interface instead that wraps around Snakemake allows

- generating configuration files without needing to use Snakemake, for automated parameter tuning and seedable configurations in #329. This can be done with parameters inside the configuration, but this requires that we have a Snakemake run that immediately stops.
- SPRAS to be installed as a pip module, so SPRAS binaries can be attached with release versions and pip-installable
  - This means users no longer have to clone SPRAS and install conda to use SPRAS, alleviating #324.

After the next SPRAS release cycle, one could do:

```bash
pip install spras
spras run --configfile config.yaml --cores 8
```

We use the `run` subcommand to make extending this CLI easier later. We can easily supply a `spras --version` or a `spras --help` which links to the RTD, a `spras tune` for parameter tuning, etc...

---

This does make local development more annoying / nearly impossible if you install SPRAS without `python -m pip install -e .[dev]`. The pixi PR (#382) helps with this.

We prefer this over Snakemake since it's easier to test separate components outside of Snakemake.


ntalluri:
Does PR allow for the old version (the conda env version) to still exist or does this PR get rid of it entirely?

Also does this change how Justin set up the CHTC integration?

tristan-f-r:
This CLI should no longer be helpful for parameter tuning - only for determinism and packaging.

ntalluri:
Still in limbo if we need CLI for parameter tuning; next SPRAS meeting we should finalize this.

PR #335, opened 2025-07-17
feat: seeds

### LINKED ISSUE #174: Support running an algorithm multiple times with the same parameter combination
Omics Integrator 1 is sometimes run with the same parameter combination many times when r is nonzero, which introduces edge noise. That is not currently possible in SPRAS. 

We could add a new parameter type in the config file that is the number of times to run with that parameter combination. The config parser would expand that into a unique integer counter.

This could be useful for any non-deterministic algorithm.

### LINKED ISSUE #323: Guaranteed immutable output

tristan-f-r:
## Do not review this. The actual changes this PR makes are hard to see. See the two PRs below.

- [ ] Depends on #332.
- [ ] Depends on #329.

Closes #323. Closes #174.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -25,7 +25,7 @@ algorithm_params = _config.config.algorithm_params
 algorithm_directed = _config.config.algorithm_directed
 pca_params = _config.config.pca_params
 hac_params = _config.config.hac_params
-FRAMEWORK = _config.config.container_framework
+FRAMEWORK = _config.config.container_settings.framework
 
 # Return the dataset or gold_standard dictionary from the config file given the label
 def get_dataset(_datasets, label):
--- a/spras/config/schema.py
+++ b/spras/config/schema.py
@@ -117,11 +117,27 @@ class ReconstructionSettings(BaseModel):
     model_config = ConfigDict(extra='forbid')
 
 class RawConfig(BaseModel):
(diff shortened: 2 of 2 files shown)


agitter:
Why was it automatically closed? All the commits from this branch also belonged to the branch of the merged PR? I don't recall seeing that behavior before.

tristan-f-r:
> Why was it automatically closed? All the commits from this branch also belonged to the branch of the merged PR? 

They did: I designed this PR and split it up into the other PRs (e.g. in #329 you can find commits where I undo algorithm seeding. I can actually isolate these changes into two PRs: (currently making))

PR #345, opened 2025-07-22
feat: algorithm attributions

### LINKED ISSUE #5: Pathway reconstruction algorithm attribution
We should think about how to provide credit to the original pathway reconstruction algorithm authors if their methods are run using our workflow and Docker images.  One idea would be to include citation information in the algorithm class.  Then we could print or log the recommended citation(s) when we call the `run` function.

That wouldn't work for providing attribution if someone uses the Docker image outside of our workflow.  An alternative would be to print citation information as part of the Docker entrypoint.

tristan-f-r:
Closes #5.

I'm a little wary of using the `.bib` files as artifacts (they are very easily reproducible, though, in a similar fashion to our other generate-inputs/parse-outputs tests) because it does feel redundant, but I did want to leave that in, in case this actually is very useful to look for DOI correctness in future algorithm implementations.

---

This conflicts with #292, but since it's a small change, it doesn't really matter which matter it's merged - just mentioning it for a heads up.

More importantly, though, this does need to change depending on when #320 is merged, since this needs to have `attribution-{rev}` attached to it.

> [!NOTE]
> I don't know what the best practices are for bibtex? If there's formatting suggestions, I'm very open to them :+1:


spras/attribution.py
agitter: I haven't done a full review of this, but I would advise we rely more on existing packages for this functionality instead of implementing and maintaining our own.

tristan-f-r: Unfortunately, the only tool (that I'm aware of) which does what we want is [doi2bib](https://github.com/bibcure/doi2bib) which is a CLI.

PR #347, opened 2025-07-23
chore: typos

tristan-f-r:
Adds the `typos` precheck and ignores data files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,7 @@ repos:
     rev: v0.17.0
     hooks:
       - id: yamlfmt
+  - repo: https://github.com/crate-ci/typos
+    rev: v1.34.0
+    hooks:
+      - id: typos
--- a/spras/domino.py
+++ b/spras/domino.py
@@ -151,10 +151,10 @@ class DOMINO(PRM):
         out_modules_dir = Path(out_dir, 'active_genes')
 
         # Concatenate each produced module HTML file into one file
-        with open(output_file, 'w') as fo:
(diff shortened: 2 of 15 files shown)


spras/domino.py
```diff
@@ -151,10 +151,10 @@ def run(network=None, active_genes=None, output_file=None, slice_threshold=None,
         out_modules_dir = Path(out_dir, 'active_genes')
 
         # Concatenate each produced module HTML file into one file
-        with open(output_file, 'w') as fo:
+        with open(output_file, 'w') as f:
```
agitter, line 154: Is this the only variable name `typos` recommended changing? I generally would want us to be able to use any variable names we want.

agitter (commented):
I haven't used this pre-check before. Did it give you a lot of false positives you had to suppress, or the changes here were the only words it flagged?

I have used other spellcheck linters, and it took manual effort to maintain a separate list of words that should not be flagged as misspelled, which would be too much effort here.

.pre-commit-config.yaml
```diff
@@ -28,3 +28,7 @@ repos:
     rev: v0.17.0
     hooks:
       - id: yamlfmt
+  - repo: https://github.com/crate-ci/typos
```
agitter, line 31: Do we need to add the package to our environment?

agitter (commented):
Noting in case it becomes a problem later that we can add a [`_typos.toml`](https://github.com/crate-ci/typos/blob/master/docs/reference.md) file to control false positives: https://github.com/crate-ci/typos?tab=readme-ov-file#false-positives

PR #350, opened 2025-07-24
docs: responsenet

tristan-f-r:
Follow-up to #241 with #286.

### DIFF
--- a/docs/prms/responsenet.rst
+++ b/docs/prms/responsenet.rst
@@ -1,11 +1,11 @@
 ResponseNet
 ===========
 
-ResponseNet is a pathway reconstruction algorithm which finds subnetworks using integer linear programming,
+ResponseNet is a pathway reconstruction algorithm which finds subnetworks using linear programming,
 considering an interactome as a way to send 'flow' from sources to targets.
 See the `original paper <https://doi.org/10.1038/ng.337>`_ and the corresponding reimplementation of it:
 https://github.com/Reed-CompBio/ResponseNet.
 
 ResponseNet takes one optional parameter:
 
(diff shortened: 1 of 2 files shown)


docs/prms/responsenet.rst
```diff
@@ -0,0 +1,11 @@
...
+===========
+
+ResponseNet is a pathway reconstruction algorithm which finds subnetworks using integer linear programming,
+considering an interactome as a way to send 'flow' from sources to targets.
+See the `original paper <https://doi.org/10.1038/ng.337>`_ and the corresponding reimplementation of it:
+https://github.com/Reed-CompBio/ResponseNet.
+
+ResponseNet takes one optional parameter:
+
+* gamma: (int) the size of the output graph: more gamma means more 'flow' gets passed along starting from the sources.
```
agitter, line 11: I would say this "controls" the size of the output, because otherwise it could be misinterpreted as the output size in units of nodes or edges.

docs/prms/responsenet.rst
```diff
@@ -0,0 +1,11 @@
+ResponseNet
+===========
+
+ResponseNet is a pathway reconstruction algorithm which finds subnetworks using integer linear programming,
```
agitter, line 4: I'm I reading the [source](https://github.com/Reed-CompBio/ResponseNet/blob/1eb9b2685af218a28af38dbcda6e5a5abca8d682/responsenet.py#L281) correctly, it uses linear programming, not integer linear programming. Is that correct? GLOP as described [here](https://developers.google.com/optimization/lp/lp_example) appears to be an LP solver.

agitter (commented):
Some nitpicky comments about the optimization

agitter (approved):

PR #356, opened 2025-07-25
chore: run pre-commit on all files

agitter:
After updating my conda environment I ran `pre-commit run -a` with the updated pre-commit and am fixing `yamlfmt` problems.

### DIFF
--- a/docker-wrappers/SPRAS/example_config.yaml
+++ b/docker-wrappers/SPRAS/example_config.yaml
@@ -102,14 +102,13 @@ datasets:
     other_files: []
     # Relative path from the spras directory
     data_dir: "input"
-    # -
-#   label: data1
-#   # Reuse some of the same sources file as 'data0' but different network and targets
-#   node_files: ["node-prizes.txt", "sources.txt", "alternative-targets.txt"]
-#   edge_files: ["alternative-network.txt"]
-#   other_files: []
-#   # Relative path from the spras directory
-#   data_dir: "input"
+#   - label: data1
+#     # Reuse some of the same sources file as 'data0' but different network and targets
+#     node_files: ["node-prizes.txt", "sources.txt", "alternative-targets.txt"]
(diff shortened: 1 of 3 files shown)


docker-wrappers/SPRAS/example_config.yaml
```diff
@@ -103,13 +103,13 @@ datasets:
...
-    #   other_files: []
-    #   # Relative path from the spras directory
-    #   data_dir: "input"
+#   label: data1
+#   # Reuse some of the same sources file as 'data0' but different network and targets
+#   node_files: ["node-prizes.txt", "sources.txt", "alternative-targets.txt"]
+#   edge_files: ["alternative-network.txt"]
+#   other_files: []
+#   # Relative path from the spras directory
+#   data_dir: "input"
```
tristan-f-r, line 112:
```suggestion
#   - label: data1
#     # Reuse some of the same sources file as 'data0' but different network and targets
#     node_files: ["node-prizes.txt", "sources.txt", "alternative-targets.txt"]
#     edge_files: ["alternative-network.txt"]
#     other_files: []
#     # Relative path from the spras directory
#     data_dir: "input"
```

tristan-f-r (commented):
Fascinating! I don't get these change requests locally, nor is CI reporting it. I mainly checked because the formatting here:

https://github.com/Reed-CompBio/spras/blob/3dfb561f8903bc001044c580a6e8c339e151f5fc/docker-wrappers/SPRAS/example_config.yaml#L105-L112

is a little strange. (I've added a suggestion to use the "correct" formatting here heuristically)

docker-wrappers/SPRAS/example_config.yaml
agitter: I'm not able to apply the suggested comment: "Applying suggestions on deleted lines is currently not supported."

I'll make this change locally.

tristan-f-r (approved):

PR #363, opened 2025-07-29
fix: make dataset nodes ordering stable

### LINKED ISSUE #336: `generate_inputs` over all `required_inputs`
https://github.com/Reed-CompBio/spras/pull/313#discussion_r2216969861

tristan-f-r:
This is the main change:
https://github.com/Reed-CompBio/spras/blob/a61bfcea70c216150022064ff834296bf5b31bc7/spras/dataset.py#L107-L108

Some algorithms depend on the ordering of their nodes, which is why this is marked as needed for benchmarking: we can guarantee immutable generated outputs this way. We never want interactome permutations, as they don't produce immutable output for some algorithms.

---

This also adds tests to all inputs, closing #336.

This will be annoying to review. I have at least gone through and compared it with the regular input file to see that this looks right, but a thorough review of all of the test files will be unfortunately tedious.


ntalluri:
Can we name which algorithms depend on this?

PR #364, opened 2025-07-29
ci: windows

tristan-f-r:
We can have Windows CI if we ignore all of the important tests! I'm testing this through CI, so it will be a bit.


.github/workflows/test-spras.yml
```diff
@@ -25,17 +25,36 @@ jobs:
...
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: spras
+          environment-file: environment.yml
+          auto-activate-base: false
+          miniconda-version: 'latest'
+      - name: Run tests
+        shell: bash --login {0}
+        # Verbose output and disable stdout and stderr capturing
+        run: pytest -vs --ignore=test/AllPairs --ignore=test/BowTieBuilder --ignore=test/DOMINO --ignore=test/MEO --ignore=test/MinCostFlow --ignore=test/OmicsIntegrator1 --ignore=--ignore=test/OmicsIntegrator2 --ignore=test/PathLinker --ignore=test/ResponseNet --ignore=test/RWR --ignore=test/ST_RWR
```
tristan-f-r, line 44: This could be a good time to begin moving algorithms to their own folder.

tristan-f-r: #391.

tristan-f-r: Docker doesn't work on windows in GitHub CI. It would be good to at least test anything in Windows that doesn't rely on docker. [Earlier, I believe you had issues running SPRAS tests on Windows because of path issues with the analysis tests]

agitter: The [`windows-2025`](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md) runner lists these tools
- Docker 27.5.1
- Docker Compose v2 2.32.2
- Docker-wincred 0.9.3

Did Docker not work?

PR #282, opened 2025-06-20
ci: split

tristan-f-r:

PR #367, opened 2025-08-05
docs(contributing): reorganize

tristan-f-r:
Reorganization attempt for #282. Information about the release cycle and bumping `.test_durations` would go in this Maintaining SPRAS section.


docs/contributing/index.rst
```diff
@@ -1,18 +1,15 @@
-SPRAS contribution guide
-========================
+Contributing a new pathway reconstruction
```
agitter, line 1: We lost a word or phrase here. "algorithm guide"?

agitter (commented):
Great idea to not have one monolithic contribution doc.

agitter (approved):

Issue #349, opened 2025-07-24
Add More Summary Statistics


ntalluri:
When I was thinking through the grid search strategy, I came up with more statistics that we should add to the summary.txt files to help with parameter tuning.

- Density (Measures global sparsity)
- Modularity (Reflects community structure)  or Average clustering coefficient (Reflects local modularity)
- Diameter (measures longest and we can try to avoid large distances )or avg path length (measures avg path and we can try to avoid large distances)
- degree distribution (Captures hubs vs uniformity (look for heavy-tailed distribution) or max degree (try to remove one hub node problem)

agitter:
Representing the full degree distribution in the table will be hard. I suggestion picking a small number of points like median and max degree.

Same for diameter, you can report the max diameter across all connected components.

Clusters is a hard problem to automate in general. Is the average clustering coefficient easier to compute than modularity? Is it only supported for undirected graphs though?

PR #368, opened 2025-08-07
feat: add new summary statistics

ntalluri:
Adding in new statistics mentioned in #349.

### DIFF
--- a/spras/analysis/summary.py
+++ b/spras/analysis/summary.py
@@ -75,7 +75,7 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
         # Compute average shortest path length only for components with ≥2 nodes (undefined for singletons, set to 0.0)
         avg_path_lengths = [
             nx.average_shortest_path_length(nw.subgraph(c).copy()) if len(c) > 1 else 0.0
-            for c in nx.connected_components(nw)
+            for c in cc
         ]
 
         if len(avg_path_lengths) != 0:


spras/analysis/summary.py
```diff
@@ -50,8 +51,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
...
+
+        # Save the max diameter
+        cc = list(nx.connected_components(nw))
+        diameters = [
+            nx.diameter(nw.subgraph(c).copy()) if len(c) > 1 else 0
+            for c in cc
+        ]
+        max_diameter = max(diameters, default=0)
+
+        # # Save the max average clustering coefficient
```
ntalluri, line 73: This will only work for one component. With multiple the value is 0.0 always.

spras/analysis/summary.py
```diff
@@ -50,8 +51,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
         number_edges = nw.number_of_edges()
         ncc = nx.number_connected_components(nw)
 
+        # Save the max/median degree, average clustering coeffienct, and density
+        if number_nodes <= 1 or number_edges == 0:
+            max_degree = 0
```
tristan-f-r, line 56: `max_degree` and other degree statistics will be incorrectly reported as zero if we consider self loops. Is this a problem?

test/analysis/expected_output/expected_egfr_summary.txt
```diff
@@ -0,0 +1,10 @@
+Name	Number of nodes	Number of edges	Number of connected components	Density	Max degree	Median degree	Max diameter	Average path length	Nodes in prize	Nodes in sources	Nodes in targets	Nodes in active	Nodes in dummy	Parameter combination
```
tristan-f-r, line 1: (Note: The newly introduced test files will conflict with #320.)

spras/analysis/summary.py
tristan-f-r: We can drop this commented code then.

spras/analysis/summary.py
```diff
@@ -50,8 +51,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
...
+        # ]
+        # acc_max = max(component_accs, default=0.0)
+
+        # Save the average path lengths
+        avg_path_lengths = [
+            nx.average_shortest_path_length(nw.subgraph(c).copy()) if len(c) > 1 else 0.0
+            for c in nx.connected_components(nw)
+        ]
+
+        if avg_path_lengths:
```
tristan-f-r, line 85: They do the same thing (I hope), but this suggestion has clearer intent.

```suggestion
        if len(avg_path_lengths) != 0:
```

spras/analysis/summary.py
```diff
@@ -50,8 +51,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
...
+        else:
+            degrees = [deg for _, deg in nw.degree()]
+            max_degree = max(degrees)
+            median_degree = median(degrees)
+            density = nx.density(nw)
+
+        # Save the max diameter
+        cc = list(nx.connected_components(nw))
+        diameters = [
+            nx.diameter(nw.subgraph(c).copy()) if len(c) > 1 else 0
```
tristan-f-r, line 68: The restated comment here ("Save the max diameter") is less helpful than adding a comment about why we ignore the output of `nx.diameter` when the there is less than 2 nodes in any one individual component nodeset.

spras/analysis/summary.py
```diff
@@ -71,7 +108,7 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
         nw_info.append(cur_nw_info)
 
     # Prepare column names
-    col_names = ['Name', 'Number of nodes', 'Number of edges', 'Number of connected components']
+    col_names = ['Name', 'Number of nodes', 'Number of edges', 'Number of connected components', 'Density', 'Max degree', 'Median degree', 'Max diameter', 'Average path length']
```
tristan-f-r, line 111: As we add more summary statistics, we might want to use some kind of more modular system for adding statistics: the current `cur_mw_info` aligned with `col_names` approach looks pretty bad.

tristan-f-r (changes requested):
(Thanks for using the semantic commit style in the PR header! I use that commit style because it makes the changelog easier to skim [see the [0.5 release changelog for an example](https://github.com/Reed-CompBio/spras/releases/tag/0.5.0)])

spras/analysis/summary.py
agitter: Would you say that is out of scope for this PR and should be a new issue?

spras/analysis/summary.py
```diff
@@ -50,8 +51,44 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
         number_edges = nw.number_of_edges()
         ncc = nx.number_connected_components(nw)
 
+        # Save the max/median degree, average clustering coeffienct, and density
```
ntalluri, line 54:
```suggestion
        # Save the max/median degree, average clustering coefficient, and density
```

spras/analysis/summary.py
ntalluri: The idea that a self loop gets represented as an indegree of 2 for an undirected edge needs to be thought through more. Should we also also report # of self loops?

spras/analysis/summary.py
```diff
@@ -50,8 +51,41 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
...
+            for c in cc
+        ]
+        max_diameter = max(diameters, default=0)
+
+
+        # Save the average path lengths
+        # Compute average shortest path length only for components with ≥2 nodes (undefined for singletons, set to 0.0)
+        avg_path_lengths = [
+            nx.average_shortest_path_length(nw.subgraph(c).copy()) if len(c) > 1 else 0.0
+            for c in nx.connected_components(nw)
```
agitter, line 78: Can we reuse the connected components `cc` here?

agitter (commented):
I have one final small question, and then this is essentially ready to merge.

The discussion of how these summary statistics assume an undirected graph and convert directed edges does point out that we need to start writing docs. That can happen separately.

PR #371, opened 2025-08-14
docs(oi1): dummy_mode, `r` default reasoning

tristan-f-r:


docs/prms/oi1.rst
```diff
@@ -17,4 +17,10 @@ OI1 takes some optional arguments:
 * mu: controls the degree-based negative prizes (default 0.0)
 * noise: Standard Deviation of the gaussian noise added to edges in Noisy Edges Randomizations
 * g: (Gamma) multiplicative edge penalty from degree of endpoints
-* r: msgsteiner parameter that adds random noise to edges, which is rarely needed. (default 0)
+* r: msgsteiner parameter that adds random noise to edges, which is rarely needed because the ``noisy_edges`` option is recommended instead. (default 0)
+* dummy_mode: one of ``terminals``, ``all``, ``others``, or ``file``.
```
tristan-f-r, line 21: (this needs more reasoning as to why a user would want to set their own `dummy_mode` for `oi1`)

ntalluri:
Im going to look through more of the algorithm docs (most likely tomorrow). I was hoping if I find any other things to update, we can add it to this PR.

docs/prms/oi1.rst
```diff
@@ -17,4 +17,10 @@ OI1 takes some optional arguments:
 * mu: controls the degree-based negative prizes (default 0.0)
 * noise: Standard Deviation of the gaussian noise added to edges in Noisy Edges Randomizations
 * g: (Gamma) multiplicative edge penalty from degree of endpoints
```
agitter, line 19: I noticed this is the OI2 definition, not the OI1 definition. We should change it to:
`
```suggestion
* g: msgsteiner reinforcement parameter that affects the convergence of the solution and runtime, with larger values leading to faster convergence but suboptimal results (default 0.001)
```

agitter (approved):
I'll let the tests pass before merging.

Issue #370, opened 2025-08-12
Update Evaluation PR Curves to take into account Sources/Targets/Prizes Being in the Gold Standard (add a new baseline)


ntalluri:
I noticed an "issue" where the gold standard sometimes includes the same source and target nodes used in the reconstruction. Since these nodes may be easier to predict but could still be considered gold standard (for example being a part of a pathway), we decided to address this by adding an additional baseline to the PR curves that accounts for their presence in the gold standard and give them all a score of 1.0.

agitter:
The completely random ordering is an existing baseline. This could be considered a second, stronger baseline.

PR #372, opened 2025-08-14
feat: node evaluation baselines

ntalluri:
Working on #370

This PR adds in three node evaluation baselines

- [x] all nodes in the interactome (represents the precision of a random predictor)
- [x] all input nodes (assesses the overlap between the input nodes and the gold standard nodes)
- [ ] the input nodes plus their first neighbors (assesses the overlap between the input nodes + first neighbors and the gold standard nodes)

Note:

Double check what the figures look like and if they include all of these right now.


test/evaluate/test_evaluate.py
```diff
@@ -35,6 +35,8 @@ def setup_class(cls):
             'other_files': []
         })
 
+        # TODO figure out why the input-nodes file is not being included in the data.pickle file
```
ntalluri, line 38: There is an error where the input-nodes files is not being added

tristan-f-r: #379 fixes the core of the issue: I don't have a good way to emit a warning without having to manually iterate through the entire dataframe, unfortunately.

ntalluri: > Warn the user when they specify too many columns per any row (is this an option in pandas?)
We can check the length of the columns to make sure it's not past a certain number for a simple solution.

tristan-f-r: It was a very strange error - it was because there were too many tabs in the first (non-header) row for the node table, which made pandas eagerly parse the `NODEID` column as the index column.

Snakefile
```diff
@@ -544,7 +544,7 @@ rule evaluation_per_algo_ensemble_pr_curve:
     run:
         node_table = Evaluation.from_file(input.gold_standard_file).node_table
         node_ensembles_dict = Evaluation.edge_frequency_node_ensemble(node_table, input.ensemble_files, input.dataset_file)
-        Evaluation.precision_recall_curve_node_ensemble(node_ensembles_dict, node_table, output.pr_curve_png, output.pr_curve_file, include_aggregate_algo_eval)
+        Evaluation.precision_recall_curve_node_ensemble(node_ensembles_dict, node_table, input.dataset_file,output.pr_curve_png, output.pr_curve_file, include_aggregate_algo_eval)
```
ntalluri, line 547: update to send the node file or network, not the whole dataset. Do this for all the functions that are doing this (for edge_frequency_node_ensemble and precision_recall_curve_node_ensemble)

spras/evaluation.py
```diff
@@ -380,13 +378,46 @@ def precision_recall_curve_node_ensemble(node_ensembles: dict, node_table: pd.Da
 
         prc_dfs = []
         metric_dfs = []
+        prc_input_nodes_baseline_df = None
 
         baseline = None
+        precision_input_nodes = None
```
ntalluri, line 384: I don't think these are needed

Snakefile
```diff
@@ -524,8 +524,10 @@ rule evaluation_ensemble_pr_curve:
         pr_curve_file = SEP.join([out_dir, '{dataset_gold_standard_pairs}-eval', 'pr-curve-ensemble-nodes.txt']),
     run:
         node_table = Evaluation.from_file(input.gold_standard_file).node_table
-        node_ensemble_dict = Evaluation.edge_frequency_node_ensemble(node_table, input.ensemble_file, input.dataset_file)
-        Evaluation.precision_recall_curve_node_ensemble(node_ensemble_dict, node_table, output.pr_curve_png, output.pr_curve_file)
+        input_interactome = Evaluation.from_file(input.dataset_file).get_interactome()
+        node_ensemble_dict = Evaluation.edge_frequency_node_ensemble(node_table, input.ensemble_file, input_interactome)
+        input_nodes = Evaluation.from_file(input.dataset_file).get_node_columns(["sources", "targets", "prize", "active"])
```
agitter, line 529: Can you store the dataset the first time you get it `from_file`?

We may want a new helper function in Dataset to take the place of `get_node_columns(["sources", "targets", "prize", "active"])` so that isn't hard-coded here. Something that would say "get all of the columns that should be considered interesting input" so that if new algorithms update that in the future we don't have to track down all the places in the source that list those node properties.

Snakefile
```diff
@@ -543,8 +545,10 @@ rule evaluation_per_algo_ensemble_pr_curve:
         pr_curve_file = SEP.join([out_dir, '{dataset_gold_standard_pairs}-eval', 'pr-curve-ensemble-nodes-per-algorithm.txt']),
     run:
         node_table = Evaluation.from_file(input.gold_standard_file).node_table
-        node_ensembles_dict = Evaluation.edge_frequency_node_ensemble(node_table, input.ensemble_files, input.dataset_file)
-        Evaluation.precision_recall_curve_node_ensemble(node_ensembles_dict, node_table, output.pr_curve_png, output.pr_curve_file, include_aggregate_algo_eval)
+        input_interactome = Evaluation.from_file(input.dataset_file).get_interactome()
+        node_ensembles_dict = Evaluation.edge_frequency_node_ensemble(node_table, input.ensemble_files, input_interactome)
+        input_nodes = Evaluation.from_file(input.dataset_file).get_node_columns(["sources", "targets", "prize", "active"])
+        Evaluation.precision_recall_curve_node_ensemble(node_ensembles_dict, node_table, input_nodes,output.pr_curve_png, output.pr_curve_file, include_aggregate_algo_eval)
```
agitter, line 551: Same comments above apply here

spras/evaluation.py
```diff
@@ -380,13 +378,40 @@ def precision_recall_curve_node_ensemble(node_ensembles: dict, node_table: pd.Da
...
         for label, node_ensemble in node_ensembles.items():
             if not node_ensemble.empty:
                 y_true = [1 if node in gold_standard_nodes else 0 for node in node_ensemble['Node']]
                 y_scores = node_ensemble['Frequency'].tolist()
+
+                # input nodes (sources, targets, prizes, actives) may be easier to recover but are still valid gold standard nodes;
+                # the Input_Nodes_Baseline PR curve highlights their overlap with the gold standard.
+                if prc_input_nodes_baseline_df is None:
+                    input_nodes_set = set(input_nodes['NODEID'])
+                    input_nodes_gold_intersection = input_nodes_set & gold_standard_nodes # TODO should this be all inputs nodes or the intersection with the gold standard for this baseline? I think it should be the intersection
```
agitter, line 393: Very good question. For a synthetic dataset like Panther pathways, the inputs are sampled from the pathway nodes that make up the gold standard so it doesn't matter.

For an omics input like EGFR, it matters substantially. What makes you prefer the intersection? I was inclined to say all input nodes because we cannot have a baseline algorithm that makes use of gold standard information as part of its ranking. I could create a valid pathway reconstruction algorithm that takes the input nodes and simply returns those. I can't use a gold standard in a valid pathway reconstruction algorithm, however.

spras/evaluation.py
```diff
@@ -380,13 +378,40 @@ def precision_recall_curve_node_ensemble(node_ensembles: dict, node_table: pd.Da
...
+                    input_nodes_ensemble_df = node_ensemble.copy()
+
+                    input_nodes_ensemble_df["Frequency"] = (
```
agitter, line 396: It would help to have a comment explaining what this is doing. I understand it. It's for future reference.

test/evaluate/test_evaluate.py
```diff
@@ -159,9 +161,11 @@ def test_precision_recall_curve_ensemble_nodes(self):
         out_path_png.unlink(missing_ok=True)
         out_path_file = Path(OUT_DIR + 'pr-curve-ensemble-nodes.txt')
         out_path_file.unlink(missing_ok=True)
+        input_data = OUT_DIR + 'data.pickle'
+        input_nodes = Evaluation.from_file(input_data).get_node_columns(["sources", "targets", "prize", "active"])
```
agitter, line 165: I propose using the new helper function in the tests as well

agitter (commented):
This is getting close.

spras/evaluation.py
agitter: Deciding this is the last point of feedback and last decision to finalize. Then I can do a last careful review and we should be ready to merge.

In our meeting, we discussed how options 1 and 2 will give the same recall. The only difference is that one will have some precision value and the other always has precision of 1.0.

spras/dataset.py
```diff
@@ -171,6 +171,15 @@ def contains_node_columns(self, col_names: list[str] | str):
                     return False
                 return True
 
+    def get_interesting_input_nodes(self) -> pd.DataFrame:
+        """
+        Returns: a table listing the input nodes considered as starting points for pathway reconstruction algorithms,
+        restricted to nodes that have at least one of the specified attributes.
```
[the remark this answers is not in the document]
ntalluri: what is an interesting node?

config/config.yaml
ntalluri, line 1: revert back to what it was before

ntalluri:
@agitter this is ready to be reviewed

ntalluri:
https://github.com/Reed-CompBio/spras/pull/379

Do we need to double check for this issue? I don't think I was running it this bug anymore.

ntalluri:
I still need to add in the 3rd node baseline - input nodes + first neighbors vs gold standard.

PR #387, opened 2025-08-31
refactor!: move container options to shared `containers` key

tristan-f-r:
This is 1/2 PRs being extracted from #329 to make it easier to add some `dsub` changes for local testing.

I've split this PR up specifically because the extra file is motivated by #329: specifically, we move `ContainerSettings` to `container_schema.py` instead of keeping it in `schema.py` to prevent the _future_ dependency diamond of:

- `containers.py` will depend on the containers schema
- `[algorithm].py` will, because of `run_containers`, depend on `containers.py`
- `schema.py` will, thanks to #329, depend on `[algorithm].py` for all algorithms.


config/config.yaml
```diff
@@ -3,30 +3,31 @@
...
-  owner: reedcompbio
+containers:
+  # Specify the container framework used by each PRM wrapper. Valid options include:
+  # - docker (default if not specified)
+  # - singularity -- Also known as apptainer, useful in HPC/HTC environments where docker isn't allowed
+  # - dsub -- experimental with limited support, used for running on Google Cloud with the All of Us cloud environment.
+  #   - There is no support for other environments at the moment.
+  framework: docker
+
+  # Only used if container_framework is set to singularity, this will unpack the singularity containers
```
agitter, line 14: "container_framework" will now be "framework" in all these config files

spras/config/container_schema.py
```diff
@@ -0,0 +1,69 @@
...
+    hash_length: int = 7
+
+    model_config = ConfigDict(extra='forbid')
+
+@dataclass
+class ProcessedContainerSettings:
+    framework: ContainerFramework = ContainerFramework.docker
+    unpack_singularity: bool = False
+    prefix: str = DEFAULT_CONTAINER_PREFIX
+    hash_length: int = 7
```
agitter, line 46: Why is the `hash_length` part of the container settings?

spras/containers.py
```diff
@@ -388,7 +388,7 @@ def prepare_volume(filename: Union[str, PurePath], volume_base: Union[str, PureP
     if isinstance(filename, PurePath):
         filename = str(filename)
 
-    filename_hash = hash_filename(filename, config.config.hash_length)
+    filename_hash = hash_filename(filename, config.config.container_settings.hash_length)
```
agitter, line 391: It isn't intuitive to me to look for the `hash_length` in the `container settings`.

agitter (commented):
The contributing guide will also need updates.

spras/config/container_schema.py
tristan-f-r: This just makes it easier to pass around to `containers.py`, untied to the global configuration. This is also the only 'weird' change, but the file generated by `hash_length` will never appear in `output`.

By default, `hash_length` is set to the top-level hash length.

(This can be added to a doc comment for `hash_length` since this was not clear before)

spras/containers.py
tristan-f-r: (See comment from the above review)

agitter: Looking ahead at #390 I'm still having a hard time understanding this design. The hash length is a global config property, not a container setting. Now `prepare_volume`'s signature will have to be modified so that a `ProcessedContainerSettings` instance has to be supplied so that we can extract information that isn't really part of the container settings, it is a global setting.

tristan-f-r: I agree that it does make the `prepare_volume`'s signature worse.

However, `prepare_volume` is a SPRAS-internal API, and hash length is a setting passed to containers (as it is explicitly a dependency of `prepare_volume` used to set up the name of prepared volumes used in containers), making it a container setting.

The dependency diamond that motivates this change still only appears in #329 and not #390.

PR #386, opened 2025-08-30
ci: use main branch

tristan-f-r:
Leftover from #8 that silently killed `build-and-remove-template.yml`. Also corrects the branch name in docs.

PR #389, opened 2025-09-06
fix: correct oi1 container

tristan-f-r:
[This PRs CI is failing because the new OI1 image has not been pushed.]

The OI1 docker image has been broken for quite some time, but was not being rebuilt until #386 was merged. This PR fixes that.

This PR can be overshadowed in favor of #314, which contains this patch, and already has the new OI1 image pushed to GHCR.

Regardless, either this PR or #314 are required for CI to pass. (If CI is passing in other PRs, they are likely not up-to-date with `main`).


agitter:
I built and pushed the old `v1` image so we have a tagged version on DockerHub. I had to modify the Dockerfile
```
RUN sed -i 's|http://deb.debian.org/debian|http://archive.debian.org/debian|g' /etc/apt/sources.list && \
    sed -i 's|http://security.debian.org/debian-security|http://archive.debian.org/debian-security|g' /etc/apt/sources.list && \
    sed -i '/buster-updates/d' /etc/apt/sources.list && \
    apt-get -o Acquire::Check-Valid-Until=false update && \
    apt-get -qq update --allow-releaseinfo-change && \
    apt-get install -y build-essential libx11-dev libboost-dev libboost-program-options-dev
```

I also pushed this new `v2`.

PR #390, opened 2025-09-06
refactor: pass around container settings

tristan-f-r:
- [ ] Depends on #387

PR 2/2 split from #321. This is a continuation of #387.


spras/analysis/cytoscape.py
```diff
@@ -33,25 +35,28 @@ def run_cytoscape(pathways: List[Union[str, PurePath]], output_file: str, contai
...
 
     # Map the pathway filenames and add them to the Python command
     for pathway in pathways:
-        bind_path, mapped_pathway = prepare_volume(pathway, work_dir)
+        bind_path, mapped_pathway = prepare_volume(pathway, work_dir, container_settings)
         volumes.append(bind_path)
         # Provided the mapped pathway file path and the original file path as the label Cytoscape
         command.extend(['--pathway', f'{mapped_pathway}|{pathway}'])
 
+    out_dir = Path(output_file).parent
```
agitter: Where should these log files go for Cytoscape? It isn't a pathway algorithm, so it's out_dir isn't self-contained like the others are.

spras/containers.py
```diff
@@ -176,50 +176,50 @@ def env_to_items(environment: dict[str, str]) -> Iterator[str]:
...
     if normalized_framework == 'docker':
         return run_container_docker(container, command, volumes, working_dir, environment, network_disabled)
     elif normalized_framework == 'singularity' or normalized_framework == "apptainer":
-        return run_container_singularity(container, command, volumes, working_dir, out_dir, environment)
+        return run_container_singularity(container, command, volumes, working_dir, out_dir, container_settings, environment)
     elif normalized_framework == 'dsub':
         return run_container_dsub(container, command, volumes, working_dir, environment)
     else:
-        raise ValueError(f'{framework} is not a recognized container framework. Choose "docker", "dsub", or "singularity".')
+        raise ValueError(f'{container_settings.framework} is not a recognized container framework. Choose "docker", "dsub", or "singularity".')
```
agitter: We should add "apptainer" to this message

test/test_util.py
```diff
@@ -41,7 +41,7 @@ def test_hash_params_sha1_base32(self):
                               ('test/OmicsIntegrator1/output', PurePosixPath('/spras'), '/spras/TNDO5TR/output'),
                               ('../src', '/spras', '/spras/NNBVZ6X/src')])
     def test_prepare_volume(self, filename, volume_base, expected_filename):
-        _, container_filename = prepare_volume(filename, volume_base)
+        _, container_filename = prepare_volume(filename, volume_base, config.config.container_settings)
```
agitter: This gets us a default `ProcessedContainerSettings`?

tristan-f-r: (We should provide that explicitly instead of misdirecting it with the `container_settings` from the global config :+1:)

spras/analysis/cytoscape.py
```diff
@@ -48,6 +48,8 @@ def run_cytoscape(pathways: List[Union[str, PurePath]], output_file: str, contai
         # Provided the mapped pathway file path and the original file path as the label Cytoscape
         command.extend(['--pathway', f'{mapped_pathway}|{pathway}'])
 
+    # out_dir does not contain all Cytoscape output files, but it is where
+    # profiling's `usage-profile.tsv` file goes if `enable_profiling` is enabled.
```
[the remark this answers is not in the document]
tristan-f-r: There are naming collisions and this is a bad change; however, this also should not have worked in the original #260 PR. The current change in the cytoscape file, thanks to python not doing any type-checking, has been working, and because we ignore the Cytoscape singularity test in CI, it was never pointed out that Cytoscape's `env` is _being passed as `out_dir`_: profiling silently broke Cytoscape on Singularity.

This was never caught in Canek or my HTCondor runs because the original configuration Canek ran was before #260, and I turned off Cytoscape in my runs [just to focus on general docker errors].


I'll continue passing an ill-typed value for `out_dir`. See https://github.com/Reed-CompBio/spras/pull/390/commits/50d7b50eb8dc88c249157adfe1abe1b96a6ebc57. (This is also a good indication that we should start considering type-checking in SPRAS)

agitter (approved):

Issue #23, opened 2021-05-10
Cannot run Linux Docker image on Windows in GitHub Actions


agitter:
[This test](https://github.com/Reed-CompBio/spras/runs/2550298844?check_suite_focus=true) that attempted to pull the Omics Integrator Docker images on Windows failed with the error message `image operating system "linux" cannot be used on this platform`.

agitter:
This is a known issue that we cannot fix ourselves.  See [this thread](https://github.com/actions/virtual-environments/issues/1143) for more information.

PR #391, opened 2025-09-06
test: mv algorithm tests to subfolder

tristan-f-r:
The result of

```bash
mkdir test/algorithms
git mv test/AllPairs/ test/BowTieBuilder/ test/DOMINO/ test/LocalNeighborhood/ test/MEO test/MinCostFlow/ test/OmicsIntegrator1 test/OmicsIntegrator2 test/PathLinker/ test/ResponseNet/ test/RWR/ test/ST_RWR/ test/algorithms/
```

with manual changes for file paths.

This is a convenient change for #364, as it helps the Windows CI ignore the unrunnable docker containers (see #23). We may also want to do this for the actual `spras` source code, but that's more refactor-heavy than this.


agitter:
How did this help #364? I wasn't clear on why we ignore so many tests on Windows, which we could discuss there instead of here.

PR #396, opened 2025-09-23
docs(htcondor): 0.2.0 -> 0.5.0

tristan-f-r:


agitter (approved):
I'm okay with this change. I don't plan to update this version number with every release. I see the version as an example, and we should write the docs in a way that makes that clear.

PR #406, opened 2025-10-06
feat: Dataset#category

tristan-f-r:
For parameter tuning. This is a short PR that just adds the Dataset#category parameter without usage, as I want to make sure that we agree on `category` over `categories`: the latter, while more general, seems biologically useless.


spras/config/schema.py
```diff
@@ -121,12 +121,14 @@ class Dataset(BaseModel):
     # validation & coercion logic before we check it against our own
     # requirements
     label: Annotated[str, AfterValidator(label_validator("Dataset"))]
+    category: Optional[str] = None
+    "The dataset category, for working with multiple datasets at once in the configuration."
```
ntalluri, line 125: I would be more specific on this: 

```suggestion
    "The dataset category, for working with multiple datasets collections at once in the configuration."
```

Issue #128, opened 2023-09-18
Update Summary.py For Mixed Graphs


ntalluri:
Summary.py currently utilizes networkx to summarize the outputted file paths. The issue with networkx is that its library does not handle mixed directionality. As a result, summary.run appears to be useless for the time being and has been commented out until the code is updated to operate on mixed graphs. summary.run computes degree, nx.clustering, and nx.betweenness_centrality; however, when the graph is mixed, picking either directed or undirected makes no sense. We maintained summary.summarize_networks as an undirected graph because it was initially used to derive its information. The information it returns is valuable and does not require directionality to explain the information it returns. 

This will hopefully be an easy future summer project for a new incoming researcher

agitter:
Implementing this would require either using a different Python graph package that supports mixed graphs or implementing the graph summary statistics from scratch using our own graph data structure that supports mixed graphs.

PR #411, opened 2025-10-10
refactor: separate statistic computation

tristan-f-r:
We also make graph statistics lazy. Laziness isn't used in `summary.py`, but I assume that we'll have more computationally expensive graph statistics as SPRAS develops, especially when it can take long to compute for our larger graphs, so this also splits up statistic generation into different rules.

Most importantly, this allows us to re-use statistics by consuming specific statistics as input files, which is currently used in #431.

- [x] Depends on #320 for the summary statistics test (integration testing over unit testing is now required since the heavy workflow lifting is done by Snakemake).


spras/analysis/summary.py
```diff
@@ -39,52 +41,17 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
...
-            for c in cc
-        ]
-
-        if len(avg_path_lengths) != 0:
-            avg_path_len = sum(avg_path_lengths) / len(avg_path_lengths)
-        else:
-            avg_path_len = 0.0
+
+        # We use literal_eval here to easily coerce to either ints or floats, depending.
+        graph_statistics = [ast.literal_eval(Path(file).read_text()) for file in statistics_files]
```
agitter: Is there an example of the statistics file? I'm not sure why whether `literal_eval` makes sense or this could be loaded from JSON or something else because I haven't tested locally yet.

spras/statistics.py
```diff
@@ -0,0 +1,70 @@
...
+
+    return max_diameter, avg_path_len
+
+# The type signature on here is quite bad. I would like to say that an n-tuple has n-outputs.
+statistics_computation: dict[tuple[str, ...], Callable[[nx.DiGraph], tuple[float | int, ...]]] = {
+    ('Number of nodes',): lambda graph : (graph.number_of_nodes(),),
+    ('Number of edges',): lambda graph : (graph.number_of_edges(),),
+    ('Number of connected components',): lambda graph : (nx.number_connected_components(graph.to_undirected()),),
+    ('Density',): lambda graph : (nx.density(graph),),
+
```
agitter: Extra line?

Snakefile
```diff
@@ -310,18 +310,33 @@ rule viz_cytoscape:
     run:
         cytoscape.run_cytoscape(input.pathways, output.session, container_settings)
 
+for keys, values in statistics_computation.items():
```
agitter: This needs some context via comments

agitter (commented):
I like the new design. I'm not confident I follow everything in `spras/statistics.py` correctly or how the testing updates intersect with #320.

spras/statistics.py
```diff
@@ -0,0 +1,70 @@
+"""
+Graph statistics, used to power summary.py.
```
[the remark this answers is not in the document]
tristan-f-r: Sounds good. I believe every assumption made about graph directionality is a direct consequence of the networkx calls, but that isn't obvious without an LSP.

Snakefile
```diff
@@ -310,18 +310,33 @@ rule viz_cytoscape:
...
+    pythonic_name = 'generate_' + '_and_'.join([key.lower().replace(' ', '_') for key in keys])
+    rule:
+        name: pythonic_name
```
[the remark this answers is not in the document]
tristan-f-r: Yes! That's a slightly rarer feature so I'll add a link to it, but [it is on Snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#procedural-rule-definition).

spras/statistics.py
tristan-f-r: It's barely for style. I think it's more confusing though now that I'm re-looking at it.

spras/analysis/summary.py
tristan-f-r: I'll clarify a little bit more, but the graph statistics are (right now) `int`s or `float`s, which only differ in their string representation by whether or not they have a `.` prefix at the end.
One could make a little string manipulation check that reads the text from the file and checks if it has `.` then returns a `float` else returns an `int`, but this seemed clearer at the time.

spras/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    # number_of_nodes is a cheap call
+    if graph.number_of_nodes() == 0:
+        return (0, 0.0)
+    else:
+        degrees = [deg for _, deg in graph.degree()]
+        return max(degrees), median(degrees)
+
+def compute_on_cc(directed_graph: nx.DiGraph) -> tuple[int, float]:
+    # We convert our directed_graph to an undirected graph as networkx (reasonably) does
```
ntalluri, line 32: I can't remember why we do this. @agitter I remember we talked about this years ago.

spras/analysis/summary.py
```diff
@@ -120,67 +99,3 @@ def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, alg
     )
 
     return nw_info
-
-
-def degree(g):
-    return dict(g.degree)
-
-# TODO: redo .run code to work on mixed graphs
```
ntalluri, line 128: please add this code from 128 to the end somewhere, we are saving this for the future todo.

spras/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+    return max_diameter, avg_path_len
+
+# The type signature here is meant to be 'an n-tuple has n-outputs.'
+statistics_computation: dict[tuple[str, ...], Callable[[nx.DiGraph], tuple[float | int, ...]]] = {
+    ('Number of nodes',): lambda graph : (graph.number_of_nodes(),),
+    ('Number of edges',): lambda graph : (graph.number_of_edges(),),
+    ('Number of connected components',): lambda graph : (nx.number_connected_components(graph.to_undirected()),),
+    ('Density',): lambda graph : (nx.density(graph),),
+    ('Max degree', 'Median degree'): compute_degree,
+    ('Max diameter', 'Average path length'): compute_on_cc,
```
ntalluri: Can we separate 'Max diameter' and 'Average path length' into two separate functions?

spras/analysis/summary.py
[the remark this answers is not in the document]
ntalluri: was there an issue made for this?

Snakefile
```diff
@@ -312,18 +313,48 @@ rule viz_cytoscape:
...
         cytoscape.run_cytoscape(input.pathways, output.session, container_settings)
 
+# We generate new Snakemake rules for every statistic
+# to allow parallel and lazy computation of individual statistics
+for keys in statistics_computation.keys():
+    pythonic_name = 'generate_' + '_and_'.join([key.lower().replace(' ', '_') for key in keys])
+    rule:
+        # (See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#procedural-rule-definition)
+        name: pythonic_name
+        input: pathway_file = rules.parse_output.output.standardized_file
```
ntalluri, line 323: what is the standardized file? Where is this being defined?

ntalluri:
Is there a reason why we need to have separate statistic folders within each output subnetwork folder? I don't understand this motivation.

ntalluri:
I read the conversion between you and tony, I see the motivation now.

spras/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if len(avg_path_lengths) != 0:
+        avg_path_len = sum(avg_path_lengths) / len(avg_path_lengths)
+    else:
+        avg_path_len = 0.0
+
+    return max_diameter, avg_path_len
+
+# The type signature here is meant to be 'an n-tuple has n-outputs.'
+statistics_computation: dict[tuple[str, ...], Callable[[nx.DiGraph], tuple[float | int, ...]]] = {
```
ntalluri, line 64: are we going to have these options separated in the config file or are all of these going to be computed each time summary: true is enabled? (the latter is simpler)

Snakefile
ntalluri: can you explain again how this works?

Snakefile
```diff
@@ -312,18 +313,48 @@ rule viz_cytoscape:
...
+        output: [SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'statistics', f'{key}.txt']) for key in keys]
+        # It is very tempting to use `.items()` instead of `.keys()` above, but
```
ntalluri, line 325: do we need to motivate this itself? I think you can just explain what is happening instead of saying we could do something else.

Snakefile
```diff
@@ -312,18 +313,48 @@ rule viz_cytoscape:
     run:
         cytoscape.run_cytoscape(input.pathways, output.session, container_settings)
 
+# We generate new Snakemake rules for every statistic
+# to allow parallel and lazy computation of individual statistics
```
ntalluri, line 317: is there a way you can explain what is happening in 1 - 3 sentences so someone doesn't have to try to figure it out from the snakemake code (this is a little too complicated to understand fast).

spras/statistics.py
ntalluri: I think the latter is what is happening in this PR. It makes sense to keep it like this.

ntalluri (commented):
I am running this locally still, but this is my current review

spras/analysis/summary.py
ntalluri: Is there an issue?

spras/statistics.py
tristan-f-r: I'm not sure what 'this' in the conclusion was, but to clarify, we internally do the former, and externally (in the configuration) do the latter.

In the future, one can now easily extend the configuration to do the former, but the more important use of this refactor is to have heuristics request specific graph statistics, which would have previously been hacky.

Snakefile
tristan-f-r: At least for me, seeing this without the motivation of this choice would lead me to believe that this admits the "obvious" `items` refactor when this is not the case.

[In some cases, I would be more inclined to leave out this motivation even when the 'obvious' refactor causes issues, but the issue that this causes is nearly invisible and Snakemake considers this more edge-case behavior, so it felt worth to mention. More concretely, this issue laid dormant in this PR when I originally made it, and I wouldn't have noticed it if it wasn't for the hash PR]

Snakefile
tristan-f-r: These are precisely from the [rule dependencies](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#rule-dependencies) feature. In SPRAS, I believe this is the first time we are using them, though they are a very old Snakemake feature and we frankly really should be using them more.

spras/statistics.py
tristan-f-r: [I'm a little confused - is it why we compute the number of connected components in the first place?]

spras/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+    ]
+
+    if len(avg_path_lengths) != 0:
+        avg_path_len = sum(avg_path_lengths) / len(avg_path_lengths)
+    else:
+        avg_path_len = 0.0
+
+    return max_diameter, avg_path_len
+
+# The type signature here is meant to be 'an n-tuple has n-outputs.'
```
tristan-f-r:
```suggestion
# The type signature here is meant to be 'an n-tuple has n outputs.'
```

Snakefile
tristan-f-r: (Assuming this is for the procedural rule generation) Does the 'We generate new Snakemake rules for every statistic' not suffice?

spras/analysis/summary.py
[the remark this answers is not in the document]
tristan-f-r: https://github.com/Reed-CompBio/spras/issues/128

spras/statistics.py
tristan-f-r: If it's about the undirected graph conversion, that comment should be why.

agitter: @ntalluri was your comment asking why we convert to undirected for the connected component calculation?

Related to my comment above, I recommend we stay with reading undirected graphs and use them throughout. That affects other statistics like degree as well.

spras/analysis/summary.py
agitter: I believe this is the code I asked to have deleted here https://github.com/Reed-CompBio/spras/pull/411#discussion_r2806454214. Because we have an issue, I'm okay removing the dead code.

spras/analysis/summary.py
```diff
@@ -1,14 +1,17 @@
...
 
-import networkx as nx
 import pandas as pd
 
+from spras.statistics import from_output_pathway, statistics_options
+
 
 def summarize_networks(file_paths: Iterable[Path], node_table: pd.DataFrame, algo_params: dict[str, dict],
-                       algo_with_params: list[str]) -> pd.DataFrame:
+                       algo_with_params: list[str], statistics_files: Mapping[str, Iterable[str | os.PathLike]]) -> pd.DataFrame:
```
agitter, line 14: We now have `LoosePathLike` in `util`. Does it make sense to use it here?

spras/statistics.py
```diff
@@ -0,0 +1,80 @@
+"""
+Graph statistics, used to power summary.py.
+
+We allow for arbitrary computation of any specific statistic on some graph,
+computing more than necessary if we have dependencies. See the top level
+`statistics_computation` dictionary for usage.
+
+To make the statistics allow directed graph input, they will always take
+in a networkx.DiGraph, which contains even more information, even though
+the underlying graph may be just as easily represented by networkx.Graph.
```
agitter, line 10: Is this a change in functionality? In summary.py we had the opposite
> Network directionality is ignored and all edges are treated as undirected

When we previously assessed treating all graphs as directed or undirected, we decided that undirected would be less wrong.

spras/statistics.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        avg_path_len = 0.0
+
+    return (avg_path_len,)
+
+# The type signature here is meant to be 'an n-tuple has n outputs.'
+statistics_computation: dict[tuple[str, ...], Callable[[nx.DiGraph], tuple[float | int, ...]]] = {
+    ('Number of nodes',): lambda graph : (graph.number_of_nodes(),),
+    ('Number of edges',): lambda graph : (graph.number_of_edges(),),
+    ('Number of connected components',): lambda graph : (nx.number_connected_components(graph.to_undirected()),),
```
agitter, line 67: Again, having to convert everything to undirected is messy.

spras/statistics.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if len(avg_path_lengths) != 0:
+        avg_path_len = sum(avg_path_lengths) / len(avg_path_lengths)
+    else:
+        avg_path_len = 0.0
+
+    return (avg_path_len,)
+
+# The type signature here is meant to be 'an n-tuple has n outputs.'
+statistics_computation: dict[tuple[str, ...], Callable[[nx.DiGraph], tuple[float | int, ...]]] = {
```
agitter, line 64: This syntax is somewhat confusing. It took me a minute to realize why everything needs to be a tuple, that sometimes we have a function return one statistic and sometimes multiple. The design makes sense now that I get it, but a comment would have helped.

spras/statistics.py
```diff
@@ -0,0 +1,80 @@
...
+    ('Density',): lambda graph : (nx.density(graph),),
+    ('Max degree', 'Median degree'): compute_degree,
+    ('Max diameter',): compute_max_diameter,
+    ('Average path length',): compute_avg_path_lengths,
+}
+
+# All of the keys inside statistics_computation, flattened.
+statistics_options: list[str] = list(itertools.chain(*(list(key) for key in statistics_computation.keys())))
+
+def from_output_pathway(lines) -> nx.Graph:
```
agitter, line 77: When I saw it in summary.py, I didn't recognize this function as a graph loader from the name.

test/analysis/test_summary.py
```diff
@@ -32,7 +31,7 @@ def snakemake_output(request):
     param = request.param
     subprocess.run(["snakemake", "--cores", "1", "--configfile", f"test/analysis/input/{param}.yaml"])
     yield param # this runs the test itself: once this is passed, we go to test cleanup.
-    shutil.rmtree(f"test/analysis/input/run/{param}")
+    # shutil.rmtree(f"test/analysis/input/run/{param}")
```
agitter, line 34: Why do we not need this anymore?

Snakefile
```diff
@@ -312,18 +313,48 @@ rule viz_cytoscape:
...
+for keys in statistics_computation.keys():
```
agitter, line 318: Because of the complicated typing, I find it hard to track what `keys` is here. I'm pretty sure it is the tuples describing the statistics as strings.

Snakefile
```diff
@@ -312,18 +313,48 @@ rule viz_cytoscape:
...
+    pythonic_name = 'generate_' + '_and_'.join([key.lower().replace(' ', '_') for key in keys])
+    rule:
+        # (See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#procedural-rule-definition)
+        name: pythonic_name
+        input: pathway_file = rules.parse_output.output.standardized_file
+        output: [SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'statistics', f'{key}.txt']) for key in keys]
```
agitter, line 324: Do we need to sanitize the keys here as done in the pythonic_name above? When I run locally, I see output files like `Number of connected components.txt`.

agitter (commented):
I finally understand the design better and like it.

The tests don't pass for me locally
```
FAILED test/analysis/test_summary.py::TestSummary::test_example_networks[example] - AssertionError: assert False
FAILED test/analysis/test_summary.py::TestSummary::test_example_networks[egfr] - AssertionError: assert False
```

spras/statistics.py
ntalluri: The question was why we convert to undirected graphs to do all of the graph statistics.

agitter: We read in undirected graphs in the original summary table pull request #68. That was also before we had full directed edge support in SPRAS though.

The [connected components](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.number_connected_components.html) calculation is most affected by this. For directed graphs, we would need different networkx functions for strongly or weakly connected components.

spras/statistics.py
ntalluri: Yes, can we change all of these back to undirected

PR #418, opened 2025-10-14
docs: naming conventions

### LINKED ISSUE #196: Naming standards for inputs to algorithms
An index of each method and the type of input that it accepts would be very helpful. Perhaps we could add it to a README in the `/SPRAS/spras/` directory.

In addition, multiple input types appear to be the same. Is there a difference between ['sources', 'targets', 'edges'] and ['nodetypes', 'network'] inputs? Or can we condense these into just ['nodetypes', 'network']?

Along these same lines, what is the difference between 'network' and 'edges' when mentioned here? Are these the same things?

Here is a picture for context:
image

tristan-f-r:
With light touchups to title formatting. Closes #196 via documentation.

### DIFF
--- a/docs/contributing/maintain.rst
+++ b/docs/contributing/maintain.rst
@@ -8,7 +8,7 @@ be installed manually.
 
 Naming conventions
 ------------------
-SPRAS wrappers tries to follow the naming conventions of the algorithms they wrap.
+SPRAS wrappers try to follow the naming conventions of the algorithms they wrap.
 This includes their name and their ``required_inputs``: when sensible, we preserve
 the file name of inputs required by the internal algorithm.
 


docs/contributing/maintain.rst
```diff
@@ -2,10 +2,19 @@ Maintaining SPRAS
...
 Go dependency
------------------------
-The yamlfmt tool requires `Go <https://go.dev/doc/install>`__, which must
+-------------
+The ``yamlfmt`` tool requires `Go <https://go.dev/doc/install>`__, which must
 be installed manually.
 
+Naming conventions
+------------------
+SPRAS wrappers tries to follow the naming conventions of the algorithms they wrap.
```
agitter:
```suggestion
SPRAS wrappers try to follow the naming conventions of the algorithms they wrap.
```

PR #408, opened 2025-10-08
docs: alg usage & impl details

tristan-f-r:
Work on #26. Closes #28.

PR #426, opened 2025-10-20
fix(rwr,strwr): directionality conversions

tristan-f-r:
I mentioned that I would fix this in #408.


agitter (approved):
My understanding of these methods is that both run in a networkx DiGraph, which is why we need this update to ensure we pass them a fully directed graph. If that's correct, this looks good to me.

PR #429, opened 2025-10-23
refactor(config)!: drop directed setting

tristan-f-r:
It was marked as deprecated.

### DIFF
--- a/spras/config/schema.py
+++ b/spras/config/schema.py
@@ -11,7 +11,7 @@ We declare models using two classes here:
 """
 
 import re
-from typing import Annotated, Optional
+from typing import Annotated
 
 from pydantic import AfterValidator, BaseModel, ConfigDict, Field
 


agitter:
Now if I provide an old config file I get
```
TypeError in file "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\Snakefile", line 21:
'bool' object is not iterable
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\Snakefile", line 21, in <module>
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\spras\config\config.py", line 35, in init_global
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\spras\config\config.py", line 114, in __init__
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\spras\config\config.py", line 313, in process_config
  File "C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\spras\config\config.py", line 186, in process_algorithms
```

Previously I would get
```
C:\Users\agitter\Desktop\madison\collaborators\Ritz\spras\spras\config\config.py:316: UserWarning: UPDATE: we no longer use the directed key in the config file
  self.process_algorithms(raw_config)
```
and the workflow would run.

I agree we don't want to be too tethered to backwards compatibility, but the unintelligible error message is much worse than the user-friendly warning.

tristan-f-r:
I am always in favor of better error messages (including in this case), and we can (and do!) provide better error messages in general with #321, since we'll know in advance what algorithms support what keys.

> I agree we don't want to be too tethered to backwards compatibility, but the unintelligible error message is much worse than the user-friendly warning.

Especially since we currently get this with any typo of algorithm parameters.

Issue #318, opened 2025-07-09
Support two-stage parameter grid search


agitter:
A common pathway reconstruction workflow is to start with an initial grid of parameter combinations, inspect the output pathways by applying heuristics, and filter the output pathways that do not fall within the desirable heuristic ranges. An optional final step could be to refine the initial grid of parameter combinations to remove those that produced undesirable outputs and provide more granular step sizes around parameter values that produced desirable outputs.

SPRAS could directly support this with a few new features:
- Expanding the graph properties (heuristics) computed in the summary table. This could include graph statistics like clustering coefficient, etc.
- Allow a user to specify ranges of graph properties (heuristics) they find tolerable in the config file
- Filter output pathways based on tolerable ranges
- Rewrite a new config file with the parameter combinations that produced those tolerable ranges so the user can use it as a template for a second run

An example of filtering from https://journals.plos.org/plospathogens/article?id=10.1371/journal.ppat.1011492:
> After generating subnetworks for each parameter combination, we discarded: any empty networks; networks that did not have a connected component with at least 25 vertices; networks in which one vertex was connected to 10% or more of the total vertices (which indicates a dominant hub node); and networks that contained a chain of more than 8 vertices without any nonzero prizes. An additional filter for networks whose vertices were not at least half prize nodes did not affect any of the networks.

ntalluri:
<img width="969" height="587" alt="Image" src="https://github.com/user-attachments/assets/482bd5f6-5401-482e-b309-ce6d384127a3" />

tristan-f-r:
Filtering will be manual, but I think in that comment I was mainly curious about just the first step: how will the grid be generated in the first place?

ntalluri:
I will be predefining the first coarse grid, but I'm still deciding what that grid should be. I'm assuming I can figure it out when I do the performance testing of the parameters, otherwise I will set a grid per algorithm based on my knowledge of the parameters.

ntalluri:
This document currently contains the new idea for two stage parameter tuning: 

https://docs.google.com/document/d/1PJFGeNkE63RkopGdRIkEJYOJ2yPFKkxeIMtqXANep6U/edit?usp=sharing

ntalluri:
I have a couple of questions/comments:

1) for the metadata to be added to individual algorithm parameters, what is this supposed to be? Are these just algorithm parameters or other information now?

[...]

PR #431, opened 2025-10-30
feat: heuristics

tristan-f-r:
- [ ] Depends on #411.

We implement @ntalluri's graph heuristics, or `heuristics` in configurations.

```yaml
heuristics:
  number_of_nodes: "<1500"
  number_of_connected_components: ["10 < x < 100", "500<"]
  ...
```

If any heuristic is not met, the **run will stop**. Future error handling will better handle these errors (#21). [I avoid handling it here, since I want to split up the Snakefile first, mostly to make variable scoping cleaner. This change would cause this PR to be a mess.]

Most of this code is just parsing interval inequality notation.


spras/statistics.py
agitter: Ignored this for the scope of this PR

spras/config/heuristics.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @staticmethod
+    def format_failed_heuristic(heuristic: tuple[str, float | int, list[Interval]]) -> str:
+        name, desired, intervals = heuristic
+        if len(intervals) == 1:
+            interval_string = str(intervals[0])
+        else:
+            formatted_intervals = ", ".join([str(interval) for interval in intervals])
+            interval_string = f"one of the intervals ({formatted_intervals})"
+        return f"{name} expected {desired} in interval {interval_string}"
```
agitter: This text doesn't quite match up. You could get "in interval one of the intervals..."

spras/interval.py
```diff
@@ -0,0 +1,234 @@
...
+@dataclass
+class Interval:
+    EMPTY_STRING: ClassVar[str] = "{empty interval}"
+
+    lower: Optional[float]
+    upper: Optional[float]
+    lower_closed: bool
+    upper_closed: bool
+
+    def mem(self, num: float) -> bool:
```
agitter: What does `mem` stand for?

spras/interval.py
```diff
@@ -0,0 +1,234 @@
+"""
+Utilities for defining inequality intervals (e.g. l < x <= u)
+
+For graph heuristics, we allow inequality intervals of the form (num) < (id)?. For example,
+we can say "1500 <" for "1500 < x", or "1000 < x < 2000", etc.
+
+[If there is ever a library that does this, we should replace this code with that library.]
```
agitter: I didn't review this file yet. I looked for any existing libraries:
- https://boolean-parser.readthedocs.io/en/latest/intro.html is the best match. I'm still reading to confirm that it avoid `eval`
- SymPy and https://docs.sympy.org/latest/modules/parsing.html looks very powerful but uses `eval`
- https://github.com/AaryamanBhute/OpenExpressions may be able to do this. It's not as clear how to use it.

agitter (commented):
> If any heuristic is not met, the run will stop.

Is that temporary behavior? My understanding of our discussion of heuristics for #318 is that we would flag an output pathway as "failed" or "errored" per the current heuristics but the rest of the Snakemake workflow would continue.

We will need to add an example of the heuristics in a config file and document their usage once it is finalized.

spras/interval.py
tristan-f-r: `boolean-parser` avoids eval, and we can use it. I was worried about that library falling into unmaintained status, though it would be better than what is here right now. I can use it here :+1: 

(Optimally, I would like to use the parsing logic for booleans inside some kind of SMT, though I wasn't able to find a nice isolated library for this.)

spras/interval.py
tristan-f-r: If `num` is inside the interval: (we should swap this out with a library, and I'll check out the library from the below comment), but this should be `in`.

ntalluri:
For a heuristics, could we also have something like percentages? For example for the input_prizes, what if I wanted 10% - 50% of them?

ntalluri:
does this depend on #470?

tristan-f-r:
Right! Not yet - I should make that change :+1:

PR #433, opened 2025-10-30
refactor(evaluation): types

tristan-f-r:
For typehints in `Evaluation` objects. As discussed below, this should also be in `Dataset` once #326 is merged.

Also turns `load_files_from_dict -> __init__`, since `load_files_from_dict` overrides every single field in Evaluation. It's a constructor.


ntalluri:
This is the pattern we use for each Dataset right now, why does it need to change for GoldStandard?

PR #435, opened 2025-11-07
test(ml): use pathlib

tristan-f-r:
Switches `ml` to use `pathlib`, adding types for paths when needed.


agitter (approved):
I confirmed the changed tests pass locally in Windows

PR #436, opened 2025-11-07
fix(meo): use non-deprecated base image

tristan-f-r:
OpenJDK pulled versions `8` and `11` from DockerHub sometime ago. These containers no longer build.

(Sorry @jhiemstrawisc! I forgot you don't have access to DockerHub.)

---

This is in the review queue since it breaks CI, which is happening in #329.


agitter (approved):
I pulled the existing `reedcompbio/meo:latest` and tagged it as `reedcompbio/meo:v1`. Then I built and pushed `v2.`

I'm reruning the failed test jobs before merging.

PR #441, opened 2025-11-16
feat([st]rwr): consider weights

tristan-f-r:
[ST]RWR already supported weights internally.


ntalluri (commented):
this all looks good to me. I'll just have to run it locally to double check the implementations are working.

PR #442, opened 2025-11-18
chore([st]rwr): add doi

tristan-f-r:
https://doi.org/10.1093/bioinformatics/btq680


ntalluri (commented):
For random walk with restarts, is there an actual paper we can reference as well?

tristan-f-r:
You're right - I assumed Personalized PageRank came from TrustRank, but it was merely adopted _for_ TrustRank :+1:

PR #89, opened 2023-06-08
RWwR and TieDIE integration

Lyce24:
## Progress made:

1. Random Walk with Restart Algorithm: I have created a dedicated repository for implementing the random walk with restart algorithm. The repo can be found [here](https://github.com/Lyce24/random-walk-with-restart). I have also pushed the [Docker image](https://hub.docker.com/repository/docker/erikliu24/rwwr/general) for this project, which can be found in the `docker-wrapper/RandomWalk/Dockerfile`. To ensure the algorithm runs and parses outputs smoothly, I have written the necessary code in `src/random-walk.py`. Furthermore, I have included tests for this algorithm, similar to other algorithms. You can locate the tests in the `test/RandomWalk`.

2. TieDIE Algorithm Dockerization: I have successfully dockerized the TieDIE algorithm repository, which can be accessed [here](https://github.com/epaull/TieDIE). The [Docker image](https://hub.docker.com/repository/docker/erikliu24/tiedie/general) for TieDIE is using the setting in `docker-wrapper/TieDIE/Dockerfile`. Additionally, I have implemented the required code in `src/tiedie.py` to execute the algorithm. However, due to some issues with the TieDIE algorithm's additional parameters, which I will discuss in the next section, I have not yet started writing tests for it.

## Questions and Concerns:

1. For Random Walk with Restart algorithm, I think I need guidance on constructing the final pathway (subnetwork) based on the algorithm's output. The algorithm provides information about the importance of each node and edge, determined by their pagerank values. To create the subnetwork, I need to understand how to utilize this information effectively. Once I grasp the process of constructing the subnetwork using the available data, I can proceed to tackle this issue accordingly.
2. For TieDIE, the algorithm can run seemlessly with the default settings. However, some problems and questions I have are:
    - I have discovered a bug in the original TieDIE repository when setting `pagrank` to `True`, and interestingly, someone else reported the same issue years ago. Unfortunately, the author has not responded to the problem yet. Upon examining the code, I realized that the bug is simply due to a missing indent. Now, I'm unsure about the appropriate course of action to take in this situation. What would you recommend I do?
    - Another observation I made is that the example pathway file used in the original TieDIE repository is unweighted. I'm curious to know if this unweighted format affects our specific research objectives. I'm seeking a better understanding of the rationale behind using an unweighted pathway file and exploring possible changes to align it with our research goals.

PR #445, opened 2025-11-19
feat: TieDIE

tristan-f-r:
Follow-up and clean-up of #92, #89, for [TieDIE](https://github.com/reed-CompBio/tiedie).


ntalluri:
the algorithm needs to be added to the config

ntalluri:
currently, this algorithm doesn't work on the egfr dataset within SPRAS

PR #447, opened 2025-12-05
chore: bump version to 0.6.0

agitter:
I missed this in the version 0.6.0 release


tristan-f-r:
We should bump the SPRAS Docker image to 0.6.0 (as well as the other 12 mentions of version 0.5.0) - this does not have to happen here.

agitter:
I bumped all of the mentions I found in a search. I don't think a lot of these need to say current with each release because they are example version numbers. The files used with HTCondor jobs should be though.

PR #457, opened 2026-01-13
feat!: timeouts

### LINKED ISSUE #316: Generic spras `limit` option on containers
In the #225 PR, netmix2 graciously offers a `time_limit` parameter if its algorithm takes too long. While this isn't going to be added in that PR, this may be something worth considering in general.

Like the #297 issue, we should have a generic `limit` option that allows limiting both memory and total time ran on containers.

tristan-f-r:
Adds `timeout` to algorithms as a demonstration of passing through errors, and introduces run settings, breaking the original configuration design by introducing a new `params` option. Closes #316.

_Caveat_: ML requires at least one pathway, and failing pathways can break ML-work. How do we want to handle downstream analysis when errors occur (including in the future heuristic errors?)


tristan-f-r:
Based on the meeting just now, I'll look into a nice way to get this working with Singularity.

config/config.yaml
```diff
@@ -102,6 +102,7 @@ algorithms:
 
   - name: "allpairs"
     include: true
+    timeout: 1d
```
ntalluri, line 120: What are the other options? Can we have a list [5m, 1d, 10d]?

spras/analysis/cytoscape.py
```diff
@@ -58,5 +58,6 @@ def run_cytoscape(pathways: List[Union[str, PurePath]], output_file: str, contai
                          # (https://github.com/Reed-CompBio/spras/pull/390/files#r2485100875)
                          None,
                          container_settings,
+                         None,
```
ntalluri, line 61: what is this None associated with?

spras/config/config.py
```diff
@@ -157,6 +160,16 @@ def process_algorithms(self, raw_config: RawConfig):
                 # Do not parse the rest of the parameters for this algorithm if it is not included
                 continue
 
+            if alg.timeout:
+                # Coerce to an `int` if an int isn't possible.
+                timeout = parse(alg.timeout, granularity='seconds')
```
ntalluri: how does this convert the 1d into seconds?

spras/containers.py
```diff
@@ -290,13 +347,25 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
     bind_paths = [f'{prepare_path_docker(src)}:{dest}' for src, dest in volumes]
 
-    out = client.containers.run(container,
-                                command,
-                                stderr=True,
-                                volumes=bind_paths,
-                                working_dir=working_dir,
-                                network_disabled=network_disabled,
-                                environment=environment).decode('utf-8')
+    container_obj = client.containers.run(
```
ntalluri, line 354: can you explain this new update to this part of the code.

spras/prm.py
```diff
@@ -76,17 +77,23 @@ def run_typeless(cls, inputs: dict[str, str | os.PathLike], output_file: str | o
...
+        @param inputs: specified inputs
+        @param output_file: designated reconstructed pathway output
+        @param args: (T) typed algorithm params
+        @param container_settings: what settings should be associated with the individual container.
+        @param timeout: timeout in seconds to run the container with
+
+        See the algorithm-specific `PRM.generate_inputs` and `PRM.parse_output`
         for information about the input and output format.
+
+        Also see `PRM.run_typeless` for the non-pydantic version of this method (where `args` is a dict).
```
ntalluri, line 97: why would we want a typeless version?

Snakefile
```diff
@@ -260,6 +263,22 @@ def collect_prepared_input(wildcards):
 
     return prepared_inputs
 
+def mark_error(file, **details):
```
ntalluri: can we start documentation somewhere to define what an error is / what causes an error.

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
     # Overwriting files can happen because the pathway reconstruction algorithms often generate output files with the
     # same name regardless of the inputs or parameters, and these aren't renamed until after the container command
     # terminates
-    output: pathway_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt'])
+    output:
+        pathway_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt']),
+        # Despite this being a 'log' file, we don't use the log directive as this rule doesn't actually throw errors.
+        resource_info = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'resource-log.json'])
```
ntalluri: what is the resource-log.json. What is inside it?

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
-            params.pop('_spras_run_name')
-        runner.run(wildcards.algorithm, inputs, output.pathway_file, params, container_settings)
+        if '_spras_run_name' in algorithm_params:
+            algorithm_params.pop('_spras_run_name')
+        try:
+            runner.run(wildcards.algorithm, inputs, output.pathway_file, algorithm_params, container_settings, params.timeout)
+            Path(output.resource_info).write_text(json.dumps({"status": "success"}))
+        except TimeoutError as err:
+            # We don't raise the error here (and use `--keep-going` to avoid re-running this rule [or others!] unnecessarily.)
+            Path(output.resource_info).write_text(json.dumps({"status": "error", "type": "timeout", "duration": params.timeout}))
```
ntalluri: So we do have a file that explains the type of errors that cause the empty files. This makes sense to me as long as the html file is easy to read.

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
+            # and we touch pathway_file still: Snakemake doesn't have optional files, so
```
ntalluri:
```suggestion
            # and we touch pathway_file still: Snakemake doesn't have optional files, so we output an output file.
```

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
+            # we'll filter the ones that didn't time out by passing around empty files.
```
ntalluri:
```suggestion
            # we filter for the successful files  (such as ones that didn't time out) with the filter_successful fucntion.
```

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
+            Path(output.pathway_file).touch()
 
 # Original pathway reconstruction output to universal output
 # Use PRRunner as a wrapper to call the algorithm-specific parse_output
 rule parse_output:
-    input: 
-        raw_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt']),
+    input:
+        raw_file = rules.reconstruct.output.pathway_file,
```
ntalluri, line 318: can you explain why this has to be updated and how this is working? I'm not sure what the rules.reconstruct are.

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
+        resource_info = rules.reconstruct.output.resource_info,
         dataset_file = SEP.join([out_dir, 'dataset-{dataset}-merged.pickle'])
     output: standardized_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'pathway.txt'])
     run:
```
ntalluri, line 321: can you add a one line comment on what this is specifically doing

Snakefile
```diff
@@ -268,25 +287,47 @@ rule reconstruct:
...
+        resource_info = json.loads(Path(input.resource_info).read_bytes())
+        if resource_info["status"] != "success":
+            mark_error(output.standardized_file)
```
ntalluri: I'm confused why we need mark_error when we have the line 312 doing something similar.

Snakefile
```diff
@@ -308,7 +349,7 @@ rule viz_cytoscape:
     output: 
         session = SEP.join([out_dir, '{dataset}-cytoscape.cys'])
     run:
-        cytoscape.run_cytoscape(input.pathways, output.session, container_settings)
+        cytoscape.run_cytoscape(filter_successful(input.pathways), output.session, container_settings)
```
ntalluri, line 347: is there a way we can do a one time filter_successful(input.pathways) somewhere above?

ntalluri (commented):
Here is my first round review. I like the empty files if an error occurs, and it is good to have an associated log explaining why.
(Adding this comment again here) My only issue with this is the definition of "error." If a parameter combination fails the heuristics check, I do not want the output to be empty. I want the output to reflect what was actually produced, so I do not have to rerun that combination even though it "failed" the heuristics. I should be able to freely update the heuristics and have that output counted if it now passes, without rerunning combinations that previously produced empty output. In short, a parameter combination failing the heuristics should not be classified as an error as defined in this PR.

This PR not working with singularity is a very big problem because of the chtc integration. 

"ML requires at least one pathway, and failing pathways can break ML-work. How do we want to handle downstream analysis when errors occur." This seems like a separate problem that I will fix internally in the ML code (I want to still make figures if we only have one pathway or a set of empty pathways).

config/config.yaml
ntalluri: based on my review I thought everything needs to be in seconds.

spras/analysis/cytoscape.py
tristan-f-r: An extra timeout parameter.

spras/prm.py
tristan-f-r: We use this since our `config.py` logic still treats parameter arguments as dictionaries. This could be removable entirely, but that's out of the scope of this PR. (This should be flagged as an issue)

Snakefile
tristan-f-r: Yes - I didn't notice that I actually added that when I spoke with you. I'm a little worried about the `--keep-going` flag, as I believe that's outdated? I'll look into this.

Snakefile
tristan-f-r: [I'll comment about this in the documentation comment above]

spras/config/config.py
tristan-f-r: (I'll mention more in your comment "I thought everything needs to be in seconds") but this is `parse` from the python library [pytimeparse](https://pypi.org/project/pytimeparse/).

config/config.yaml
tristan-f-r: [pytimeparse](https://pypi.org/project/pytimeparse/) allows strings such as `1d2h32m`, but I never mentioned this anywhere: I'll write up some docs :+1:

spras/prm.py
tristan-f-r: Actually, after looking into this further, I don't think this admits an obviously clean refactor :(

Snakefile
tristan-f-r: We don't need the flag as the code currently stands, and I've replaced that comment :+1:

Snakefile
tristan-f-r: 312 should be using `mark_error` as well [presumably I accidentally forgot to refactor old code after making `mark_error`].

Snakefile
tristan-f-r: That's a snakemake refactor: instead of referring to the file name by path, we just refer to it via its rule. I just commented about resource_info below.

Snakefile
tristan-f-r: I unfortunately can't figure it out if such a way does exist, but I'm also wary to say that Snakemake can't do that either.

tristan-f-r:
> My only issue with this is the definition of "error." If a parameter combination fails the heuristics check, I do not want the output to be empty. I want the output to reflect what was actually produced, so I do not have to rerun that combination even though it "failed" the heuristics.

The small part here that is adaptable is that errors, in this PR, are only made in the `reconstruct` rule. We can define other errors in some `heuristics` rule, and we can rewire other rules to depend on the success of heuristics instead of on reconstruct (via that `resource_info = rules.reconstruct.output.resource_info` input rule above, but instead we would say `rules.heuristics.output.resource_info` instead.)

tristan-f-r:
This works with apptainer now, but this now touches an untested part of profiling, so that part needs a review from @jhiemstrawisc.

docs/design/errors.rst
```diff
@@ -0,0 +1,21 @@
+########
+ Errors
+########
+
+By default, whenever SPRAS runs into a container error (i.e. an internal
+algorithm error), the full workflow will fail. However, there are
+certain designated errors where we don't want this to be the case (at
```
ntalluri: can we make this a separate section that explains each error instead of putting it in parentheses (even though right now it is only timeouts). We can expand this in the future.

docs/timeout.rst
```diff
@@ -0,0 +1,20 @@
...
+timeout:
+
+.. code:: yaml
+
+   - name: "allpairs"
+     include: true
+     timeout: 1d
+
+The timeout string parsing is delegated to `pytimeparse
+<https://pypi.org/project/pytimeparse/>`__, which allows for more
```
ntalluri:
```suggestion
<https://pypi.org/project/pytimeparse/>`__, (examples linked here). This allows for more
```

docs/timeout.rst
```diff
@@ -0,0 +1,20 @@
+##########
+ Timeouts
+##########
+
+SPRAS allows for per-algorithm timeouts, specified under the global
```
ntalluri:
```suggestion
SPRAS allows for optional per-algorithm timeouts, specified under the global
```

ntalluri: can you also add a sentence on what happens when timeout is not included?

spras/containers.py
```diff
@@ -166,6 +167,20 @@ def streams_contain(self, needle: str):
...
         return self.message
 
+class TimeoutError(RuntimeError):
+    """Raises when a function times out."""
+    timeout: int
+    message: str
+
+    def __init__(self, timeout: int, *args):
+        self.timeout = timeout
+        self.message = f"Timed out after {timeout}s."
```
ntalluri: could we convert this back into other time scales (if I put 5.6 days I don't want to have to read that as 483840 seconds). 
```suggestion
        self.message = f"Timed out after {timeout} seconds."
```

ntalluri: f"Timed out after {timeout} seconds; {hours} hours; {days} days ."

Something like this?

ntalluri: (only do this if it is easy to do)

spras/containers.py
```diff
@@ -345,7 +417,16 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
     return out
 
 
-def run_container_singularity(container: str, command: List[str], volumes: List[Tuple[PurePath, PurePath]], working_dir: str, out_dir: str, config: ProcessedContainerSettings, environment: Optional[dict[str, str]] = None):
+def run_container_singularity(
```
ntalluri, line 421: can you give me a couple sentence explanation on how you got singularity to work and what parts of the code you had to change to make it work?

ntalluri (commented):
here is half of a review on this pr

spras/containers.py
tristan-f-r: The small refactor here is that Justin Hiemstra already had manually ran `apptainer` instead of using the `spython` interface to run apptainer containers _for profiling_, so I instead removed the usage of `spython` execute and directly inserted the unix `timeout` command instead for timing out containers for all apptainer runs, including those that aren't profiling.

[This is why the diff removes a conditional statement.]

spras/containers.py
tristan-f-r: I'll do this with `datetime`. It doesn't create an intuitive string (e.g., 100 seconds produces the output `0:01:40`), but it is also a native Python library, so we can at least avoid some of the annoying English string-fiddling (i.e., the `if days == 1 then '' else 's'`) and conform to how other Python CLIs handle duration representation as strings.

tristan-f-r:
I've made a few changes:
- With your line of thought of considering errors as important objects in documentation, I've pushed a commit that moves out some of the error object handling out from `Snakefile` into an `errors.py`, and we further use pydantic to make sure we avoid arbitrary JSON handling (also to give SPRAS automation users a way to easily parse our error objects). This was also motivated by encountering a type error with `mark_error`, which is why I've opted to make this type-safe.
- I've also renamed `resource` -> `artifact`, as I found the former a little more confusing, though the latter isn't an optimal name either: see the top-level comment on `errors.py`.
- `timeout` is now per-run, with optional specification for an entire algorithm. This refactor was intentionally done for conditionals, though it can also find its use with many other configuration settings. This also introduces a new key, `params` under all runs. We mark the PR as breaking because of this.

The final bullet point here also means that the conditionals PR will now actually depend on this PR.

ntalluri:
does this need to use #464?

tristan-f-r:
No - I was thinking that the `RunSettings` here can be more generally applicable to #464.

tristan-f-r:
I seem to have messed with imports when I went to pass in the entirety of `run_settings` instead of just `timeout`. I'll fix this :+1:.

config/config.yaml
```diff
@@ -73,65 +73,75 @@ algorithms:
     include: true
     runs:
       run1:
-        k: range(100,201,100)
+        params:
```
ntalluri, line 76: is there a reason why we are getting rid of run1, run2, ...?

tristan-f-r: Oh! We aren't. I actually don't think I mentioned this in the timeout RTD, so I'll write it there instead of talking about it here.

ntalluri: Yeah, if you can update the documentation on this, that would be helpful.

tristan-f-r: Not much to mention, unfortunately, since I am also just realizing that we currently only have `timeout` as a useful non-algorithm-parameter setting, but I've at least updated the documentation to reflect the use of this structure.

config/egfr.yaml
```diff
@@ -31,77 +31,89 @@ algorithms:
     include: true
     runs:
       run1:
-        k:
-          - 10
-          - 20
-          - 70
+        params:
```
jhiemstrawisc, line 34: Just a general note that changing the config file like this constitutes a breaking change -- it will require a new spras major release. Is it strictly necessary?

Snakefile
```diff
@@ -270,25 +281,48 @@ rule reconstruct:
...
+        try:
+            runner.run(detach_spras_revision(_config.config.immutable_files, wildcards.algorithm), inputs, output.pathway_file, algorithm_params, container_settings, params.timeout)
+            mark_success(output.artifact_info)
+        except TimeoutError as err:
+            # We don't raise the error here (analogous to `--keep-going`, except we avoid unnecessarily re-running this rule.)
+            mark_error(output.artifact_info, TimeoutArtifactError(duration=params.timeout))
+            # and we touch pathway_file still: Snakemake doesn't have optional files, so we output a 'artifact info' file,
+            # which contains the status (success/failure) of specific Snakemake jobs.
+            # We filter for the successful files (such as ones that didn't time out) with the `filter_successful` function.  
+            Path(output.pathway_file).touch()
```
jhiemstrawisc, line 310: What are the consequences to the user here? This seems like it has the potential to confuse someone by producing empty pathway files alongside a new error reporting mechanism outside of Snakemake.

Snakefile
```diff
@@ -270,25 +281,48 @@ rule reconstruct:
...
     input:
-        raw_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt']),
+        # We propagate up the artifact_info error if it exists.
+        artifact_info = rules.reconstruct.output.artifact_info,
+        raw_file = rules.reconstruct.output.pathway_file,
         dataset_file = SEP.join([out_dir, 'dataset-{dataset}-merged.pickle'])
     output: standardized_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'pathway.txt'])
     run:
+        if is_error(input.artifact_info):
+            mark_error(output.standardized_file)
```
jhiemstrawisc: Is this call to `mark_error` missing an argument?

spras/errors.py
```diff
@@ -0,0 +1,76 @@
...
+# Collection of Snakemake utilities.
+
+def artifact_info_to_str(artifact_info: ArtifactInfo) -> str:
+    """Converts some `ArtifactInfo` into a string."""
+    return json.dumps(ArtifactInfoAdapter.validate_python(artifact_info).model_dump(mode='json'))
+
+def artifact_info_from_file(file: LoosePathLike) -> ArtifactInfo:
+    """Converts a file into ArtifactInfo."""
+    with open(file, 'r') as f:
+        return ArtifactInfoAdapter.validate_json(json.load(f))
```
jhiemstrawisc: Will this work? `json.load` should produce a dict, but `validate_json` expects raw JSON.

spras/errors.py
```diff
@@ -0,0 +1,76 @@
+"""
```
jhiemstrawisc, line 1: Can you add unit test coverage for the new stuff in this file?

spras/containers.py
```diff
@@ -193,15 +220,25 @@ def run_container(container_suffix: str, command: List[str], volumes: List[Tuple
 
     container = container_settings.prefix + "/" + container_suffix
     if normalized_framework == 'docker':
-        return run_container_docker(container, command, volumes, working_dir, environment, network_disabled)
+        return run_container_docker(container, command, volumes, working_dir, environment, timeout, network_disabled)
     elif normalized_framework == 'singularity' or normalized_framework == "apptainer":
         return run_container_singularity(container, command, volumes, working_dir, out_dir, container_settings, environment)
```
jhiemstrawisc: I think this is missing the timeout.

spras/containers.py
```diff
@@ -290,13 +348,28 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
+        command,
+        volumes=bind_paths,
+        working_dir=working_dir,
+        network_disabled=network_disabled,
+        environment=environment,
+        detach=True
+    )
+
+    try:
+        container_obj.wait(timeout=timeout)
```
jhiemstrawisc, line 365: I brushed up on the [docs for this API](https://docker-py.readthedocs.io/en/stable/containers.html?highlight=wait#docker.models.containers.Container.wait) -- it seems like `wait()` should be returning the container's ultimate status:
```
Returns: The API’s response as a Python dictionary, including
    the container’s exit code under the StatusCode attribute.
```
Should we be checking it?

spras/containers.py
```diff
@@ -290,13 +348,28 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
+    except requests.exceptions.ReadTimeout as err:
+        container_obj.stop()
+        client.close()
+        if timeout: raise TimeoutError(timeout) from err
+        else: raise RuntimeError("Timeout error but no timeout specified. Please file an issue with this error and stacktrace at https://github.com/Reed-CompBio/spras/issues/new.") from None
```
jhiemstrawisc, line 370: Should this catch other types of errors?

spras/containers.py
```diff
@@ -290,13 +348,28 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
...
+
+    out = container_obj.attach(stderr=True).decode('utf-8')
```
jhiemstrawisc, line 372: I think the completion of `wait()` implies the container is already stopped (either of its own accord or because the timeout was reached). What happens to this attach command if the container is already stopped?

spras/containers.py
```diff
@@ -417,37 +500,44 @@ def run_container_singularity(container: str, command: List[str], volumes: List[
...
-        cmd = [wrapper, my_cgroup] + singularity_cmd
-        proc = subprocess.run(cmd, capture_output=True, text=True, stderr=subprocess.STDOUT)
+        cmd = [wrapper, my_cgroup] + cmd
+    if timeout is not None:
+        cmd = ["timeout", f"{timeout}s"] + cmd
+    proc = subprocess.run(cmd, text=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+    # As per unix `timeout`, this is the status if the command times out and --preserve-status is not initially specified
+    # (where the latter above holds).
+    if proc.returncode == 124:
```
jhiemstrawisc, line 529: Should this have handling for other error codes?

jhiemstrawisc (commented):
A few comments sprinkled throughout. The bigger issue for me is that this doesn't appear to run in its current state. Using a relatively stock `config/config.yaml` I get a runtime error:
```
RuleException:
AttributeError in file "/Users/jhiemstra/Desktop/dev/spras/spras/Snakefile", line 302:
'int' object has no attribute 'timeout'
  File "/Users/jhiemstra/Desktop/dev/spras/spras/Snakefile", line 302, in __rule_reconstruct
  File "/Users/jhiemstra/Desktop/dev/spras/spras/spras/runner.py", line 58, in run
  File "/Users/jhiemstra/Desktop/dev/spras/spras/spras/prm.py", line 81, in run_typeless
  File "/Users/jhiemstra/Desktop/dev/spras/spras/spras/pathlinker.py", line 119, in run
Exiting because a job execution failed. Look below for error messages
WorkflowError:
At least one job did not complete successfully.
```

config/egfr.yaml
tristan-f-r: I'm moving this now to also allow room for conditional runs, or something of the form:

```
alg1:
  runs:
    run1:
      if: [...]
      params:
```

This also has the nice consequence of run-specific timeouts, so I've included it in this PR because of that. [I've also marked this PR as breaking as per the `!` in the title.]

Snakefile
tristan-f-r: This does - we have no good way to log this. Collecting these errors with Snakemake's report system (whose base code is in #434) would help to make this easier, but it wouldn't fully solve it.

I would have liked for Snakemake itself to come with recoverable errors, but its `--keep-going` flag is incredibly limited.

PR #458, opened 2026-01-15
fix: don't have overlapping jaccard similary matrix text

tristan-f-r:
10 is notably a magic value.

jaccard-heatmap
jaccard-heatmap-second


agitter (approved):
This makes sense to me. Let's give @ntalluri a chance to review before merging. We can mention it in Slack if this gets lost.

PR #459, opened 2026-01-23
Update htcondor instructions

jhiemstrawisc:
This largely reformats the directory structure needed to run SPRAS workflows with HTCondor. In particular, it moves a lot of the helper code/submit files out of `docker-wrappers/SPRAS/` into a top-level `htcondor/` directory. I can do this now that the HTCondor executor has matured significantly, and can handle all the paths as they're configured in this diff.

To run a test SPRAS workflow, try following along with the instructions in `docs/htcondor.rst`. If anything is confusing, or you get hung up on any of the steps, let's discuss what I can do to make things more clear.


tristan-f-r (commented):
[I'll have to restore my HTCondor access to follow this.]

tristan-f-r (commented):
Some code nitpicks

docs/htcondor.rst
```diff
@@ -63,64 +63,54 @@ image does not use a "v" in the tag.
...
-.. code:: bash
-
-   cp ../../Snakefile . && \
-   cp -r ../../input .
-
-Instead of editing ``spras.sub`` to define the workflow, this scenario
-requires editing the SPRAS profile in ``spras_profile/config.yaml``.
-Make sure you specify the correct container, and change any other config
-values needed by your workflow (defaults are fine in most cases).
+3. Instead of editing ``spras.sub`` to define the workflow, this scenario
```
agitter: This language does a good job with respect to my comment above about making it explicit which files are used where. It *may* still be useful to itemize that all in one place.

htcondor/spras.sub
```diff
@@ -21,24 +21,25 @@ SNAKEFILE = ../../Snakefile
 # first                                                    #
 ############################################################
 universe = container
-container_image = <your spras image>.sif
-# container_image = docker://reedcompbio/spras:v0.6.0
+#container_image = <your spras image>.sif
+container_image = instructions-overhaul.sif
+# container_image = docker://reedcompbio/spras:v0.2.0
```
agitter: Do we need both of the commented lines?

agitter (commented):
I'm testing the Snakemake long execution mode. The first time my jobs went on hold because I put my `spras-v0.6.0.sif` file in the `htcondor/` directory instead of the root directory. That should have been obvious based on the comment in the `.yaml` file.

On the second attempt my jobs went on hold with
```
Transfer output files failure at execution point slot1_24@e2591.chtc.wisc.edu while sending files to access point ap2001. Details: 1 total failures: first failure: reading from file /var/lib/condor/execute/slot1/dir_3699332/scratch/output: (errno 2) No such file or directory
```

jhiemstrawisc:
I converted this to a draft because these docs will depend on the explicit sif transfer PR, and I haven't yet tested everything here in that paradigm.

jhiemstrawisc:
Also, apologies for the poor git etiquette in the last commit that rolled too many things into one diff (including running an `rst` formatter). I think some of the updates are sufficiently large that the whole file should more or less be re-assessed as a fresh document.

htcondor/spras.sub
```diff
@@ -21,24 +21,25 @@ SNAKEFILE = ../../Snakefile
...
 ############################################################
 # Specify names for log/stdout/stderr files generated by   #
 # HTCondor.                                                #
 # NOTE: You should `mkdir logs/` before running, or the    #
 # spras_$(Cluster).log file won't be available.            #
 ############################################################
-log = logs/spras_$(Cluster)_$(Process).log
-output = logs/spras_$(Cluster)_$(Process).out
-error = logs/spras_$(Cluster)_$(Process).err
+log = htcondor/logs/spras_$(Cluster)_$(Process).log
```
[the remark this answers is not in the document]
jhiemstrawisc: I think one per cluster/process is the right way to go. In theory, one could still `queue N>1`.

docs/htcondor.rst
```diff
@@ -63,64 +63,54 @@ image does not use a "v" in the tag.
 Submitting All Jobs to a Single EP
```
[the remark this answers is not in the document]
jhiemstrawisc: Addressed in a new commit.

docs/htcondor.rst
jhiemstrawisc: Addressed in a new commit.

jhiemstrawisc:
This is a note for myself -- one thing I should document in the htcondor rst is the need to pre-create apptainer images before launching workflows.

agitter (commented):
During my testing, I triggered a Snakemake lock error by launching a long job, killing it, changing the config file, and relaunching. That may be a common error.
```
python3.11/site-packages/snakemake/persistence.py", line 211, in lock
    raise snakemake.exceptions.LockException()
snakemake.exceptions.LockException: Error: Directory cannot be locked. Please make sure that no other Snakemake process is trying to create the same files in the following director
y:
/home/agitter/spras
If you are sure that no other instances of snakemake are running on this directory, the remaining lock was likely caused by a kill signal or a power loss. It can be removed with th
e --unlock argument.

LockException:
Error: Directory cannot be locked. Please make sure that no other Snakemake process is trying to create the same files in the following directory:
/home/agitter/spras
If you are sure that no other instances of snakemake are running on this directory, the remaining lock was likely caused by a kill signal or a power loss. It can be removed with th
e --unlock argument.
```
Should we add it to troubleshooting?

I also hit this error
```
$ cat htcondor/logs/merge_input/merge_input-5_7645955.err
ModuleNotFoundError in file "/var/lib/condor/execute/slot1/dir_1046931/scratch/Snakefile", line 8:
No module named 'spras.config.revision'
  File "/var/lib/condor/execute/slot1/dir_1046931/scratch/Snakefile", line 8, in <module>
```
I'm guessing that means a need a newer version of the SPRAS sif image. However, we haven't released a SPRAS version recently. What version of the image are you testing with?

ntalluri:
@agitter 

For the first bug, this requires the spras conda environment to be activated and then the command `snakemake --configfile <path to config file> --unlock` to be run anytime the config file of choice is updated. I was planning on commenting this as step five for the parallel jobs.

For the second issue, you are right, this is because the version of SPRAS in the docker image v0.6 isn't up to date with the current version of SPRAS. Justin has a docker image you can pull from dockerhub  (i think it is this jhiemstra/spras:update-htcondor-instructions-v2) or you will need to build the image with Docker on your local machine of the updated version of SPRAS, push the image to Docker Hub, and then use that image.

jhiemstrawisc:
> For the second issue, you are right, this is because the version of SPRAS in the docker image v0.6 isn't up to date with the current version of SPRAS.

In general, you should either:
- always rebuild the SPRAS container to match the version repo you're working with, OR
- check out the repo at a specific release (e.g. `git checkout 0.6.0`) to match the container you want to use

The key is that the repo you're using to submit from the AP should match what's in the image.

There's a callout relatively early in the documentation covering this, but I'm open to edits it seems like this is often missed:
```
   It is best practice to make sure that the Snakefile you copy for your
   workflow is the same version as the Snakefile baked into your
   workflow's container image. When this workflow runs, the Snakefile
   you just copied will be used during remote execution instead of the
   Snakefile from the container. As a result, difficult-to-diagnose
   versioning issues may occur if the version of SPRAS in the remote
   container doesn't support the Snakefile on your current branch. The
   safest bet is always to create your own image so you always know
   what's inside of it.
```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+
+   .. code:: yaml
+
+      containers:
+        framework: singularity
+        unpack_singularity: true
+        images:
+          omicsintegrator1: "images/omics-integrator-1_v2.sif"
+          pathlinker: "images/pathlinker_v2.sif"
+
```
ntalluri, line 234: Could you add to this yaml block to show what you mean by the names having to match what is in the config.

```suggestion
...
algorithms:
  - name: "pathlinker"
    include: true
    runs:
      run1:
        k: 10
  - name: "omicsintegrator1"
    include: true
    runs:
      run1:
        b: 5
        w: 0
        d: 10
```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
-the previous section, but with two additions. First, it requires an
-activated SPRAS conda environment with a ``pip install``-ed version of
-the SPRAS module (via ``pip install .`` inside the SPRAS directory).
+Parallelizing SPRAS workflows with HTCondor requires much of the same
+setup as the previous section, but with two additions.
 
-Second, it requires an experimental executor for HTCondor that has been
-forked from the upstream `HTCondor Snakemake executor
-<https://github.com/htcondor/snakemake-executor-plugin-htcondor>`__.
+#. Build/activate the SPRAS conda/mamba environment and ``pip install``
```
ntalluri: Could you add a note block to a link on how to set up the conda environment on an ap?

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+efficient than using HTCondor's distributed capabilities. This approach
+is best suited for workflows that are not computationally intensive, or
+for testing and debugging purposes.
 
-.. code::
+Before submitting all SPRAS jobs to a single remote Execution Point
+(EP), you'll need to set up three things:
+
+#. You'll need to modify ``htcondor/spras.sub`` to point at your
+   container image, along with any other configuration changes you want
```
ntalluri: I got confused about what specific image this was after reading about all the image stuff above.

```suggestion
   SPRAS container image, along with any other configuration changes you want
```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+   to make like choosing a logging directory or toggling OSPool
```
ntalluri: can you add a note on what toggling the OSPool submission looks like for a user in the spras.sub. 

.. note::

In spras.sub uncomment +WantGlideIn = true and requirements = (HAS_SINGULARITY == True) && (Poolname =!= "CHTC")

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+   ``htcondor/spras_profile/config.yaml``. Make sure you specify the
+   correct container, and change any other config values needed by your
+   workflow (defaults are fine in most cases).
 
-Instead of editing ``spras.sub`` to define the workflow, this scenario
-requires editing the SPRAS profile in ``spras_profile/config.yaml``.
-Make sure you specify the correct container, and change any other config
-values needed by your workflow (defaults are fine in most cases).
+#. Modify your SPRAS configuration file to set ``unpack_singularity:
+   true`` and ``containers.framework: singularity``.
```
ntalluri, line 376: To deal with the --unlock bug 
```suggestion
#. Activate the spras conda environment and run the command ``snakemake --configfile <path to config file> --unlock``

... note::

Whenever you change the config file, run ``snakemake --configfile <path to config file> --unlock`` before submitting jobs. Otherwise, the workflow will appear to complete immediately but is actually raising a ``snakemake.exceptions.LockException()``.

```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
-.. code:: bash
+      pip install snakemake-executor-plugin-htcondor
 
-   cp ../../Snakefile . && \
-   cp -r ../../input .
+#. Instead of editing ``spras.sub`` to define the workflow, this
+   scenario requires editing the SPRAS profile in
+   ``htcondor/spras_profile/config.yaml``. Make sure you specify the
+   correct container, and change any other config values needed by your
+   workflow (defaults are fine in most cases).
```
ntalluri:
```suggestion
   workflow (defaults are fine in most cases). Memory and hardware requirements are also set here. To use a config file other than config/config.yaml, set the path next to the configfile: variable in this file.
```

htcondor/spras_profile/config.yaml
ntalluri, line 1: this is what my config.yaml looks like. I wasn't sure if we need to add:

> ... && versionGE(split(Target.CondorVersion)[1], "24.8.0") && (isenforcingdiskusage =!= true)' 
>  stream_ouput: true
> stream_error: true

parts to the documentation



```
# Default configuration for the SPRAS/HTCondor executor profile. Each of these values
# can also be passed via command line flags, e.g. `--jobs 30 --executor htcondor`.

# NOTE: File paths in here should be relative to where you submit from, typically the
# root of the SPRAS repository

# 'jobs' specifies the maximum number of HTCondor jobs that can be in the queue at once.
jobs: 30
executor: htcondor
configfile: config/egfr.yaml
htcondor-jobdir: htcondor/logs

# Indicate to the plugin that jobs running on various EPs do not share a filesystem with
# each other, or with the AP.
shared-fs-usage: none
# Distributed, heterogeneous computational environments are a wild place where strange things
# can happen. If something goes wrong, try again up to 2 times. After that, we assume there's
# a real error that requires user/admin intervention
retries: 2

# Default resources will apply to all workflow steps. If a single workflow step fails due
# to insufficient resources, it can be re-run with modified values. Snakemake will handle
# picking up where it left off, and won't re-run steps that have already completed.
default-resources:
  job_wrapper: "htcondor/spras.sh"
  # If running in CHTC, this only works with apptainer images
  # Note requirement for quotes around the image name
  container_image: "test-htc.sif"
  universe: "container"
  # The value for request_disk should be large enough to accommodate the runtime container
  # image, any additional PRM container images, and your input data.
  request_disk: "16GB"
  request_memory: "12GB"
  retry_request_memory_increase: "RequestMemory + 4"
  retry_request_memory_max: "32GB"
  classad_WantGlideIn: true
  requirements: |
    '(HAS_SINGULARITY == True) && (Poolname =!= "CHTC") && versionGE(split(Target.CondorVersion)[1], "24.8.0") && (isenforcingdiskusage =!= true)' 
  stream_ouput: true
  stream_error: true
```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+.. warning::
+
+   Do not run ``apptainer build`` (or otherwise pull/convert large
+   images) directly on an Access Point. APs are shared, login-style
+   nodes, and image builds are resource-intensive enough that doing so
+   is discouraged and may violate your pool's usage policies. Instead,
+   build images inside an interactive job on an Execution Point. If
+   you're working at CHTC, follow their guide for building Apptainer
+   images in an interactive job:
+   https://chtc.cs.wisc.edu/uw-research-computing/apptainer-htc.html
```
ntalluri, line 152:
```suggestion
   https://chtc.cs.wisc.edu/uw-research-computing/apptainer-htc.html. Specifically, create the apptainer.sub file on the AP and run ``condor_submit -i apptainer.sub`` on the AP.
```

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
-
-.. code:: bash
+      pip install snakemake-executor-plugin-htcondor
 
-   cp ../../Snakefile . && \
-   cp -r ../../input .
+#. Instead of editing ``spras.sub`` to define the workflow, this
+   scenario requires editing the SPRAS profile in
+   ``htcondor/spras_profile/config.yaml``. Make sure you specify the
+   correct container, and change any other config values needed by your
```
ntalluri:
```suggestion
   correct SPRAS container image, and change any other config values needed by your
```

docs/htcondor.rst
```diff
@@ -2,41 +2,95 @@
...
+   Images to Apptainer/Singularity Images`_ for instructions.
+   Importantly, the Apptainer image must be built for the linux/amd64
+   architecture. Most HTCondor APs will have ``apptainer`` installed,
+   but they may not have ``docker``. If this is the case, you can build
+   the image with Docker on your local machine, push the image to Docker
+   Hub, and then convert it to Apptainer's ``sif`` format on the AP.
+
+.. tip::
+
+   It is best practice to make sure that the Snakefile you copy for your
```
ntalluri: Potential update for this tip::

To avoid versioning issues, the Snakefile copied for your workflow in SPRAS should match the Snakefile baked into the SPRAS container image. When this workflow runs, the Snakefile you just copied will be used during remote execution instead of the Snakefile from the container. A mismatch between the repo version and the container can cause difficult-to-diagnose errors, including ``ModuleNotFoundError``.

To keep these in sync, either:

- rebuild the SPRAS container locally, push it to Docker Hub, and use that image for submitting jobs, or
- or, check out the SPRAS repo at the release matching your container

ntalluri (commented):
The updated documentation looks great, I added some suggestions on how to help users more. 

I also remember we were changing the config.yaml file in spras_profile and wasn't sure if any of the commands needed to be added to the documentation.

htcondor/spras_profile/config.yaml
ntalluri, line 32: do we need to add request_cpus = $(NUM_PROCS)?

docs/htcondor.rst
agitter: I thought the instructions in the .sub file comments were clear enough, but I'm not totally opposed to this.

docs/htcondor.rst
```diff
@@ -63,69 +117,192 @@ After running this command, a new file called ``spras-v0.6.0.sif`` will
...
+
+#. Build an Apptainer ``.sif`` image for each algorithm you intend to
+   run, placing each one in ``images/``. As with the SPRAS runtime
+   image, these must be built for the ``linux/amd64`` architecture. For
+   example, to pre-build the Omics Integrator 1 and PathLinker images:
+
+   .. code:: bash
+
+      apptainer build images/omics-integrator-1_v2.sif docker://reedcompbio/omics-integrator-1:v2
+      apptainer build images/pathlinker_v2.sif docker://reedcompbio/pathlinker:v2
```
agitter, line 211: People might run this directly on the AP if we don't remind them to use a build job again.

docs/htcondor.rst
```diff
@@ -2,41 +2,95 @@
...
+      -  ✓
+      -  Launches Snakemake as a long-running managed job so the
+         workflow survives terminal disconnects.
+
+   -  -  ``run_htcondor.sh``
+      -
+      -  ✓
+      -  Convenience wrapper (in the repository root) around
+         ``snakemake_long.py``.
+
```
agitter, line 116: The next section is what I found confusing. It gives instructions to create the .sif from the existing DockerHub image. That usually breaks. I recommend we remove it and only give instructions to build a new image from source.

docs/htcondor.rst
jhiemstrawisc: I can link back to SPRAS's own instructions for building/activating the environment. There's nothing special about doing this step on an AP.

docs/htcondor.rst
jhiemstrawisc: In a new commit, I reference "to point at your general spras container image (built from ``docker-wrappers/SPRAS``)"

docs/htcondor.rst
jhiemstrawisc: I'll make a note linking to the specific instructions in the `.sub` file so that there aren't two sets of instructions to maintain.

docs/htcondor.rst
jhiemstrawisc: Took your advice, and I'm also adding a note that explains the difference between the spras profile "config.yaml" and the general spras workflow "config.yaml"

htcondor/spras_profile/config.yaml
jhiemstrawisc: I definitely don't think we should be covering the `stream_{output/error}` in the general purpose profile for fear that an unknowing user will take that as the default (and make a CHTC sys admin very sad when they crash an AP). These should be intentionally hard to fine because users will find them very tempting to use without understanding the detrimental effects they can have on shared computing resources.

As for the other requirements, I'd also like to avoid sticking those in the profile -- they're very specific to your run that needs both the OSPool and profiling, and these requirements are generally documented elsewhere.

docs/htcondor.rst
jhiemstrawisc: Done.

docs/htcondor.rst
jhiemstrawisc: I expanded this section in an upcoming commit.

htcondor/spras_profile/config.yaml
jhiemstrawisc: I don't think so -- the `$NUM_PROCS` macro is specific to the single-EP mode launched by `spras.sub`, so it wouldn't work here. In this mode, each step is launched as a single core job. If a user decides they want each job to launch with added CPUs, they can set the value here, but the default of 1 is good enough IMO.

docs/htcondor.rst
jhiemstrawisc: I'm adding a more explicit warning here in an upcoming commit.

docs/htcondor.rst
jhiemstrawisc: Maybe you can explain how these steps break for you? I don't think I've ever run into issues building from an existing Dockerhub image, outside of the mismatched Snakefile conundrum (which I try to cover more heavily in my latest round of revisions).

htcondor/spras_profile/config.yaml
agitter: What was the goal of these settings for Neha's OSPool runs? I don't think we need to add them here but am curious for our SPRAS benchmarking in OSPool.

jhiemstrawisc: - `classad_WantGlideIn: true` configures a classad that enables submission to OSPool
- `(HAS_SINGULARITY == True)` makes sure you land on OSPool EPs that support apptainer/singularity, which is a hard requirement for SPRAS
- `(Poolname =!= "CHTC")` -- this can probably be omitted -- it disables submission to CHTC EPs. If you enable OSPool submissions and don't include this, you're submitting to both the OSPool and CHTC at once.
- `versionGE(split(Target.CondorVersion)[1], "24.8.0") && (isenforcingdiskusage =!= true)` makes sure you match with EPs that have the features/configuration needed for the apptainer profiling code to work.

docs/htcondor.rst
agitter: The new option to checkout a version of the SPRAS repo that matches the existing image makes sense and should help.

My prior confusion was around what a user should do if they need to build a SPRAS image themselves to match a newer commit. I don't see a way to build the Apptainer sif image entirely on CHTC. We link to [CHTC docs](https://chtc.cs.wisc.edu/uw-research-computing/htc-docker-to-apptainer.html) that expect you to have a def file, which we don't have. The Apptainer [instructions](https://apptainer.org/user-docs/master/build_a_container.html#building-containers-from-dockerfile-with-buildkit) for building directly from a Dockerfile didn't work for me. My understanding is that I would have to build a Docker image on a local machine with Docker, push it to my DockerHub, then run a build job in CHTC to convert it to a sif file.

agitter (approved):
Looks ready to me. I'll let Neha check it as well.

ntalluri (approved):
This also looks good to me.

Issue #462, opened 2026-03-03
Provide guidance on working around docker rate limiting for large CHTC runs


jhiemstrawisc:
I was helping @nisthapanda debug failures in one of her workflows, where Snakemake reported issues unpacking a singularity container image:
```
RuleException:
ContainerError in file "/var/lib/condor/execute/slot1/dir_3175487/scratch/spras/Snakefile", line 285:
An unexpected non-zero exit status (255) occurred while running this singularity container:
    FATAL:   While checking container encryption: could not open image /var/lib/condor/execute/slot1/dir_3175487/scratch/unpacked/omics-integrator-1_v2: failed to retrieve path for /var/lib/condor/execute/slot1/dir_3175487/scratch/unpacked/omics-integrator-1_v2: lstat /var/lib/condor/execute/slot1/dir_3175487/scratch/unpacked/omics-integrator-1_v2: no such file or directory
```

The real cause lived in the job's individual `.out` file (for some reason apptainer writes errors stdout, while Snakemake writes regular output to stderr... go figures):
```
FATAL:   While making image from oci registry: error fetching image to cache: while building SIF from layers: conveyor failed to get: GET https://index.docker.io/v2/reedcompbio/omics-integrator-1/manifests/v2: TOOMANYREQUESTS: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

This error occurs when the EP a job lands on tries to fetch an image from Docker but has already exceeded a rate limiting threshold. It is a consequence of trying to pull images inside the jobs instead of pre-declaring them as part of the submit description (I think this would be difficult to do with the SPRAS is designed, which is why I'm not proposing it as the solution).

I think there's probably a two-part fix here:
1. We should catch the error at fetch time instead of at open time for these images. That is, fail early where we might provide a more meaningful error message so users don't have to hunt through logs.
2. I should work on providing guidance for the SPRAS profile that would set an apptainer image repository cache. This can then be configured to point at whatever hostname CHTC uses, which would help us avoid this rate limiting issue.

agitter:
Thanks for diagnosing the root cause from the out file. Your proposed fixes make sense to me.

Would the image cache be CHTC specific and expose us to this problem in hypothetical future large-scale OSG runs? I'm wondering if a cache is sufficient or if we need to address the inability to transfer the image to the job more directly.

agitter:
Closed by #464

PR #464, opened 2026-03-11
Add override mechanism for algorithm containers, transfer explicit `.sif` images with HTCondor

jhiemstrawisc:
This PR aims to accomplish a few things:
1. It adds a new valid keys to config.yaml that let a user say "For reconstruction algorithm `foo`, I want you to use container image `bar`". In the general case, I'm not sure how much this is needed, but @agitter and I have floated the concept in the past, and it helps me solve 2. This implements a 4-tier hierarchy over how much of the image name is overridden (with special logic for `.sif` extensions), e.g.:
```
containers:
  registry:
    base_url: docker.io
    owner: reedcompbio
    
    images:
      omicsintegrator1: "oi1:latest"
      omicsintegrator2: "jhiemstra/oi2:latest
      mincostflow: "hub.opensciencegrid.org/jhiemstra/mincostflow:latest
      allpairs: "images/allpairs.sif"
```
would result in container URIs of
- `omicsintegrator1 --> docker.io/reedcompbio/oi1:latest`
- `omicsintegrator2 --> docker.io/jhiemstra/oi2:latest1`
- `mincostflow --> hub.opensciencegrid.org/jhiemstra/mincostflow:latest`
- `allpairs --> images/allpairs.sif (local file)`

The caveat here is that I don't believe _all_ container registries follow this kind of `<base_url>/<owner>/<container>` hierarchy, but it's at least true for docker, ghcr and hub.opensciencegrid.org. If a user finds themselves somewhere outside those, they can always declare the entire URI explicitly.

2. When a container override contains the `.sif` extension, the file is added to the reconstruct rule's `htcondor_transfer_input_files` resource key. This triggers HTCondor to transfer the sif image as part of the job's input sandbox.
3. When a `.sif` override is present and apptainer/singularity is the configured container framework, SPRAS uses the local file instead of pulling/building from a remote registry. This is combined with 2) are the key fixes for #462
4. Some of the container resolution logic started to get sufficiently nested with if/else logic that I wanted to split it out into a helper function I could more easily test. This is where most of the new test code comes from.
5. Finally, I fixed a small-but-inconsequential bug I noticed in the way we treated `singularity` and `apptainer` container frameworks differently in some cases. I added a helper function that makes it easier to treat these as synonyms.

Note that I haven't yet documented this guidance, as #462 requests -- I'd rather get this over the finish line first, then add the documentation to #459 (so I don't create conflicts for myself)


config/config.yaml
```diff
@@ -41,6 +41,29 @@ containers:
...
+  #
+  # Image reference (e.g., "pathlinker:v3"):
+  #   Prepends the registry prefix. Works with both Docker and Apptainer.
+  #
+  # Full image reference with registry (e.g., "ghcr.io/myorg/pathlinker:v3"):
+  #   Used as-is (prefix NOT prepended). Works with both Docker and Apptainer.
+  #
+  # Local .sif file path (e.g., "images/pathlinker_v2.sif"):
+  #   Apptainer/Singularity only. Skips pulling from registry and uses the
+  #   pre-built .sif directly. When running via HTCondor with shared-fs-usage: none,
```
agitter: Because `shared-fs-usage: none` isn't in this config file, it could help to state the place where is it set (the spras_profile config).

spras/config/container_schema.py
```diff
@@ -34,17 +40,20 @@ class ContainerSettings(BaseModel):
     framework: ContainerFramework = ContainerFramework.docker
     unpack_singularity: bool = False
 
-    model_config = ConfigDict(extra='forbid')
```
agitter: @tristan-f-r was there a reason `ConfigDict` was used twice before?

spras/containers.py
```diff
@@ -189,14 +190,48 @@ def run_container(container_suffix: str, command: List[str], volumes: List[Tuple
...
+        if not container_settings.framework.is_singularity_family:
+            warnings.warn(
+                f"Image override '{image_override}' is a .sif file, but the container framework is "
+                f"'{container_settings.framework}'. .sif overrides are only supported with "
+                f"apptainer/singularity. Falling back to default image.",
+                stacklevel=2
+            )
+        else:
+            print(f'Container image override (local .sif): {image_override}', flush=True)
+    elif image_override:
```
agitter: I'm wondering whether this block needs to be more robust to malformed overrides. What if I provide `hello.world` or `a/b/c/d/e/f`? Do we want to pass that through until there is an error later?

agitter (commented):
These changes make sense to me overall.

config/config.yaml
```diff
@@ -41,6 +41,29 @@ containers:
...
+  #   .sif paths listed here are automatically included in htcondor_transfer_input_files.
+  #   Ignored with a warning if the framework is Docker.
+  #
+  # Example (one of each type):
+  # images:
+  #   omicsintegrator1: "images/omics-integrator-1_v2.sif"        # local .sif (Apptainer only)
+  #   pathlinker: "pathlinker:v1234"                              # image name only (base_url/owner prepended)
+  #   omicsintegrator2: "some-other-owner/oi2:latest"             # owner/image (base_url prepended)
+  #   mincostflow: "ghcr.io/reed-compbio/mincostflow:v2"          # full registry reference (used as-is)
```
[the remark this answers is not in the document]
agitter: Upon further consideration, would it make more sense to nest these image overrides under each algorithm below? They the key would be `image` instead of <algorithm-name>, which may be less typo prone.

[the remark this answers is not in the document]
jhiemstrawisc: I considered that, but to me it felt more appropriate that the container overrides be defined under the container section of configuration. It could just as easily go the other way if you disagree.

spras/config/container_schema.py
tristan-f-r: No reason!

test/test_container_image_resolution.py
```diff
@@ -0,0 +1,201 @@
...
+    @patch("spras.containers.run_container_singularity", return_value="ok")
+    def test_sif_override_with_singularity_passes_through(self, mock_singularity):
+        """.sif override + singularity --> container arg is still prefix/suffix (the .sif is
+        resolved inside run_container_singularity via config.image_override)."""
+        settings = make_settings(framework="singularity", image_override="images/pathlinker_v1234.sif")
+        from spras.containers import run_container
+
+        run_container(CONTAINER_SUFFIX, DUMMY_COMMAND, DUMMY_VOLUMES, DUMMY_WORKDIR, DUMMY_OUTDIR, settings)
+        container_arg = mock_singularity.call_args[0][0]
+        # The actual .sif is used inside run_container_singularity; run_container itself
```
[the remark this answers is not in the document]
jhiemstrawisc: I added additional tests that cover something more end-to-end by mocking the Apptainer client and capturing the image it's called against.

spras/containers.py
jhiemstrawisc: I decided to log a warning but continue the attempt in case the user really does have something that doesn't fit the mold.

config/config.yaml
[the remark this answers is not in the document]
jhiemstrawisc: I made a few important modifications in recent commits. Namely, I created a single source of truth for algorithm names in `spras/config/utils.py`. It was pretty easy to validate the config input from there. This new algorithm "registry" also gets used to build the list of algorithms for `spras/runner.py`, so I also updated the docs instructions for "how to add a new PRM".

config/config.yaml
jhiemstrawisc: Done!

tristan-f-r (commented):
This seems fine from a big picture: I am of course wary of:
- The dynamic algorithm import for the new `ALGORITHM_REGISTRY`, but the dependency diamond there seems unavoidable.
- The strange conflation for having both `image_override` and `images` in the same structure, but I have no clean ideas for addressing that at the moment [and I don't want to block this PR on waiting for a better refactor to come along].

I'll add some smaller comments 👍

spras/config/config.py
```diff
@@ -61,6 +62,21 @@ def __init__(self, raw_config: dict[str, Any]):
...
+        # Validate that all keys in containers.images are recognized algorithm names
+        if self.container_settings.images:
+            valid_names = get_valid_algorithm_names()
+            for key in self.container_settings.images:
+                try:
+                    AlgorithmName(key)
+                except ValueError as err:
+                    raise ValueError(
+                        f"Unknown algorithm name '{key}' in configured containers.images. Is there a typo?"
+                        f"Valid algorithm names are: {sorted(valid_names)}"
```
agitter, line 75: Should add a space or newline here. I got:
```
Unknown algorithm name 'omicsintegrator3' in configured containers.images. Is there a typo?Valid algorithm names are: ['allpairs', 'bowtiebuilder', 'diamond', 'domino', 'meo', 'mincostflow', 'omicsintegrator1', 'omicsintegrator2', 'pathlinker', 'responsenet', 'rwr', 'strwr']
```

agitter (commented):
The updates addressed my major comments. I only have one minor new one.

However, my local tests fail. Some are unrelated to these changes, some are related:
```
FAILED test/RWR/test_RWR.py::TestRWR::test_rwr - AssertionError: Output file does not match expected output file
FAILED test/ResponseNet/test_rn.py::TestResponseNet::test_responsenet_required - AssertionError: assert False
FAILED test/ResponseNet/test_rn.py::TestResponseNet::test_responsenet_all_optional - AssertionError: assert False
FAILED test/ST_RWR/test_STRWR.py::TestSTRWR::test_strwr - AssertionError: Output file does not match expected output file
FAILED test/analysis/test_cytoscape.py::TestCytoscape::test_cytoscape - spras.containers.ContainerError: (Command formatted as list: `['python', '/py4cytoscape/cytoscape_util.py', '--output', '/spras/T32FPZG/cytoscape.cys', '--pathway', '/spras/5M63HVH/pathway.txt|test\\analysis\\...
FAILED test/analysis/test_summary.py::TestSummary::test_example_networks[example] - AssertionError: assert False
FAILED test/analysis/test_summary.py::TestSummary::test_example_networks[egfr] - AssertionError: assert False
FAILED test/evaluate/test_evaluate.py::TestEvaluate::test_node_precision_recall_pca_chosen_pathway - assert False
FAILED test/generate-inputs/test_generate_inputs.py::TestGenerateInputs::test_prepare_inputs_networks - AssertionError: assert False
FAILED test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_parse_outputs - AssertionError: assert False
FAILED test/parse-outputs/test_parse_outputs.py::TestParseOutputs::test_duplicate_edges - AssertionError: assert False
FAILED test/test_container_image_resolution.py::TestPrepareSingularityImage::test_local_sif_with_unpack_builds_sandbox - AttributeError: module 'spython' has no attribute 'main'
FAILED test/test_container_image_resolution.py::TestPrepareSingularityImage::test_registry_with_unpack_pulls_and_builds - AttributeError: module 'spython' has no attribute 'main'
FAILED test/test_container_image_resolution.py::TestPrepareSingularityImage::test_local_sif_no_unpack_returns_sif_path - ModuleNotFoundError: No module named 'pwd'
FAILED test/test_container_image_resolution.py::TestPrepareSingularityImage::test_no_override_no_unpack_returns_docker_uri - ModuleNotFoundError: No module named 'pwd'
FAILED test/test_container_image_resolution.py::TestPrepareSingularityImage::test_unpack_skips_build_if_sandbox_exists - AttributeError: module 'spython' has no attribute 'main'
```

@tristan-f-r do you want to leave more detailed comments still?

agitter:
I suspect the root cause of my spython and pwd errors is that pwd is Unix only: https://docs.python.org/3/library/pwd.html. If that's true, we could skip those tests or mock pwd.

tristan-f-r:
Sounds good. I didn't find anything in my code passthrough, unfortunately.

tristan-f-r:
[After tinkering with #457,] there does seem to be a better solution to this that avoids the dynamic algorithm loading entirely:

We can localize `images` to specific algorithms, e.g.:

```
allpairs:
  image_override: "..."
  runs: ...
```

and we can consider this as a kind of "run setting." I'll see if I can lift this run setting configuration out of #457 for general usage.

jhiemstrawisc:
> [After tinkering with #457,] there does seem to be a better solution to this that avoids the dynamic algorithm loading entirely:
> 
> We can localize `images` to specific algorithms, e.g.:
> 
> ```
> allpairs:
>   image_override: "..."
>   runs: ...
> ```

Tony and I discussed this approach and ultimately decided against it because the chosen approach keeps container logic in the yaml's container configuration.

spras/config/util.py
```diff
@@ -48,6 +48,33 @@ def _missing_(cls, value: Any):
...
+    "responsenet":     ("spras.responsenet", "ResponseNet"),
+    "rwr":             ("spras.rwr", "RWR"),
+    "strwr":           ("spras.strwr", "ST_RWR"),
+}
+
+# Auto-generated enum from the registry keys. Inherits CaseInsensitiveEnum so
+# AlgorithmName("PathLinker") resolves to AlgorithmName.pathlinker.
+AlgorithmName = CaseInsensitiveEnum("AlgorithmName", {k: k for k in ALGORITHM_REGISTRY})
+
+def get_valid_algorithm_names() -> set[str]:
```
tristan-f-r, line 73: Should we have a way to override the Cytoscape docker image as well?

tristan-f-r (commented):
Small cytoscape note: I don't have an immediate way to address that problem without making cytoscape a psuedo-algorithm, so I'm wondering if we even need to worry about it.

tristan-f-r:
> Should we have a way to override the Cytoscape docker image as well?

In line with this, noting that `resolved_container_image` effectively ([with one warning-based exception](https://github.com/Reed-CompBio/spras/pull/464/changes#diff-60126de57de4773c60e11f224d387e916517789cef9fc8db3f1a4db5af3ff4e1R221)) ignores `container_suffix` when the respective `image_override` is specified, we can instead have it so that instead of each algorithm specifying their own hard-coded `container_suffix` that is overriden in `containers.py`, we can instead have each algorithm poll `ProcessedContainerSettings` directly. As a more concrete sketch of this with e.g. PathLinker, we can have the following:

(for [this line in `pathlinker.py`](https://github.com/Reed-CompBio/spras/blob/87b314c53c07c2ecf96c9016db24883beab7f2d9/spras/pathlinker.py#L109):)
```diff
-container_suffix = "pathlinker:v2"
+container_suffix = container_settings.resolve_image("pathlinker")
```

where then, in `config.yaml`, `containers.registry.images` would in-code have all of the defaults set to the appropriate container versions, which we can do quite easily with pydantic. For the example above, `containers.registry.images.pathlinker` would have the default value `pathlinker:v2`, and we would use your existing "container string -> resolved image" pipeline for `ProcessedContainerSettings#resolve_image`.

This means that we do have to maintain two separate lists of docker containers and algorithms, but then that also means we can differentiate between our list of algorithms versus other containers (e.g. Cytoscape), while getting free test-time guarantees that all algorithms are specified, since we explicitly would depend on the default values specified in such a container registry when calling `resolve_image`. (This also generalizes to, say, cases where algorithms have run several docker containers with different tooling: we already see behavior approaching this with DOMINOs usage of two resource-heavy tools, though they are luckily present in the same container with little difference in dependencies.)

[If you agree with this change but also don't want to implement it, I can try my hand at making a PR over your fork of SPRAS to make these changes.]

agitter:
@tristan-f-r in your proposed redesign, what would the new `config.yaml` syntax look like? Would that change?

Where would the two lists of docker containers and algorithms need to be maintained? We would still have the `ALGORITHM_REGISTRY` in `config.util` and now we would also have a list of default images in `containers.registry.images`? I'm usually reluctant to have two similar lists to manually maintain because it creates opportunities for errors as new contributors add methods or things diverge over time.

 It was a good catch that we can't override Cytoscape currently. Our existing Cytoscape support is weak, so if we exclude it in this pull request and document that, I am okay deferring a more general solution for a later time when someone needs that functionality.

tristan-f-r:
`config.yaml` would look exactly the same: it's the internals I'm worried about. Specifically, we can drop the use of dynamic module loading in `ALGORITHM_REGISTRY` with this (instead continuing to use the old algorithms dictionary currently in `spras`), since the containers configuration would no longer require knowing algorithm names to inject `image_override` into their respective container settings.

Effectively, I would like to decouple container logic from the rest of SPRAS more than the currently implemented  `ALGORITHM_REGISTRY` solution allows.

agitter (approved):
I created #482 so we can decide what to do with the Cytoscape override separately.

I still get pytest errors when running locally, but they are not related to the changes here so I am not concerned.

PR #467, opened 2026-04-08
RST formatting

jhiemstrawisc:
This PR adds a pre-commit hook for RST formatting, and then subsequently formats all the docs we have.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,7 @@ repos:
     rev: v1.34.0
     hooks:
       - id: typos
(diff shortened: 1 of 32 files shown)


tristan-f-r (approved):
I can confirm locally that running `pre-commit run --all-files` after adding `rstfmt` produces this exact same patch. The output on the website also seems fine to me, but the website can use another look from someone else if anyone else is unconfident about rstfmt. I trust it.

(This also won't have the same problem as yamlfmt, as rstfmt is written in Python.)

agitter (approved):
I skimmed through the preview and didn't see any problems.

PR #470, opened 2026-04-29
chore: update dependencies

tristan-f-r:
I wanted to try out [pyparsing](https://pyparsing-docs.readthedocs.io/en/latest/index.html) for #431, but ran into some outdated dependency issues. We also:
- Bump Python to 3.14 ([the maximal version on its bugfix phase](https://devguide.python.org/versions/))
- Address a note in `get_params_generic` and move validation logic into it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
-requires-python = ">=3.11"
+requires-python = ">=3.13"
 dependencies = [
     "adjusttext==1.3.0",
     "snakemake==9.19.0",
@@ -32,7 +32,7 @@ dependencies = [
     "spython==0.3.14",
 
     # toolchain deps
-    "pip==26.1",
+    "pip==26.0.1",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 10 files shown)


tristan-f-r:
See https://github.com/DataBiosphere/dsub/issues/328 for the associated issue blocking a dependency bump.

ntalluri:
What specifically is this being used for the heuristics? Do we need to use pyparsing; is there a simpler way to parse?

tristan-f-r:
Tony mentioned boolean-parse, but it is also an unmaintained library with what seems to be no users. I'm much more comfortable just inlining the parsing logic we need because of that.

At the moment, we use a handwritten parser, but I notably made a few mistakes in said handwritten parser, so I would like to avoid using that in the future (one can also avoid that by adding a bucket of tests to said handwritten parser, but that seems suboptimal compared to having a nice way to declare the parser in the first place). Using `pyparsing` would end up being an easier implementation.

tristan-f-r:
Oh, I can just bump down dsub.

tristan-f-r:
> the parsing logic we need

We could also use an object representation, which would solve the parsing issue with the cost of making it a pain for users to specify heuristics.

pyproject.toml
```diff
@@ -16,31 +16,31 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
-requires-python = ">=3.11"
+requires-python = ">=3.13"
```
agitter, line 19: What do we gain and lose by bumping the minimal Python version we support? The typing related changes here don't look significant enough to drop support for older Python versions.

agitter (commented):
This pull request combines bumping dependencies and the typing code. It would be faster to review if we decouple this.

I also want to eliminate all install errors before the ISMB tutorial, so I will make a separate PR just for that.

I noticed that the tests are also failing.

PR #471, opened 2026-05-07
feat: conditional runs

tristan-f-r:
- Depends on #457.

Thanks to a little bit of infrastructure building in the timeout PR, we can give conditional runs a very straightforward patch. **This has no tests**, so I would like to get #431 before this PR so I can add tests without injecting some kind of fake "timing out" algorithm or some kind of test error.

(Also, I need to add docs.)


config/config.yaml
```diff
@@ -73,65 +73,75 @@ algorithms:
...
+          local_search: true
+          rand_restarts: 10
 
   - name: "mincostflow"
     include: true
     runs:
       run1:
-        flow: 1
-        capacity: 1
+        params:
```
ntalluri, line 114: are we allowed to set multiple runs? like params1, params2...?

tristan-f-r: That is from #457, but no, just `params`.

config/config.yaml
ntalluri, line 13: does this also require that immutable_files is set to true for parameter tuning to work?

tristan-f-r: It doesn't! Or, at least for non-HTCondor runs.

ntalluri:
does this depend on #332 #431 #411 as well?

tristan-f-r:
This actually depends on none of those! The next PR, however, will depend on #332 and #457. [Since we start our error handling logic in #457, we don't need to depend on #431 at all.]

PR #474, opened 2026-05-11
refactor: preserve list-generating data

tristan-f-r:
_This PR will also contain configuration generation, especially since it serves as the nicest way to test this refactor, but I'm currently writing a new way to detect neighbors that I like more._

To assist in configuration generation, instead of re-creating the parsing at the CLI level, we store the backing information used to generate any programmatically-generated list in a generic `Tunable` interface, such that all algorithm parameters in the `AlgorithmUnion` inherit from `Tunable`. For the sake of parameter tuning, `Tunable` also exposes `Tunable#tune`, whose contract is specified in-documentation.

- [ ] Depends on #470 for more powerful types.
- [ ] Depends on #471 [The current code, however, does not]
- [ ] Depends on #332 [The current code, however, does not].


spras/config/tunable.py
ntalluri, line 1: is there a way to allow for someone to just put a list and have that tuned? I was assuming that was what TunableSet was but wanted to double check.

PR #483, opened 2026-06-08
Propagate enable_profiling through ProcessedContainerSettings

jhiemstrawisc:
`ProcessedContainerSettings.from_container_settings()` rebuilds the processed container settings from the raw `ContainerSettings`, but it never copied the `enable_profiling` field, so the processed settings always used the dataclass default of False. `run_container_singularity()` reads the flag off the processed settings (`config.enable_profiling`), so the profiling branch never ran: no peer cgroup was created, cgroup_wrapper.sh never executed, and no usage-profile.tsv was written, regardless of the user's `containers.enable_profiling` config value.

This regressed in #390, which moved `enable_profiling` from a top-level config field to the nested container settings and added the field to both `ContainerSettings` and `ProcessedContainerSettings`, but missed wiring it through the conversion in `from_container_settings()`.

Add the missing assignment, plus a regression test asserting that `enable_profiling` survives the full config-parsing path.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -273,6 +273,16 @@ def get_algorithm_image(wildcards):
     return None
 
 # Run the pathway reconstruction algorithm
+# When profiling is enabled, each reconstruct job writes a 'usage-profile.tsv' alongside its raw-pathway.txt.
+# Declare it as a rule output so that Snakemake tracks it and, critically, transfers it back from the execute
+# node in environments without a shared filesystem (e.g. HTCondor with shared-fs-usage: none). This is only
+# produced when profiling is actually performed, which requires the singularity/apptainer container framework
+# (see run_container_singularity), so we gate on the same condition to avoid declaring an output that is never
+# created (which would fail the rule).
+profiling_outputs = {}
+if container_settings.enable_profiling and container_settings.framework.is_singularity_family:
+    profiling_outputs['profile_file'] = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'usage-profile.tsv'])
+
 rule reconstruct:
     input: collect_prepared_input
     # Each reconstruct call should be in a separate output subdirectory that is unique for the parameter combination so
@@ -280,7 +290,9 @@ rule reconstruct:
     # Overwriting files can happen because the pathway reconstruction algorithms often generate output files with the
     # same name regardless of the inputs or parameters, and these aren't renamed until after the container command
     # terminates
-    output: pathway_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt'])
+    output:
+        pathway_file = SEP.join([out_dir, '{dataset}-{algorithm}-{params}', 'raw-pathway.txt']),
+        **profiling_outputs
     resources:
         htcondor_transfer_input_files=get_algorithm_image
     run:
--- a/spras/config/container_schema.py
+++ b/spras/config/container_schema.py
@@ -95,6 +95,7 @@ class ProcessedContainerSettings:
             unpack_singularity=unpack_singularity,
             base_url=container_base_url,
             prefix=container_prefix,
+            enable_profiling=settings.enable_profiling,
             hash_length=hash_length,
             images=dict(settings.images),
         )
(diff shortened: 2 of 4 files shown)


agitter (approved):
The code changes look good. I have not tested it yet, which one of us may want to do before merging. I tried testing in CHTC but hit other problems described in https://github.com/Reed-CompBio/spras/pull/459#pullrequestreview-4477907217

ntalluri:
I updated my code on my current CHTC set up with these changes. I am currently waiting for my jobs to get scheduled to test if this fixed the bug.

ntalluri (dismissed):

ntalluri:
I accidentally approved this. I still need to test this on CHTC, I realized that the SPRAS Docker image I was using wasn't including this code change so it still looked like a bug.

jhiemstrawisc:
I'm testing this now, and I've found a few more subtle issues -- don't merge yet!

jhiemstrawisc:
The most recent commits address two other issues I ran into while testing. The first fixes a runtime error that I think I introduced at some point related to the way I used `subprocess` to invoke apptainer on the EP. The second declares the profiling file as an output so it's transferred back to the AP. This second point was previously hidden by the fact that the executor used to transfer the EP's entire output director rather than individual files (which was itself a bug because it caused Snakemake to re-run completed jobs).

At this point I've tested profiling end to end and this works for me.

ntalluri:
I tested this update and was able to get the profiling files. However, every value inside the files is N/A:

```
egfr_large-strwr-params-CYXCVH5/usage-profile.tsv 
peak_memory_bytes	cpu_usage_usec	cpu_user_usec	cpu_system_usec
N/A	N/A	N/A	N/A

egfr_large-strwr-params-KZG4KAJ/usage-profile.tsv 
peak_memory_bytes	cpu_usage_usec	cpu_user_usec	cpu_system_usec
N/A	N/A	N/A	N/A
```

ntalluri:
Looking at one of the reconstruct stdout outputs I get:

```
Container image override (local .sif): apptainer_images/strwr.sif
singularity build --sandbox unpacked/strwr apptainer_images/strwr.sif
WARNING: 'nodev' mount option set on /srv, it could be a source of failure during build process
INFO:    Starting build...
INFO:    Verifying bootstrap image apptainer_images/strwr.sif
INFO:    Extracting local image...
INFO:    Creating sandbox directory...
INFO:    Build complete: unpacked/strwr
WARNING: 'nodev' mount option set on /srv, it could be a source of failure during build process
INFO:    Starting build...
INFO:    Verifying bootstrap image apptainer_images/strwr.sif
INFO:    Extracting local image...
INFO:    Creating sandbox directory...
INFO:    Build complete: unpacked/strwr
Resolved singularity image to sandbox: unpacked/strwr
Failed to create cgroup: [Errno 30] Read-only file system: '/sys/fs/cgroup/spras-peer-19'
Reading memory and CPU stats from cgroup
Failed to read memory usage from cgroup: [Errno 2] No such file or directory: '/sys/fs/cgroup/spras-peer-19/memory.peak'
Failed to read cpu.stat from cgroup: [Errno 2] No such file or directory: '/sys/fs/cgroup/spras-peer-19/cpu.stat'
    My cgroup path is: /sys/fs/cgroup/spras-peer-19
    /usr/local/lib/python3.11/site-packages/spras/cgroup_wrapper.sh: line 15: /sys/fs/cgroup/spras-peer-19/cgroup.procs: No such file or directory
    Executing command: apptainer exec --bind /srv/scratch/output/egfr/prepared/egfr_large-strwr-inputs:/spras/AOZVQUF --bind /srv/scratch/output/egfr/prepared/egfr_large-strwr-inputs:/spras/4IWLUPR --bind /srv/scratch/output/egfr/prepared/egfr_large-strwr-inputs:/spras/ZBD3RLG --bind /srv/scratch/output/egfr/egfr_large-strwr-params-CYXCVH5:/spras/GN4QBOE/egfr_large-strwr-params-CYXCVH5 --cleanenv --containall --pwd /spras --env SPRAS=True unpacked/strwr python /ST_RWR/ST_RWR.py --network /spras/ZBD3RLG/network.txt --sources /spras/AOZVQUF/sources.txt --targets /spras/4IWLUPR/targets.txt --output /spras/GN4QBOE/egfr_large-strwr-params-CYXCVH5/output.txt --alpha 0.05
```

ntalluri (changes requested):
Something is broken where the profiling files all have N/A values.

agitter:
The error `Failed to create cgroup: [Errno 30] Read-only file system: '/sys/fs/cgroup/spras-peer-19'` seems to point to
https://github.com/Reed-CompBio/spras/blob/11091cde575156df351cbd62f7dfe268a70e6ecc/spras/profiling.py#L23-L32

Did that work previously and no longer does? Should we return None or fail differently in that block of code if we can't create the directory?

Issue #480, opened 2026-06-03
Notes from a first-time contributor going through the tutorial


Jeebjean:
## Context

I just went through the full contributing tutorial (Steps 0–7) implementing the Local Neighborhood algorithm as part of my SROP internship with @agitter. The tutorial explains the concepts clearly. I'm opening this issue to share the spots where I had to guess or trial-and-error my way through, in case any of it is useful for future contributors. @agitter suggested I post these as an issue first so the team can decide what's worth acting on.

These are observations from one person on macOS, so a lot of the setup notes are macOS-specific and may not generalize.

### Environment setup

The prerequisites section lists topics and links but assumes a working dev environment. On my machine I had to figure out:

- Installing Docker on Mac. @ntalluri recommended OrbStack over Docker Desktop and it worked really well for me, so it might be worth mentioning as an option in the docs. One small gotcha with OrbStack: I had to set `DOCKER_HOST` so the Python `docker` library could find the socket.
- Installing the right Miniconda installer for Apple Silicon
- Running `conda init zsh` (zsh is the default shell on recent macOS)
- Quoting `python -m pip install -e '.[dev]'` in zsh
- Generating a GitHub PAT with both `repo` and `workflow` scopes

A short pre-flight checklist might help (`docker --version`, `docker run hello-world`, `docker login`, `conda --version`).

### Step 1: running `local_neighborhood_alg.py`

The exact `cp` commands and the full `python` invocation took me a while to assemble from the repo. I'm not sure if showing them outright would defeat the learning purpose; maybe a hidden "if you get stuck" block would work.

### Step 2: Dockerfile

PathLinker is referenced as the example, but it has `wget`, `apk add`, `envsubst`, etc., which is more than Local Neighborhood needs. A simpler existing wrapper might be a better reference for a first contribution.

### Step 3: wrapper functions

Two small things:

1. The interactive Python block uses `Dataset(dataset_dict)` with a plain dict, but the current API expects a `DatasetSchema` object, so the example fails as-written.
2. A skeleton with the imports and the three `PRM` method stubs would have saved me time figuring out the structure. @agitter mentioned the idea of a stub template file, which would cover this.

### Step 4: registering the algorithm

The exact edits to `spras/runner.py` (import + dictionary entry) and the `config/config.yaml` entry were straightforward once I figured them out, but explicit examples would have been quicker. Side note: the Pydantic config schema wants `include:` directly under `name:` and `runs:`, not under `params:`, which tripped me up briefly.

[...]

PR #486, opened 2026-06-11
Improve contributing tutorial documentation

Jeebjean:
## Summary
Addresses documentation gaps identified in #480.

## Changes made
- Fix `Dataset` API example in Step 3 to use `DatasetSchema` instead of a plain dict (the old example fails with the current API)
- Add pre-flight environment checklist to Prerequisites section
- Add note about PAT `workflow` scope requirement in Step 7

## Out of scope for this PR (team decisions needed)
- Stub template directory for new algorithms
- Replacing PathLinker with a simpler Dockerfile reference
- OS-specific instructions (macOS vs Linux)

cc @agitter
cc @ntalluri

### DIFF
--- a/docs/contributing/index.rst
+++ b/docs/contributing/index.rst
@@ -47,6 +47,18 @@ SPRAS contributions:
    or `10 minutes to pandas
    <https://pandas.pydata.org/pandas-docs/stable/user_guide/10min.html>`__
 
+
+
+   Before getting started, verify your environment is ready with these checks:
+
+.. code:: bash
+
+   docker --version
+   docker run hello-world
+   docker login
+   conda --version
+   git --version
+
 *************************************************
  Step 0: Fork the repository and create a branch
 *************************************************
@@ -202,8 +214,9 @@ and explore the nodes and interactome.
 .. code:: python
 
    > from spras.dataset import Dataset
-   > dataset_dict = {'label': 'data0', 'node_files': ['node-prizes.txt', 'sources.txt', 'targets.txt'], 'edge_files': ['network.txt'], 'other_files': [], 'data_dir': 'input'}
-   > data = Dataset(dataset_dict)
+   > from spras.config.dataset import DatasetSchema
+   > dataset_schema = DatasetSchema(label='data0', node_files=['node-prizes.txt', 'sources.txt', 'targets.txt'], edge_files=['network.txt'], other_files=[], data_dir='input')
+   > data = Dataset(dataset_schema)
    > data.node_table.head()
      NODEID  prize active sources targets
    0      C    5.7   True     NaN    True
@@ -416,6 +429,16 @@ branch and create a pull request. Make sure to commit all of the new and
 modified files and push them to the ``local-neighborhood`` branch on
 your fork.
 
+
+.. note::
+
+   If you are using HTTPS to push and your commit touches files in
+   ``.github/workflows/``, your PAT must have both ``repo`` and
+   ``workflow`` scopes. A PAT with ``repo`` scope only will be rejected
+   with a non-obvious error. SSH push works without this restriction.
+
+
+
 The SPRAS maintainers will review the pull request and provide feedback
 and suggested changes. If you are not already in communication with
 them, you can open a `GitHub issue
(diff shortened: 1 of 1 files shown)


docs/contributing/index.rst
```diff
@@ -47,6 +47,18 @@ SPRAS contributions:
...
+   Before getting started, verify your environment is ready with these checks:
+
+.. code:: bash
+
+   docker --version
+   docker run hello-world
+   docker login
+   conda --version
+   git --version
+
```
ntalluri, line 63: should we also add a check to make sure SPRAS conda env is ready?

ntalluri: python -c "import spras; print('SPRAS import successful')" is what is in the spras tutorial.

docs/contributing/index.rst
```diff
@@ -426,6 +429,16 @@ branch and create a pull request. Make sure to commit all of the new and
 modified files and push them to the ``local-neighborhood`` branch on
 your fork.
 
+
+.. note::
+
+   If you are using HTTPS to push and your commit touches files in
```
ntalluri: Can we just explicitly say for developers to use SSH to push to SPRAS in the contributing guide? And then add a warning that if you use https the issues you mentioned will happen.

docs/contributing/index.rst
```diff
@@ -435,6 +448,8 @@ feedback. However, once the pull request has been approved, it will
 ``main`` branch of the fork stays synchronized with the ``main`` branch
 of the main SPRAS repository.
 
+
```
ntalluri: Can we remove these new white space lines?

ntalluri (commented):
This looks good to me. I left some questions on stuff that could be added or changed but they're more suggestions.

docs/contributing/index.rst
```diff
@@ -47,6 +47,18 @@ SPRAS contributions:
    or `10 minutes to pandas
    <https://pandas.pydata.org/pandas-docs/stable/user_guide/10min.html>`__
 
+
+
+   Before getting started, verify your environment is ready with these checks:
```
agitter: Remove the indentation

docs/contributing/index.rst
```diff
@@ -47,6 +47,18 @@ SPRAS contributions:
...
+
+.. code:: bash
+
+   docker --version
```
agitter, line 55: If we want to have these checks, should we also add the expected output? The specific output will be different depending on what each contributor has installed, but I think we should give some guidance about how someone can tell if everything is ready or not. Maybe the signal is that they don't get errors?

docs/contributing/index.rst
agitter: If this affects the workflows updates, let's put it in the step that first affects files there. That's Step 5, right?

There are two types of tokens, which makes things even more confusing. How is this rephrased language?

> These tests modify files in `.github/workflows/`. If you push over HTTPS using a Personal Access Token, make sure the token has permission to modify GitHub Actions workflows. Otherwise, GitHub may reject the push. SSH push works without this restriction.

agitter (changes requested):
If we want to add another Note section in the Prereqs with some of the suggestions for macOS users, that could be a place to provide links about the different shell, miniconda, OrbStack, etc.

I don't see a better example Dockerfile than PathLinker because I find the COPY example important. If you want to propose a language change that suggests parts of the PathLinker Dockerfile to ignore, we could do that.

agitter:
The pre-commit test is failing. The instructions [here](https://spras.readthedocs.io/en/latest/contributing/index.html#pre-commit-hooks) describe how to locally run the file formatting so it passes.

Jeebjean:
Hi @agitter and @ntalluri, I've addressed all the review comments:
- Fixed indentation in the pre-flight checklist
- Added expected output guidance (no errors = environment is ready)
- Added SPRAS environment check
- Moved PAT note to Step 5 where .github/workflows/ is first modified
- Reformulated PAT note as suggested
- Removed extra whitespace lines

agitter (approved):
These changes look good to me. I'm waiting for the tests to pass before merging.

PR #488, opened 2026-06-12
feat: generalize config float parsing

### LINKED ISSUE #484: Config file parsing breaks for scientific notation
`egfr.yaml` contains
```
  - name: omicsintegrator1
    include: true
    runs:
      run1:
        b:
          - 0.55
          - 2
          - 10
        d: 10
        g: 1e-3
        r: 0.01
        w: 0.1
        mu: 0.008
        dummy_mode: ["file"]
```

The `1e-3` no longer works and gives the error
```
ValidationError in file "spras\Snakefile", line 22:
1 validation error for RawConfig
algorithms.1.omicsintegrator1.runs.run1.g
  Value error, This argument "1e-3" was interpreted as a non-function-calling string: it should be a function call (e.g. range(100, 201, 50)), or an int or a float. [type=value_error, input_value='1e-3', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
  File "spras\Snakefile", line 22, in <module>
  File "spras\spras\config\config.py", line 36, in init_global
  File "spras\spras\config\config.py", line 50, in __init__
  File "miniforge3\envs\spras\Lib\site-packages\pydantic\main.py", line 705, in model_validate
```

agitter:
Closes #484

This implements two fixes for the limited syntax supported by the default YAML float specification: https://yaml.org/type/float.html `add_implicit_resolver` is used when we load a YAML file directly. `sanitize_scientific_notation` is used when Snakemake loads the config file because we cannot modify it's YAML loader soon enough.

I used Gemini to help plan the patch and draft code.

### DIFF
--- a/spras/config/config.py
+++ b/spras/config/config.py
@@ -15,6 +15,7 @@ will grab the top level registry configuration option as it appears in the confi
 import copy as copy
 import functools
 import itertools as it
+import re
 import warnings
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


spras/config/config.py
```diff
@@ -28,6 +29,15 @@
 from spras.config.util import AlgorithmName, get_valid_algorithm_names
 from spras.util import LoosePathLike, NpHashEncoder, hash_params_sha1_base32
 
+# Modify YAML float specification when a YAML file is parsed directly
+# Default requires decimal point for scientific notation: https://yaml.org/type/float.html
```
tristan-f-r, line 33: While the YAML specification itself doesn't support this, Pydantic handles coercion of floats from strings just fine, including with scientific notation. This error was instead caused by my assumption that this coercion would happen before any `BeforeValidator` was run, which is notably the opposite of the characterization of `BeforeValidator`: [I've proposed an alternative fix instead](https://github.com/Reed-CompBio/spras/pull/489).

ntalluri: The alternative fix is very clean, I am curious if it will pass the tests Tony set up.

agitter:
Replaced by #489

PR #489, opened 2026-06-14
fix: allow for numeric coercion after python_evalish_coerce

tristan-f-r:
My assumption that BeforeValidators ran after numeric coercion was wrong.

Since the only return types for `python_evalish_coerce` are from `is_numpy_friendly`, we can bake in a check for `python_evalish_coerce`'s guard statement that checks if the input is coercable to a float: this does not cover all of the behavior for Pydantic's coersion (https://pydantic.dev/docs/validation/latest/concepts/conversion_table), but it will at least allow pydantic to coerce numeric strings. [Optimally, we would trigger pydantic's coercion logic ourselves, but it doesn't seem to expose a nice API for this.]

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -67,6 +67,9 @@ def get_test_config():
                 "include": True,
                 "runs": {
                     "strings": {"dummy_mode": ["terminals", "others"], "b": 3},
+                    # We include slightly absurd numbers here for b
+                    # to test https://github.com/Reed-CompBio/spras/issues/484.
+                    "scientific": {"dummy_mode": "terminals", "b": ["1e-2", "1e2"]},
                     # spacing in np.linspace is on purpose
                     "singleton_string_np_linspace": {"dummy_mode": "terminals", "b": "np.linspace(0,    5,2,)"},
                     "str_array_np_logspace": {"dummy_mode": ["others", "all"], "g": "np.logspace(1,1)"}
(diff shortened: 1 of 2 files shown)


agitter:
I'll review and test this as a replacement for #488

test/test_config.py
ntalluri, line 1: Reminder to add the test cases that tony had in #488

tristan-f-r: A few of these tests ([link](https://github.com/Reed-CompBio/spras/pull/488/changes#diff-45fa5f8530eb815c0cf606032e587adedb04dc70e59b203535d51c44654cfc4dR500-R533)) would only serve to test Pydantic itself, but [the tests which check for specific scientific notation](https://github.com/Reed-CompBio/spras/pull/488/changes#diff-45fa5f8530eb815c0cf606032e587adedb04dc70e59b203535d51c44654cfc4dR456-R467) would be good to ensure that the weak `float` test does capture all of the behavior we want.

test/test_config.py
```diff
@@ -67,6 +67,9 @@ def get_test_config():
                 "include": True,
                 "runs": {
                     "strings": {"dummy_mode": ["terminals", "others"], "b": 3},
+                    # We include slightly absurd numbers here for b
+                    # to test https://github.com/Reed-CompBio/spras/issues/484.
+                    "scientific": {"dummy_mode": "terminals", "b": ["1e-2", "1e2"]},
```
agitter:
```suggestion
                    "scientific": {"dummy_mode": "terminals", "b": ["1e-2", "1e2", "+5E+2"]},
```

test/test_config.py
```diff
@@ -296,6 +299,11 @@ def test_config_values(self):
             OmicsIntegrator2Params(dummy_mode=DummyMode.others, b=3)
         ])
 
+        value_test_util('omicsintegrator2', 'scientific', OmicsIntegrator2Params, [
+            OmicsIntegrator2Params(dummy_mode=DummyMode.terminals, b=100),
```
agitter, line 304:
```suggestion
            OmicsIntegrator2Params(dummy_mode=DummyMode.terminals, b=500),
            OmicsIntegrator2Params(dummy_mode=DummyMode.terminals, b=100),
```

agitter (commented):
This implementation is cleaner than mine so I closed my pull request. My small suggestion is to test one additional type of scientific notation, which passes when I try it locally.

agitter (approved):
I will merge. The relevant test `test_config_values` passes. The flaky `test_node_precision_recall_pca_chosen_pathway` failed.

Issue #452, opened 2025-12-08
Flaky PCA test


tristan-f-r:
This `test_node_precision_recall_pca_chosen_pathway` test [(logs_51725883966.zip)](https://github.com/user-attachments/files/24041509/logs_51725883966.zip) is very flaky (failing a considerable amount of the time):

```
2025-12-08T20:08:27.4124818Z =================================== FAILURES ===================================
2025-12-08T20:08:27.4125344Z __________ TestEvaluate.test_node_precision_recall_pca_chosen_pathway __________
2025-12-08T20:08:27.4125658Z 
2025-12-08T20:08:27.4125823Z self = <test_evaluate.TestEvaluate object at 0x7f4157e3b950>
2025-12-08T20:08:27.4126070Z 
2025-12-08T20:08:27.4126225Z     def test_node_precision_recall_pca_chosen_pathway(self):
2025-12-08T20:08:27.4126642Z         output_file = Path(OUT_DIR + 'pr-per-pathway-pca-chosen.txt')
2025-12-08T20:08:27.4127024Z         output_file.unlink(missing_ok=True)
2025-12-08T20:08:27.4127407Z         output_png = Path(OUT_DIR + 'pr-per-pathway-pca-chosen.png')
2025-12-08T20:08:27.4127775Z         output_png.unlink(missing_ok=True)
2025-12-08T20:08:27.4128129Z         output_coordinates = Path(OUT_DIR + 'pca-coordinates.tsv')
2025-12-08T20:08:27.4128748Z         output_coordinates.unlink(missing_ok=True)
2025-12-08T20:08:27.4129034Z     
2025-12-08T20:08:27.4129817Z         file_paths = [INPUT_DIR + 'data-test-params-123/pathway.txt', INPUT_DIR + 'data-test-params-456/pathway.txt',
2025-12-08T20:08:27.4130538Z                       INPUT_DIR + 'data-test-params-789/pathway.txt',  INPUT_DIR + 'data-test-params-empty/pathway.txt']
2025-12-08T20:08:27.4130985Z     
2025-12-08T20:08:27.4131226Z         dataframe = ml.summarize_networks(file_paths)
2025-12-08T20:08:27.4131817Z         ml.pca(dataframe, OUT_DIR + 'pca.png', OUT_DIR + 'pca-variance.txt', str(output_coordinates), kde=True, remove_empty_pathways=True)
2025-12-08T20:08:27.4132330Z     
[...]

PR #497, opened 2026-07-09
chore: bump version to 0.7.0

agitter:
I'm choosing to not update all the docs like in #447.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spras"
-version = "0.6.0"
+version = "0.7.0"
 description = "Signaling Pathway Reconstruction Analysis Streamliner"
 authors = [
   { name = "Anthony Gitter", email = "gitter@biostat.wisc.edu" },


agitter:
The usual PCA test fails.

ntalluri:
Has the PCA test been failing in other PRs? I'm not sure if it is supposed to be failing.

agitter:
Yes, this failed test is the same one from #452. If you check our recent [commits](https://github.com/Reed-CompBio/spras/commits/main/), you can see it passes sporadically.

I started looking at the root cause but will make that post-version 0.7.0 release. The workarounds in #209 aren't enough.

ntalluri:
For not updating the documentation, is that because we want to wait until after the tutorial?

agitter:
For the documentation, there are a lot of places in the repo that mention a version number: https://github.com/Reed-CompBio/spras/pull/447/changes

Most of them have language like "for example, this version". I thought it would be okay if the example version is not the latest because otherwise it introduces another manual step into the release process. Maybe we could template or automate this?

ntalluri:
That makes sense. Can we make an issue to remind us that we need to update the documentation and suggest to automate this process when we put out a release? After that we are ready to merge.

ntalluri (approved):
Approved but waiting on an issue to be made to update documentation.

PR #498, opened 2026-07-10
fix(devcontainer): bump docker-in-docker, introduce lockfile

tristan-f-r:
[A PR was opened which fixed docker-in-docker for Ubuntu 26.04](https://github.com/devcontainers/features/pull/1637), which was merged on May 12th and was a minor version bump for docker-in-docker version 2 (the version of docker-in-docker we use), but [this caused a regression on Ubuntu/**Debian** backends for hosts whose kernel does not load the `ip_tables` kernel module by default (e.g. RHEL-based distributions like Fedora)](https://github.com/devcontainers/features/issues/1659). Our anaconda base image happens to be a Debian-based image, and I happen to be running on a host which does not load the `ip_tables` kernel module by default.

We fix this by both:
1. Bumping docker-in-docker to address the immediate issue.
2. Adding an associated lockfile which pins devcontainer features, while also pinning the base anaconda image version, to make sure issues of this sort never happen again.

_Reviewers: Make sure to test this in a devcontainer locally/in a codespace as well to make sure this works,_ though this issue shouldn't & doesn't seem to affect codespaces.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/anaconda:1-3
+FROM mcr.microsoft.com/devcontainers/anaconda:1.3.22-3
 
 # Install dependencies necessary to build Singularity
 RUN sudo apt-get update && sudo apt-get install -y \
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,14 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:4": {
+      "version": "4.0.0",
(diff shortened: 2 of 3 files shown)


agitter:
How do you generate the `.devcontainer/devcontainer-lock.json` file if we ever need to update it in the future?

This change makes sense. Merging it before or after the tutorial will depend on whether any of us can test it in time.

tristan-f-r:
They're now generated by default when running `devcontainer build` (or, in more recent VSCode versions, when you build a dev container with the `>Dev Containers: Rebuild and Reopen in Container` action.

agitter:
For testing, would it be sufficient to open a codespace with this new devcontainer and make sure the tutorial or a standard SPRAS run with the default config file still works?

Issue #271, opened 2025-06-12
Discussion on Using PCA on Binary Data


ntalluri:
We wanted to start a discussion about whether PCA is the correct dimensionality reduction method for our current use case in SPRAS. PCA typically assumes that the input data is continuous and approximately Gaussian-distributed, whereas the data we’re working with is binary. While PCA can technically be applied to binary data (and we’re currently using StandardScaler(with_std=False) to center it), we’ve seen indications that the results may be less meaningful in this context. 

Tony shared a helpful paper that outlines some of the limitations of applying PCA to binary data: https://arxiv.org/abs/1510.06112. I found an implementation of logistic PCA that may be useful for our needs: https://cran.r-project.org/web/packages/logisticPCA/vignettes/logisticPCA.html

I also came across a few forum discussions that explore this topic and suggest other possible methods:
- https://stackoverflow.com/questions/40795141/pca-for-categorical-features
- https://stats.stackexchange.com/questions/159705/would-pca-work-for-boolean-binary-data-types

Looking ahead, especially for the benchmarking paper, I will be trying to find a solution that moves beyond using PCA. While it makes sense to keep PCA in SPRAS, given its widespread use, we should also integrate dimensionality reduction methods that are better suited to the binary nature of our data. Just as importantly, we should clearly explain these choices to users so they understand when and why an alternative might be more appropriate.

If anyone has thoughts, relevant papers, tools, or implementations, please feel free to share them below.

There is some more conversation on this idea in #209.

tristan-f-r:
[Also, it could be useful later to preserve ranking information for algorithms that rank edges]

As I understand it, PCA is being performed on a dataframe which lists masks per-algorithm pertaining to a large list of edges.

Is the main goal of PCA here specifically to find which algorithms are similar to each other, like the Jaccard Similarity Heatmap PR, or is the PCA output meant for further preprocessing somewhere else?

If it is the former, it might be worth investigating statistical tools used specifically to analyze graph similarity instead of trying to generalize it.

agitter:
Like many SPRAS ideas, this one comes from [BEELINE](https://www.nature.com/articles/s41592-019-0690-6). I found this figure in their supplement useful for seeing how some types of algorithms produces similar outputs when you project their outputs into a 2D space:
![Image](https://github.com/user-attachments/assets/9878777f-253b-4b8e-80ce-bc11453f6cf3)

[...]

PR #500, opened 2026-07-10
Add Logistic PCA (LPCA) analysis

Jeebjean:
## Summary

Adds Logistic PCA (LPCA) as a new SPRAS analysis method. LPCA is the binary-data counterpart to the existing PCA analysis. It is designed for binary matrices (0/1) like the edge-by-run matrix built from pathway reconstruction outputs, whereas standard PCA assumes continuous data.

### New config block
`analysis.lpca` with five options:
- `include`: enable/disable LPCA (default `false`)
- `k`: number of principal components (default `2`)
- `m`: logisticPCA tuning parameter, used when `cv` is `false` (default `6`)
- `cv`: if `true`, determine m by cross-validation; if `false`, use fixed m (default `false`)
- `transpose`: if `false`, run LPCA on edges x runs matrix (one embedding per edge); 
  if `true`, run on runs x edges matrix (one embedding per pathway run, 
  mirrors the classic PCA analysis) (default `false`)

### New Snakemake rule `lpca_analysis`
- Restricted to algorithms with multiple parameter combinations (same guard as per-algorithm PCA)
- Outputs per algorithm: `{algorithm}-lpca-scores.csv`, `{algorithm}-lpca.png`, `{algorithm}-lpca-coordinates.txt`

### New module `spras/analysis/lpca.py`
- Builds the binary matrix via `summarize_networks`
- Optionally transposes the matrix
- Runs the logisticPCA container via `run_container_and_log`
- Generates a PC1 vs PC2 scatterplot and saves coordinates

### New Docker wrapper `docker-wrappers/lpca/`
- Pinned `rocker/r-base:4.4.2` Dockerfile installing `logisticPCA` and `ggplot2`
- `run_lpca.R`: runs LPCA and saves scores + deviance explained
- `run_cv.R`: cross-validation to find optimal m
- README documenting config options and Docker image publication

### Unit tests `test/analysis/test_lpca.py`
- 3 tests: output file existence, score shape (edges x k), score shape (runs x k with transpose)

### Docker timeout
- Increased from 60s to 600s in `containers.py` to handle larger binary matrices

## Testing locally

Build the image and run the pipeline:
```bash
docker build -t reedcompbio/lpca:v1 docker-wrappers/lpca/
```
Enable `analysis.lpca.include: true` in your config and run as usual.
Tested end-to-end with OmicsIntegrator2 (216 combos) and MincostFlow (14 combos).

## Before / at merge

The image must be published to `reedcompbio/lpca:v1` on Docker Hub by a maintainer:
```bash
docker build -t reedcompbio/lpca:v1 docker-wrappers/lpca/
docker push reedcompbio/lpca:v1
```

cc @agitter @ntalluri


config/config.yaml
```diff
@@ -50,30 +50,6 @@ containers:
   #       requirements = versionGE(split(Target.CondorVersion)[1], "24.8.0") && (isenforcingdiskusage =!= true)
   enable_profiling: false
 
-  # Override the default container image for specific algorithms.
```
agitter, line 53: We need to keep these deleted comments

config/config.yaml
```diff
@@ -272,3 +248,15 @@ analysis:
     # adds evaluation per algorithm per dataset-goldstandard pair
     # evaluation per algorithm will not run unless ml include and ml aggregate_per_algorithm are set to true
     aggregate_per_algorithm: true
+  lpca:
+    include: false
```
agitter, line 278: We should document what true/false does. If true, will SPRAS run PCA and LPCA? That behavior would make sense to me. I initially view LPCA as an additional analysis, not a replacement for PCA. Running them in parallel will help us assess the similarity of the two. If we later find that LPCA is "better" than PCA by some criteria, we could change that behavior.

config/config.yaml
```diff
@@ -272,3 +248,15 @@ analysis:
...
+    # number of principal components to compute
+    k: 2
+    # fixed value of the logisticPCA tuning parameter m, used when cv is false
+    m: 6
+    # if true, choose m by cross-validation; if false, use the fixed m above
+    cv: false
```
agitter, line 288: It could be possible to collapse these into a single parameter `m` that takes `cv` as a valid value. I don't think it is necessary but am sharing the idea in case anyone else prefers that. It would complicate parsing and the config schema as a downside.

Somewhere we should document the analysis in your other repo that led to 6 as a default. It could be in the pull request comments.

docker-wrappers/lpca/README.md
```diff
@@ -0,0 +1,65 @@
+# LPCA (Logistic PCA) wrapper
+
+This wrapper runs [logisticPCA](https://github.com/andland/logisticPCA)
+(Landgraf & Lee, 2020) as a SPRAS analysis step. It reduces the binary
```
agitter:
```suggestion
([Landgraf & Lee, 2020](https://doi.org/10.1016/j.jmva.2020.104668)) as a SPRAS analysis step. It reduces the binary
```

docker-wrappers/LPCA/Dockerfile
agitter, line 1: All of the other Docker wrappers use caps in the subdirectory (for some reason). Should we rename the directory to "LPCA"?

config/config.yaml
```diff
@@ -272,3 +248,15 @@ analysis:
...
+    # matrix orientation given to LPCA:
+    #   false = edges x runs (observations are the edges)
+    #   true  = runs x edges (observations are the runs, mirroring the classic ml pca analysis)
+    transpose: false
```
agitter: I prefer to remove this option and always run with transpose true, like our PCA does currently. That is the most relevant analysis for SPRAS users.

docker-wrappers/LPCA/README.md
```diff
@@ -0,0 +1,65 @@
+# LPCA (Logistic PCA) wrapper
+
```
agitter, line 2: Add the DockerHub URL once I push the image to reedcompbio (similar to the other Docker wrapper readmes).

docker-wrappers/LPCA/README.md
```diff
@@ -0,0 +1,65 @@
...
+This wrapper runs [logisticPCA](https://github.com/andland/logisticPCA)
+(Landgraf & Lee, 2020) as a SPRAS analysis step. It reduces the binary
+edge-by-run matrix built from a set of pathway reconstruction outputs to a small
+number of components and reports the proportion of deviance explained.
+
+The analysis is driven by the `analysis.lpca` config block and the
+`lpca_analysis` Snakemake rule, and is implemented in `spras/analysis/lpca.py`.
+
+## Configuration
```
agitter, line 14: Update if needed based on the config comments

docker-wrappers/LPCA/README.md
```diff
@@ -0,0 +1,65 @@
...
+For the SPRAS default registry to resolve the image, it must be published as
+`docker.io/reedcompbio/lpca:v1`, which requires access to the `reedcompbio`
+Docker Hub organization:
+
+    docker build -t reedcompbio/lpca:v1 docker-wrappers/lpca/
+    docker push reedcompbio/lpca:v1
+
+## How SPRAS resolves the image
+
+SPRAS builds the image reference as `<registry.base_url>/<registry.owner>/<tag>`.
```
agitter: I'm not sure that we need this paragraph. It is how all images work in SPRAS, and it isn't strictly true. If that part of the config is deleted, a default will be used that has the organization set:
https://github.com/Reed-CompBio/spras/blob/07707bf766166496c40d8d29cb66166140167806/spras/config/container_schema.py#L17

docker-wrappers/LPCA/run_lpca.R
```diff
@@ -0,0 +1,30 @@
...
+args = commandArgs(trailingOnly = TRUE)
+input_file = args[1]
+output_file = args[2]
+k = as.integer(args[3])
+m = as.numeric(args[4])
+
+# Load data
+library(logisticPCA)
+data = read.csv(input_file, row.names = NULL)
+party = data[, 1]
```
agitter: This variable is specific to the voting example, right? That could be confusing.

docker-wrappers/LPCA/run_lpca.R
```diff
@@ -0,0 +1,30 @@
...
+data = data[, -1]
+data_matrix = as.matrix(data)
+data_matrix[is.na(data_matrix)] = 0
+
+# Run LPCA
+model = logisticPCA(data_matrix, k = k, m = m, partial_decomp = TRUE)
```
agitter, line 17: Should we make this a command line argument? Is it better to run with partial decomp false for small datasets? How would we make that decision?

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+    output_scores: str,
+    k: int = 2,
+    m: float = 6,
+    cv: bool = False,
+    transpose: bool = False,
+    container_settings=None
+) -> None:
+    """
+    Runs Logistic PCA on the binary edge x algorithm matrix built from SPRAS
+    algorithm output files.
```
agitter, line 32: If we only support running on the transposed dataframe, we can reword this.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+    @param transpose: if True, run LPCA on the transposed (runs x edges) matrix
+        instead of the default (edges x runs)
+    @param container_settings: configure the container runtime (Docker or Singularity)
+    """
+    if not container_settings:
+        container_settings = ProcessedContainerSettings()
+
+    # Step 1: build the binary edge x algorithm matrix
+    print('LPCA: Building binary edge x algorithm matrix...')
+    matrix = summarize_networks(file_paths)
```
agitter: Our PCA function takes the output of summarize_networks as input, doesn't it? Can we do that here? That would better support sharing the output between the two in the future.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Optionally transpose so the runs become the observations rather than the edges
+    if transpose:
+        matrix = matrix.T
+    print(f'LPCA: Matrix shape: {matrix.shape}')
+
+    # Step 2: write the matrix next to the outputs, namespaced by algorithm
```
agitter: Is this the same directory where PCA outputs are written? We should follow that convention.

It looks like it is.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+    output_dir = Path(output_scores).parent
+    output_dir.mkdir(parents=True, exist_ok=True)
+    algo_name = Path(output_scores).name.replace('-lpca-scores.csv', '')
+    matrix_path = str(output_dir / f'{algo_name}-lpca_binary_matrix.csv')
```
agitter: This is the full binary data matrix? I don't think we need to save that here. We eventually can move that to a rule in the Snakefile so it can be reused across algorithms.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+        cv_output_path = str(output_dir / f'{algo_name}-lpca_cv_result.csv')
+        bind_path, mapped_cv_output = prepare_volume(cv_output_path, LPCA_WORK_DIR, container_settings)
+        volumes.append(bind_path)
+
+        print(f'LPCA: Running cross-validation with k={k}...')
+        command_cv = ['Rscript', '/app/run_cv.R', mapped_matrix, mapped_cv_output, str(k)]
+        run_container_and_log('LPCA-CV', LPCA_CONTAINER_SUFFIX, command_cv, volumes,
+                              LPCA_WORK_DIR, None, container_settings)
+
+        m_used = pd.read_csv(cv_output_path)['best_m'][0]
```
agitter, line 86: We could check this file exists before reading it to throw a friendly error.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Step 5: run LPCA with the chosen k and m
+    print(f'LPCA: Running LPCA with k={k}, m={m_used}...')
+    command_lpca = ['Rscript', '/app/run_lpca.R', mapped_matrix, mapped_scores, str(k), str(m_used)]
+    run_container_and_log('LPCA', LPCA_CONTAINER_SUFFIX, command_lpca, volumes,
+                          LPCA_WORK_DIR, None, container_settings)
+
+    print(f'LPCA: Done! Scores saved to {output_scores}')
+
+def plot_lpca(scores_file: str, output_png: str, output_coord: str, labels: bool = True) -> None:
```
agitter, line 100: Somewhere we should document that the KDE parameter selection always uses PCA even if LPCA is run. That may change in the future.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,138 @@
...
+        print('LPCA: Scores file is empty, skipping plot.')
+        return
+
+    # Extract algorithm names from the index
+    column_names = [idx.split('-')[-3] if '-' in idx else idx for idx in scores.index]
+
+    X = scores.values
+    fig, ax = plt.subplots(figsize=(10, 8))
+
+    sns.scatterplot(x=X[:, 0], y=X[:, 1], hue=column_names, s=70, ax=ax)
```
agitter: PCA use `create_palette` to set consistent colors.

agitter (commented):
Note that this will close #271 when merged.

I pushed the image to DockerHub: https://hub.docker.com/r/reedcompbio/lpca

We should add this to our Docker image building test: https://github.com/Reed-CompBio/spras/blob/main/.github/workflows/build-containers.yml

This comment should be modified to note the alternative LPCA
https://github.com/Reed-CompBio/spras/blob/07707bf766166496c40d8d29cb66166140167806/spras/analysis/ml.py#L159

I didn't comment on every detail yet, but in general there are some small things we can to do make this more like the existing PCA.

spras/containers.py
```diff
@@ -359,7 +359,7 @@ def run_container_docker(container: str, command: List[str], volumes: List[Tuple
 
     # Initialize a Docker client using environment variables
     try:
-        client = docker.from_env()
+        client = docker.from_env(timeout=600)
```
agitter, line 362: Is this still needed? Was it a workaround for the memory issues?

This modifies Docker but not Singularity.

test/analysis/test_lpca.py
```diff
@@ -0,0 +1,99 @@
...
+    try:
+        client = docker.from_env()
+        client.images.get('reedcompbio/lpca:v1')
+        return True
+    except docker.errors.ImageNotFound:
+        return False
+    except Exception:
+        return False
+
+skip_if_no_lpca_image = pytest.mark.skipif(
```
agitter: Do we want to silently skip LPCA tests if the image is not available? Shouldn't it always been on DockerHub now that I pushed it?

test/analysis/test_lpca.py
```diff
@@ -0,0 +1,99 @@
...
+class TestLpca:
+    """
+    Run Logistic PCA (LPCA) analysis tests
+    """
+    @classmethod
+    def setup_class(cls):
+        OUT_DIR.mkdir(parents=True, exist_ok=True)
+
+    @skip_if_no_lpca_image
+    def test_lpca_output_exists(self):
```
agitter, line 29: A stronger test would be that the expected output is created when we use known inputs. These are okay first steps, but if there is time we would add the stronger test.

There is some degeneracy with these existing pathways though that cause the PCA tests to fail at random. It may be okay here because we don't do KDE testing.

Will we need to account for sign non-identifiability in LPCA?

Snakefile
```diff
@@ -356,6 +361,7 @@ rule ml_analysis:
         ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
         ml.pca(summary_df, output.pca_image, output.pca_variance, output.pca_coordinates, **pca_params)
 
+
```
agitter, line 371: Can drop the newline

Snakefile
```diff
@@ -403,6 +409,30 @@ rule ml_analysis_aggregate_algo:
         ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
         ml.pca(summary_df, output.pca_image, output.pca_variance, output.pca_coordinates, **pca_params)
 
+rule lpca_analysis:
+    input:
+        pathways = collect_pathways_per_algo
```
agitter, line 447: Can we also support running LPCA on all pathways across algorithms.

Snakefile
```diff
@@ -403,6 +409,30 @@ rule ml_analysis_aggregate_algo:
         ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
         ml.pca(summary_df, output.pca_image, output.pca_variance, output.pca_coordinates, **pca_params)
 
+rule lpca_analysis:
```
agitter: Should this be a new rule or a conditional extension of `ml_analysis` and `ml_analysis_aggregate_algo`?

agitter (commented):
Additional comments because I accidentally submitted the first batch before finishing.

spras/containers.py
Jeebjean: The 600s timeout is still needed. Large matrices (e.g. RWR with ~20k edges) can take several minutes even with partial_decomp=TRUE. Singularity is unaffected.

Snakefile
Jeebjean: Kept it as a separate rule to avoid touching existing ml_analysis logic. Happy to refactor it as a conditional extension and add cross-algorithm support if that's the preferred direction before merge.

docker-wrappers/LPCA/run_lpca.R
Jeebjean: partial_decomp=TRUE is always used. For small datasets it makes no practical difference, and for large ones it is necessary to avoid OOM errors. Happy to make it a config option if that's preferred.

agitter: If it does not have a strong effect on small datasets, I am okay leaving it true all the time. I'd like to document that in the "Configuration" section of the Docker readme.

spras/containers.py
agitter: Is Singularity unaffected because it has a longer default timeout?

I don't see anything in the Docker client documentation that suggests changing this default would be a problem: https://docker-py.readthedocs.io/en/stable/client.html

@ntalluri do you foresee any problems with a 10x longer timeout for Docker runs?

Snakefile
```diff
@@ -91,6 +91,16 @@ def make_final_input(wildcards):
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}ensemble-pathway.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-matrix.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-heatmap.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
+        
+    if _config.config.analysis_include_lpca:
+        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-lpca-scores.csv',out_dir=out_dir, sep=SEP, dataset=dataset_labels, algorithm=algorithms_mult_param_combos))
```
agitter: Looking at our config file structure, currently `ml` and `evaluation` each have a `aggregate_per_algorithm` key that can be set to true/false independently. Having that same key for `lpca` would be most consistent. Then a separate Snakemake rule would run LPCA per algorithm if it is enabled.

If we cannot implement that easily, we could create a new issue to log it as future work.

docker-wrappers/LPCA/README.md
```diff
@@ -1,7 +1,11 @@
 # LPCA (Logistic PCA) wrapper
 
+Docker image: https://hub.docker.com/r/reedcompbio/lpca
+
+This wrapper runs [logisticPCA](https://github.com/andland/logisticPCA)
+
 This wrapper runs [logisticPCA](https://github.com/andland/logisticPCA)
```
agitter: Is this text duplicated now?

spras/containers.py
ntalluri: I feel like this could be a problem, I'm not sure what problems though. Could we try a different timeout or does it have to be 600?

Snakefile
ntalluri: This is fine, we were planning on refactoring all the rest of ml to be separate rules as well.

spras/analysis/lpca.py
```diff
@@ -21,47 +19,45 @@
...
-    # Optionally transpose so the runs become the observations rather than the edges
-    if transpose:
-        matrix = matrix.T
+    matrix = dataframe
+    # Transpose so runs are the observations, mirroring the classic PCA analysis
+    matrix = matrix.T
     print(f'LPCA: Matrix shape: {matrix.shape}')
 
-    # Step 2: write the matrix next to the outputs, namespaced by algorithm
+    # Step 2: write the binary matrix for Docker volume mount
```
agitter, line 55: I understand now. We need this written to disk because LPCA runs through Docker, unlike PCA.

test/analysis/test_lpca.py
agitter, line 1: This test coverage is pretty good. We do not test that the LPCA outputs matched expected outputs. That testing causes complications for PCA, however, so it may be best to defer it.

Snakefile
```diff
@@ -409,22 +414,24 @@ rule ml_analysis_aggregate_algo:
         ml.hac_horizontal(summary_df, output.hac_image_horizontal, output.hac_clusters_horizontal, **hac_params)
         ml.pca(summary_df, output.pca_image, output.pca_variance, output.pca_coordinates, **pca_params)
 
-rule lpca_analysis:
```
agitter: As noted elsewhere, we likely want a second rule that runs per-algorithm. If that is outside the scope of this pull request, we can add an issue to do it later.

config/config.yaml
```diff
@@ -272,3 +272,15 @@ analysis:
     # adds evaluation per algorithm per dataset-goldstandard pair
     # evaluation per algorithm will not run unless ml include and ml aggregate_per_algorithm are set to true
     aggregate_per_algorithm: true
+  lpca:
+    # if true, runs LPCA in addition to the existing PCA (both analyses will run in parallel).
+    # Running both helps compare the two methods on the same data.
+    include: false
```
agitter, line 278: Let's turn this to true for at least one run in GitHub Actions. If it succeeds, we can turn it false again.

Snakefile
```diff
@@ -91,6 +91,16 @@ def make_final_input(wildcards):
...
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-matrix.txt',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
         final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}jaccard-heatmap.png',out_dir=out_dir,sep=SEP,dataset=dataset_labels,algorithm_params=algorithms_with_params))
+        
+    if _config.config.analysis_include_lpca:
+        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-lpca-scores.csv',out_dir=out_dir, sep=SEP, dataset=dataset_labels, algorithm=algorithms_mult_param_combos))
+        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-lpca.png',out_dir=out_dir, sep=SEP,dataset=dataset_labels,algorithm=algorithms_mult_param_combos))
+        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-lpca-coordinates.txt',out_dir=out_dir, sep=SEP, dataset=dataset_labels,algorithm=algorithms_mult_param_combos))
+        final_input.extend(expand('{out_dir}{sep}{dataset}-ml{sep}{algorithm}-lpca-binary-matrix.csv',out_dir=out_dir, sep=SEP, dataset=dataset_labels,algorithm=algorithms_mult_param_combos))
```
agitter: LPCA in a SPRAS workflow failed for me locally. I'm wondering if it is because we request these algorith-specific outputs but do not have a rule to create them.

```
$ snakemake --cores 4 --configfile config/config.yaml
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
MissingInputException in rule all in file "spras\Snakefile", line 146:
Missing input files for rule all:
    affected files:
        output/data0_72021389-ml/omicsintegrator2_72021389-lpca-scores.csv
        output/data1_72021389-ml/strwr_72021389-lpca.png
        output/data0_72021389-ml/rwr_72021389-lpca.png
        output/data1_72021389-ml/omicsintegrator1_72021389-lpca-coordinates.txt
        output/data1_72021389-ml/strwr_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/strwr_72021389-lpca-coordinates.txt
        output/data0_72021389-ml/omicsintegrator1_72021389-lpca.png
        output/data0_72021389-ml/omicsintegrator1_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/pathlinker_72021389-lpca-coordinates.txt
        output/data1_72021389-ml/pathlinker_72021389-lpca.png
        output/data1_72021389-ml/omicsintegrator1_72021389-lpca.png
        output/data0_72021389-ml/rwr_72021389-lpca-binary-matrix.csv
        output/data0_72021389-ml/omicsintegrator1_72021389-lpca-scores.csv
        output/data1_72021389-ml/strwr_72021389-lpca-scores.csv
        output/data1_72021389-ml/pathlinker_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/rwr_72021389-lpca-scores.csv
        output/data0_72021389-ml/omicsintegrator1_72021389-lpca-coordinates.txt
        output/data0_72021389-ml/rwr_72021389-lpca-scores.csv
        output/data0_72021389-ml/omicsintegrator2_72021389-lpca-coordinates.txt
        output/data1_72021389-ml/omicsintegrator1_72021389-lpca-binary-matrix.csv
        output/data0_72021389-ml/pathlinker_72021389-lpca-coordinates.txt
        output/data0_72021389-ml/strwr_72021389-lpca-scores.csv
        output/data1_72021389-ml/rwr_72021389-lpca.png
        output/data0_72021389-ml/pathlinker_72021389-lpca.png
        output/data0_72021389-ml/strwr_72021389-lpca-coordinates.txt
        output/data1_72021389-ml/rwr_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/pathlinker_72021389-lpca-scores.csv
        output/data1_72021389-ml/omicsintegrator2_72021389-lpca-scores.csv
        output/data0_72021389-ml/omicsintegrator2_72021389-lpca.png
        output/data0_72021389-ml/strwr_72021389-lpca.png
        output/data0_72021389-ml/strwr_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/omicsintegrator2_72021389-lpca-coordinates.txt
        output/data1_72021389-ml/rwr_72021389-lpca-coordinates.txt
        output/data0_72021389-ml/pathlinker_72021389-lpca-binary-matrix.csv
        output/data1_72021389-ml/omicsintegrator2_72021389-lpca.png
        output/data1_72021389-ml/omicsintegrator2_72021389-lpca-binary-matrix.csv
        output/data0_72021389-ml/rwr_72021389-lpca-coordinates.txt
        output/data0_72021389-ml/pathlinker_72021389-lpca-scores.csv
        output/data1_72021389-ml/omicsintegrator1_72021389-lpca-scores.csv
        output/data0_72021389-ml/omicsintegrator2_72021389-lpca-binary-matrix.csv
```

agitter (commented):
I am testing locally, and my workflow is stuck at the Cytoscape step. I wonder if that is related to the longer Docker timeout and it would have quickly failed otherwise? I eventually killed Snakemake.

After disabling Cytoscape, I ran into errors at the `ml_analysis` step. I didn't get a specific error message though, and I don't see what changes in this pull request would have affected that.

As a reminder for myself, I should rebuild and the Docker image when I merge this.

spras/containers.py
Jeebjean: > Is Singularity unaffected because it has a longer default timeout?
> 
> I don't see anything in the Docker client documentation that suggests changing this default would be a problem: https://docker-py.readthedocs.io/en/stable/client.html
> 
> @ntalluri do you foresee any problems with a 10x longer timeout for Docker runs?

Thanks, good point to check. From what I can see, Singularity doesn't use the docker-py client at all: run_container_singularity goes through spython / a subprocess, so I don't think this timeout applies there. It's a separate code path rather than a longer default. And thanks for the docs link: since containers.run is blocking, the client stays open for the whole run, which is why the 60s default was too short. Let me know if I'm missing something.

Jeebjean: > I feel like this could be a problem, I'm not sure what problems though. Could we try a different timeout or does it have to be 600?

Fair concern, I wasn't sure either at first. From what I understand, the timeout is just a ceiling, not a fixed wait, so a quick container still returns quickly and raising it shouldn't slow anything down. I picked 600 as a safe margin above the slowest run I saw. Happy to try a lower value or make it a config option (defaulting to 600), whatever you and Tony prefer.

docker-wrappers/LPCA/run_lpca.R
Jeebjean: I've documented partial_decomp in the Configuration section of the Docker README.

config/config.yaml
Jeebjean: Done. I set lpca include: true, confirmed GitHub Actions runs LPCA end-to-end (the workflow and build-and-remove-lpca jobs passed), then set it back to false in a follow-up commit.

test/analysis/test_lpca.py
Jeebjean: Added the stronger test in 5b64162. It runs LPCA on the known test inputs and compares the scores to expected values. To handle sign non-identifiability, the comparison is done in absolute value, so a flipped axis won't cause a false failure. That should also avoid the random failures the PCA tests hit.

spras/containers.py
agitter: If it truly is a ceiling, then I would like to keep it at 600s and see if it causes problems after merging.

For Singularity (Apptainer) I do not see a timeout argument for the `exec` command https://apptainer.org/docs/user/main/cli/apptainer_exec.html or package's `Client.execute` https://singularityhub.github.io/singularity-cli/api/modules/spython/main/execute.html (though it supports 
`singularity_options`, so I can't tell what else is buried there).

Apptainer `run` does have a relevant arugment https://apptainer.org/docs/user/main/cli/apptainer_run.html
```
--runscript-timeout string      set the timeout duration for runscript (default 1m)
```
We don't use it though, so let's leave this alone for now.

config/config.yaml
agitter: I checked that build and confirmed the rule ran in the logs.

Snakefile
agitter: This error is gone locally.

test/analysis/test_lpca.py
agitter: That approach should work well here. For PCA, there were additional testing robustness complications because the result of the PCA is used for KDE density estimation and representative output selection. LPCA isn't used for that though.

docker-wrappers/LPCA/README.md
```diff
@@ -26,6 +25,19 @@ binary matrix has more than one column. It also needs a reasonable number of
 observations to be meaningful; very small inputs (such as the bundled example
 datasets) produce degenerate results, which is why it is disabled by default.
 
+  ### `partial_decomp`
+
+  The LPCA wrapper always runs `logisticSVD` with `partial_decomp = TRUE`, which
+  uses a truncated (rARPACK-based) decomposition instead of a full one. This is
+  hardcoded rather than exposed as a parameter:
+
+  - On small datasets it has no practical effect on the result.
```
agitter, line 34: Noting that the package documentations suggests that it can slow down decomposition for small datasets. I think it is still fair to say that is harmless, so I don't recommend a change. I'm noting it in case we have future problems.

> Since logistic PCA requires iteratively computing the eigendecomposition, it can be slow for large data. When `k` is small relative to the number of columns, however, the eigendecomposition can be sped up by only solving for the first `k` eigenvectors. This can be done for all three formulations using the `partial_decomp` argument. We have implemented this using the `RSpectra` package. Be careful when using this argument, because it can substantially slow down computation is the number of columns is small (e.g. ~20) or if `k` is of comparable size to the number of columns. An example of the speed-up is below.

docker-wrappers/LPCA/README.md
```diff
@@ -4,7 +4,6 @@ Docker image: https://hub.docker.com/r/reedcompbio/lpca
 
 This wrapper runs [logisticPCA](https://github.com/andland/logisticPCA)
```
agitter, line 6: Extra linebreak here

docker-wrappers/LPCA/run_lpca.R
```diff
@@ -0,0 +1,30 @@
...
+
+# Save scores
+scores = model$PCs
+rownames(scores) = row_labels
+write.csv(scores, output_file, row.names = TRUE)
+
+# Save deviance explained
+deviance_file = sub("\\.csv$", "_deviance.txt", output_file)
```
agitter, line 25: I'm now able to inspect the outputs of LPCA. Does it only write a scalar deviance explained unlike PCA, which provides variance per dimension? Does this account for the `k` requested or is it the full dimensionality?

LPCA deviance file:
```
0.996362653717224
```

PCA variance file:
```
PC1: 86.36363636
PC2: 13.63636364
```

spras/analysis/lpca.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Save PNG
+    Path(output_png).parent.mkdir(parents=True, exist_ok=True)
+    plt.savefig(output_png, dpi=200)
+    plt.close()
+
+    # Save coordinates
+    coord_df = pd.DataFrame(X, columns=['PC1', 'PC2'], index=scores.index)
+    coord_df.to_csv(output_coord)
+    print(f'LPCA: Plot saved to {output_png}')
```
agitter, line 141: What information does the coordinates file provide that is not already in the scores file? It looks redundant, but I may be missing something.

docker-wrappers/LPCA/run_lpca.R
```diff
@@ -0,0 +1,30 @@
...
+writeLines(as.character(model$prop_deviance_expl), deviance_file)
+
+cat("LPCA done! Scores saved to", output_file, "\n")
+cat("Score dimensions:", nrow(scores), "x", ncol(scores), "\n")
+cat("Proportion of deviance explained:", model$prop_deviance_expl, "\n")
```
agitter, line 30: This can be `Inf`, such as data0 `pathlinker_72021389-lpca-scores_deviance.txt` in the example config. PCA gives `nan` PCs on that input.

We may want to create an issue noting this to fix both of them later.

spras/analysis/lpca.py
```diff
@@ -0,0 +1,141 @@
...
+# the registry prefix is resolved from the container settings.
+LPCA_CONTAINER_SUFFIX = 'lpca:v1'
+LPCA_WORK_DIR = '/app'
+
+
+def run_lpca(
+    dataframe: pd.DataFrame,
+    output_scores: str,
+    output_matrix: str,
+    k: int = 2,
```
agitter, line 25: We should probably do some value checking for these arguments like PCA does
```
    if components < 2:
        raise ValueError(f"components={components} must be greater than or equal to 2 in the config file.")
```

spras/analysis/lpca.py
```diff
@@ -0,0 +1,141 @@
...
+
+    X = scores.values
+    fig, ax = plt.subplots(figsize=(10, 8))
+
+    label_color_map = create_palette(column_names)
+    sns.scatterplot(x=X[:, 0], y=X[:, 1], hue=column_names, palette=label_color_map, s=70, ax=ax)
+
+    if labels:
+        for i, label in enumerate(scores.index):
+            ax.annotate(label, (X[i, 0], X[i, 1]), fontsize=6, alpha=0.7)
```
agitter, line 125: The LPCA annotations overlap when the coordinates are identical. PCA attempts to spread out the labels.

Image

Image

agitter (commented):
Now that I can run everything locally, I have more observations about the output. Anything we can't finish today we can log as a new issue so we can merge this as-is.

I am running into errors when aggregating by algorithm. The error message indicates this is because the input binary matrix is too small:
```
RuleException:
ContainerError in file "spras\Snakefile", line 456:
(Command formatted as list: `['Rscript', '/app/run_lpca.R', '/app/YZQWZK5/rwr_72021389-lpca-binary-matrix.csv', '/app/VBUONQG/rwr_72021389-lpca-scores.csv', '2', '6.0']`)
An unexpected non-zero exit status (1) inside the docker image docker.io/reedcompbio/lpca:v1 occurred:
    Error in fun(A, k, nu, nv, opts, mattype = "matrix") :
      nrow(A) and ncol(A) should be at least 3
    Calls: logisticPCA -> <Anonymous> -> <Anonymous> -> svds.matrix -> fun
    Execution halted
```

